PR #1, opened 2021-01-21
chore: Initial project setup

jonaslagoni:
**Description**
Initial Typescript setup with eslint and jest testing.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@asyncapi/generator-model-sdk",
+  "version": "0.1.1",
```
magicmatatjahu, line 3: 0.1.0 :P

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
magicmatatjahu, line 189: 2021 AsyncAPI org?

jonaslagoni: hmm, @derberg should LICENSE file not be pushed from .github repository?

derberg: don't worry, it is the same everywhere, even generator 😄 

LICENSE file is not global. I assume you ask in the context of other files we have there, like security or contributing

magicmatatjahu (approved):

PR #2, opened 2021-01-22
chore: Added basic common model and schema with test

jonaslagoni:
**Description**
Added basic common model and schema together with unit tests as described in https://github.com/asyncapi/shape-up-process/issues/51.

Tests are more or less the same from https://github.com/asyncapi/parser-js/blob/master/test/models/schema_test.js

### DIFF
--- a/src/models/CommonModel.ts
+++ b/src/models/CommonModel.ts
@@ -1,0 +1,20 @@
+import { CommonSchema } from "./CommonSchema";
+import { Schema } from "./Schema";
+
+/**
+ * Common representation for the renderers.
+ * @extends CommonSchema<CommonModel>
+ */
+export class CommonModel extends CommonSchema<CommonModel>{
+    "x-extend"?: string
+    "x-original-schema"?: Schema
+    static toCommonModel(object: Object) : CommonModel{
+        let newCommonModel = new CommonModel();
+        newCommonModel = Object.assign(newCommonModel, object);
+        newCommonModel = CommonSchema.transformSchema(newCommonModel, CommonModel.toCommonModel);
+        if(newCommonModel["x-original-schema"] !== undefined){
+            newCommonModel["x-original-schema"] = Schema.toSchema(newCommonModel["x-original-schema"]);
+        }
+        return newCommonModel;
+    }
+}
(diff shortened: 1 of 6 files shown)


src/models/CommonModel.ts
```diff
@@ -0,0 +1,27 @@
...
+     * Transform object into a type of CommonModel.
+     * 
+     * @param object to transform
+     * @returns CommonModel instance of the object
+     */
+    static toCommonModel(object: Object) : CommonModel{
+        let newCommonModel = new CommonModel();
+        newCommonModel = Object.assign(newCommonModel, object);
+        newCommonModel = CommonSchema.transformSchema(newCommonModel, CommonModel.toCommonModel);
+        if(newCommonModel["x-original-schema"] !== undefined){
```
magicmatatjahu, line 22: Why we treat `x-original-schema` as not undefined in some cases? I think that `x-extend` should be normal `extend: string` prop and `x-original-schema` should be normal `originalSchema: Schema` prop.

jonaslagoni: The idea was to have a clear separation between what is JSON Schema and what is custom keywords. 

It is only sometimes extend would be used, so its an optional property.

Regarding `x-original-schema` we could say that it is a required property? 🤔

magicmatatjahu: > Regarding x-original-schema we could say that it is a required property?

Yep, and should be as instance of `CommonSchema`

> The idea was to have a clear separation between what is JSON Schema and what is custom keywords.

But we can make `x-extend` keyword and the second one as normal JS field for easier handling in code. `schema.originalSchema` is easier to use than `schema['x-original-schema']`. We can have fallback for `x-original-schema`

jonaslagoni: > Yep, and should be as instance of `CommonSchema`

I would make it a Schema type because, original schema is supposed to contain the entire schema, how it was before the simplification stage :)

> But we can make `x-extend` keyword and the second one as normal JS field for easier handling in code. `schema.originalSchema` is easier to use than `schema['x-original-schema']`. We can have fallback for `x-original-schema`

hmm, alright, lets make them normal properties instead then 😄

magicmatatjahu: > I would make it a Schema type because, original schema is supposed to contain the entire schema, how it was before the simplification stage :)

Sorry, I thought Schema 😄

magicmatatjahu (approved):

PR #4, opened 2021-01-22
chore: Added input processor for AsyncAPI with tests

jonaslagoni:
**Description**
Added AsyncAPI input processor for AsyncAPI together with tests.

blocked by #3

### DIFF
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -16,10 +16,10 @@ export class AsyncAPIInputProcessor implements I_InputProcessor {
             doc = input;
         }
         let common = new CommonInputModel();
-        common.originalSchema = doc;
+        common.originalInput= doc;
         common.models = [];
         doc.allSchemas().forEach((schema: AsyncAPISchema) => {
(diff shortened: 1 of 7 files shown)


src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -0,0 +1,57 @@
...
+     */
+    async process(input: any): Promise<CommonInputModel> {
+        var doc: AsyncAPIDocument;
+        if(!AsyncAPIInputProcessor.isFromParser(input)){
+            doc = await parse(input);
+        }else{
+            doc = input;
+        }
+        let common = new CommonInputModel();
+        common.originalInput= doc;
```
magicmatatjahu, line 19: I think that `originalInput` should be `input` not parsed doc, but I understand why you did it in this way.

EDIT: Ok, we can leave it in this shape.

magicmatatjahu (approved):

src/processors/AsyncAPIInputProcessor.ts
jonaslagoni: Yea you are right... But `originalInput` is before simplification step, but yea lets keep it as is for now.

PR #5, opened 2021-01-22
chore: Added main input processor together with tests

jonaslagoni:
**Description**
This PR adds the main input processor together with integration tests

Blocked by #4 and #3

### DIFF
--- a/src/InputProcessor.ts
+++ b/src/InputProcessor.ts
@@ -1,4 +1,5 @@
-import { I_InputProcessor } from "interfaces/I_InputProcessor";
+
+import { AbstractInputProcessor } from "models/AbstractInputProcessor";
 import { CommonInputModel } from "models/CommonInputModel";
 import { AsyncAPIInputProcessor } from "processors/AsyncAPIInputProcessor";
 import { JsonSchemaInputProcessor } from "processors/JsonSchemaInputProcessor";
@@ -6,11 +7,12 @@ import { JsonSchemaInputProcessor } from "processors/JsonSchemaInputProcessor";
 /**
  * Main input processor which figures out the type of input it receives and delegates the processing into separate individual processors.
  */
-export class InputProcessor implements I_InputProcessor {
-  asyncAPIProcessor: I_InputProcessor;
-  jsonSchemaProcessor: I_InputProcessor;
+export class InputProcessor extends AbstractInputProcessor {
+  asyncAPIProcessor: AbstractInputProcessor;
+  jsonSchemaProcessor: AbstractInputProcessor;
 
   constructor(){
+    super();
     this.asyncAPIProcessor = new AsyncAPIInputProcessor();
     this.jsonSchemaProcessor = new JsonSchemaInputProcessor();
   }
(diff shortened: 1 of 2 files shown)


src/InputProcessor.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AbstractInputProcessor } from "models/AbstractInputProcessor";
+import { CommonInputModel } from "models/CommonInputModel";
+import { AsyncAPIInputProcessor } from "processors/AsyncAPIInputProcessor";
+import { JsonSchemaInputProcessor } from "processors/JsonSchemaInputProcessor";
+
+/**
+ * Main input processor which figures out the type of input it receives and delegates the processing into separate individual processors.
+ */
+export class InputProcessor extends AbstractInputProcessor {
+  asyncAPIProcessor: AbstractInputProcessor;
```
magicmatatjahu, line 11: I think that we should have more flexible this class. I mean at the moment we have only processors for AsyncAPI and JSON Schema. Better idea to store processors than fields will be container as Map:

```ts
export class InputProcessor extends AbstractInputProcessor {
  processors: Map<string, AbstractInputProcessor>;
  
  constructor() { super(); }
  
  addProcessor(type: string, processor: AbstractInputProcessor) {
    this.processors.set(type, processor);
  }
  
  async process(object: any, type: string = 'json-schema'): Promise<CommonInputModel> {
    const processor = this.processors.get(type);
    if (processor === undefined) {
      throw Error(...)
    }
    try {
      return processor. process(object);
    } catch(err) {
       ... err
    }
  }
}

const mainProcessor = new InputProcessor();
mainProcessor.addProcessor('json-schema', new JsonSchemaInputProcessor());
mainProcessor.addProcessor('asyncapi', new AsyncAPIInputProcessor());
```

What do you think? I know that I changed the signature of `process` function but I have a problem what we want to handle another inputs that these two (of course in future).

magicmatatjahu (commented):
Great job! I have one, big comment 😄

test/InputProcessor.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import {parse} from '@asyncapi/parser';
+import { InputProcessor } from '../src/InputProcessor';
+import { AsyncAPIInputProcessor } from '../src/processors/AsyncAPIInputProcessor';
+import { JsonSchemaInputProcessor } from '../src/processors/JsonSchemaInputProcessor';
+describe('InputProcessor', function() {
+    let mainProcessor : InputProcessor;
+    beforeAll(() => {
+        mainProcessor = new InputProcessor();
+        const asyncAPI = new AsyncAPIInputProcessor();
+        mainProcessor.addProcessor("asyncapi", asyncAPI);
```
magicmatatjahu, line 12: We should have included  `asyncapi` and `json-schema`  processors  in processors map as default.

magicmatatjahu (approved):
Accepted, but please consider my suggestion :)

test/InputProcessor.spec.ts
jonaslagoni: Alright, thought they would be added afterwards  👍

magicmatatjahu (approved):

PR #13, opened 2021-01-28
chore: add signatures for generators

magicmatatjahu:
**Description**

As in title: make signatures for generators. At the moment only for Java and TypeScript. In future, every language will follow in this same convention.

**Related issue(s)**
See also https://github.com/asyncapi/shape-up-process/issues/53


jonaslagoni (approved):
Nice 👍

PR #14, opened 2021-01-28
chore: added the start of the simplification process together with types

jonaslagoni:
**Description**
This PR introduces simplification of types together with the general setup of the simplification process. It introduces both docs, tests and implementation.

**Related issue(s)**


src/simplification/SimplifyTypes.ts
```diff
@@ -0,0 +1,109 @@
...
+
+
+/**
+ * Find the types for a simplified version of a schema
+ * 
+ * @param schema to find the simplified types for
+ */
+export default function simplifyTypes(schema: Schema | boolean) : string[] {
+  //If we find absence of data format ensure all types are returned
+  if(typeof schema === "boolean"){
```
magicmatatjahu, line 11: `type: false` should be ignored? In this case we go with `true` and `false` and I think that we should go only with `true`. Am I correct?

src/simplification/SimplifyTypes.ts
```diff
@@ -0,0 +1,109 @@
...
+  if(schema.oneOf){
+    schema.oneOf.forEach((allOfSchema) => {
+      addToTypes(simplifyTypes(allOfSchema));
+    });
+  }
+  if(schema.anyOf){
+    schema.anyOf.forEach((allOfSchema) => {
+      addToTypes(simplifyTypes(allOfSchema));
+    });
+  }
```
magicmatatjahu, line 45: You can make one function for that:

```ts
function handleOfSchemas(schemas = []) {
    schemas.forEach((schema) => {
      addToTypes(simplifyTypes(schema));
    });
}

if(schema.allOf) handleOfSchemas(schema.allOf);
// etc
```

src/simplification/SimplifyTypes.ts
```diff
@@ -0,0 +1,109 @@
...
+    };
+    if(schema.enum){
+      types = [];
+      schema.enum.forEach((value: any) => {
+        const inferredType = inferTypeFromValue(value);
+        if(inferredType !== undefined){
+          addToTypes(inferredType);
+        }
+      });
+    }
```
magicmatatjahu, line 85: What about refs for another models etc? I mean in enum you can also add ref to another model, so whats then? Will it be always object? 😄

src/simplification/Simplify.ts
```diff
@@ -0,0 +1,31 @@
...
+      if(typeof schema !== "boolean"){
+        schema.$id = schemaId;
+      }
+      model.$id = schemaId;
+    } else if (schema.$id !== undefined){
+      model.$id = schema.$id;
+    }
+  }
+
+  models.push(model);
```
magicmatatjahu, line 29: For what we have array there? As I see you only perform action on one schema with one result model.

src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -15,9 +25,17 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
...
-        //To be defined in https://github.com/asyncapi/shape-up-process/issues/53
-        throw new Error('Method not implemented.');
+    static convertSchemaToCommonModel(schema: Schema | boolean) : {[key: string]: CommonModel} {
+        const commonModels = simplify(schema);
+        const commonModelsMap : {[key: string]: CommonModel}  = {};
+        commonModels.forEach((value) => {
+            if(value.$id){
+                commonModelsMap[value.$id] = value;
+            }else{
+                console.log("Looks like someone is missing an id");
```
magicmatatjahu, line 35: Haha 😄

magicmatatjahu (commented):
Great! I left some comments with questions to clarification 😄

src/simplification/Simplify.ts
jonaslagoni: Gonna be introduced in the upcoming PR's, but with properties etc it needs to be able to create new models on the fly 😄

src/simplification/SimplifyTypes.ts
jonaslagoni: As far as I know enums can never be refs. They can be objects, but never refs 😄

src/simplification/SimplifyTypes.ts
jonaslagoni: Hmm, I decided to include both here since if you have 
```
{
  "properties": { 
    "testProp1": false,
    ...
  },
  ...
}
```
I would still like to have the property there. It might be a false assumption, but for now I think we will keep it as is 😄

src/simplification/SimplifyTypes.ts
magicmatatjahu: We must contribute to JSON Schema 😄

src/simplification/SimplifyTypes.ts
magicmatatjahu: I have problem with `false`. I remember that we had a talk about it, but `true` is understandable, it means every type, but `false` should be in my opinion undefined :)

src/processors/JsonSchemaInputProcessor.ts
jonaslagoni: yea, placeholder for now until we handle errors correctly. We also need to resolve refs and assign anonymous schemas like our parser. But thats later 😄

src/simplification/SimplifyTypes.ts
jonaslagoni: So you would expect that object above to not contain the property at all?

src/simplification/SimplifyTypes.ts
jonaslagoni: Great idea 👍

src/simplification/SimplifyTypes.ts
jonaslagoni: We kinda are in a sense 😄

src/simplification/SimplifyTypes.ts
magicmatatjahu: I can't figure what `false` can mean here tbh 😄

jonaslagoni: In validation it is simply "reject all values". Hmm I guess I answered my own question that it should be ignored 🤔

fmvilas: Yeah, if a schema is `false` it means it would always evaluate to fail to validate, so I wonder who's actually using this. I'd just parse `true` as `any` and in case of `false` I'd raise an error maybe? Not sure.

jonaslagoni: We could do that as well yea.

jonaslagoni: Lets do that for now then 😄

jonaslagoni:
Thanks 🙇

PR #17, opened 2021-01-28
chore: implemented property simplification

jonaslagoni:
**Description**
This PR introduces property simplification together with docs and tests.

**Related issue(s)**


src/models/CommonModel.ts
```diff
@@ -25,4 +25,74 @@ export class CommonModel extends CommonSchema<CommonModel>{
...
+    /**
+     * Only merge if left side is undefined and right side is sat OR both sides are defined
+     * 
+     * @param mergeTo CommonModel to merge into
+     * @param mergeFrom CommonModel to merge values from
+     */
+    static mergeCommonModels(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema) : CommonModel {
+        const mergeToProperties = mergeTo.properties;
+        const mergeFromProperties = mergeFrom.properties;
+        if(mergeToProperties === undefined && mergeFromProperties !== undefined){
```
magicmatatjahu, line 39: I think that logic like:

```ts
if (mergeFromProperties !== undefined) {
  if (mergeToProperties === undefined) {
     mergeTo.properties = mergeFromProperties;
  } else {
     ...
  }
}
```

is easier to read :) But it's only a suggestion.

src/simplification/SimplifyProperties.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export default function simplifyProperties(schema: Schema) : output {
+  let models : CommonModel[] | undefined;
+  let commonProperties : { [key: string]: CommonModel; } | undefined;
+  const addToModels = (model : CommonModel[]) => {
+    if(models === undefined){
+      models = [];
+    }
+    models = [...models, ...model];
+  }
```
magicmatatjahu, line 19:
```ts
const addToModels = (model : CommonModel[] = []) => { models = [...(models || []), ...model]; }
```

magicmatatjahu (changes requested):
Great! As we discussed, we should go in next week with another handling of CommonModel to handle complex cases like unions, traits etc.

src/simplification/SimplifyProperties.ts
jonaslagoni: Smart 🧠

src/models/CommonModel.ts
jonaslagoni: Yea I agree :)

src/models/CommonModel.ts
```diff
@@ -25,4 +25,76 @@ export class CommonModel extends CommonSchema<CommonModel>{
...
+                    if(mergeToProperties![propName] !== undefined){
+                      mergeTo.properties![propName] = CommonModel.mergeCommonModels(mergeToProperties![propName], prop, originalSchema);
+                    } else {
+                      mergeTo.properties![propName] = prop;
+                    }
+                });
+            }
+        }
+
+        if(mergeTo.items === undefined && mergeFrom.items !== undefined){
```
magicmatatjahu, line 53: I forgot to mention that if you apply changes, you can add them here too, I mean changed if-s with undefined

jonaslagoni: ahh! right, 2 sec

magicmatatjahu (approved):

PR #18, opened 2021-01-28
chore: implemented item simplification

jonaslagoni:
**Description**
This PR introduces the simplification of items together with test and docs.

Blocked by #17 

**Related issue(s)**


src/models/CommonModel.ts
```diff
@@ -49,26 +50,38 @@ export class CommonModel extends CommonSchema<CommonModel>{
...
+                    if(models.length > 0){
+                        let mergedItemsModel : CommonModel = models[0];
+                        models.forEach((model) => {mergedItemsModel = CommonModel.mergeCommonModels(mergedItemsModel, model, originalSchema);});
+                        return mergedItemsModel;
+                    } else {
+                        return undefined;
+                    }
+                } 
+            }
+            return models;
```
magicmatatjahu, line 67: You can simplify this by:

```ts
if(Array.isArray(models) && models.length > 0) {
  // logic
} else {
  return undefined;
}
```

src/models/CommonModel.ts
```diff
@@ -81,13 +94,20 @@ export class CommonModel extends CommonSchema<CommonModel>{
             if(mergeTo.type === undefined){
                 mergeTo.type = mergeFrom.type;
             } else {
-                if(!Array.isArray(mergeTo.type)){
-                    mergeTo.type = [mergeTo.type!]; 
+                const addToType = (type : string) => {
+                    if(!mergeTo.type!.includes(type)){
```
magicmatatjahu, line 98: You should change order of ifs -> first check if types is array and the push, otherwise change type to array, something like:

```ts
if(Array.isArray(mergeTo.type)) {
  mergeTo.type.includes(type) && mergeTo.type.push(type);
} else {
  mergeTo.type = type ? [mergeTo.type, type] : mergeTo.type
}
```

magicmatatjahu (dismissed):
Great! Some suggestions, but lgtm :)

src/models/CommonModel.ts
jonaslagoni: Cant use exactly that since if `models` is not array I want it to return `models` as is. But I simplified it to 
```
if(Array.isArray(models)){
       if(models.length > 0){
           ...
       } else {
              return undefined;
       }
}
...
```

src/models/CommonModel.ts
jonaslagoni: But I need to check whether the type are already included in the else branch as well, so I dont think this simplifies anything 😄

magicmatatjahu (approved):

PR #21, opened 2021-02-01
chore: fixed simplification of types removed primitives

jonaslagoni:
**Description**
This PR ensures that types can be primitives by not resolving into an array all the time. This is to simplify the rendering phase.


magicmatatjahu (approved):
LGTM

PR #24, opened 2021-02-02
chore: fix $id is preserved in original schema when it should not

jonaslagoni:
**Description**
This PR fixes that `$id` is preserved in original schema.

### DIFF
--- a/src/simplification/Simplify.ts
+++ b/src/simplification/Simplify.ts
@@ -18,8 +18,8 @@ export function simplifyRecursive(schema : Schema | boolean) : CommonModel[] {
     //Get the root model from the simplification process which is the last element in the list
     //This is because any intermediary models are put in front of the root
     const rootSimplifiedModel = simplifiedModel[simplifiedModel.length-1];
-    //Only if it contains object and is the only type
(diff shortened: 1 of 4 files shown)


src/simplification/Simplify.ts
```diff
@@ -13,13 +13,17 @@ let anonymCounter = 1;
...
 export function simplifyRecursive(schema : Schema | boolean) : CommonModel[] {
   let models : CommonModel[] = [];
-  let types = simplifyTypes(schema);
   let simplifiedModel = simplify(schema);
-  //Only if it contains object and is the only type
-  if(types !== undefined && typeof schema !== "boolean" && types.includes("object") && simplifiedModel[0].properties !== undefined){
-    let rootModel = new CommonModel();
-    rootModel.$ref = schema.$id;
-    models[0] = rootModel;
+  if(simplifiedModel.length > 0){
```
magicmatatjahu, line 17: `if (simplifiedModel.length)` is enough 😄  As I remember you cannot have length of Array less than 0.

magicmatatjahu (approved):

src/simplification/Simplify.ts
jonaslagoni: True, I like the explicit syntax though 😄

PR #27, opened 2021-02-03
chore: add generator and renderers for Java

magicmatatjahu:
**Description**

- Add generator for Java
- Add class and enum (for string and integer type) renderer for Java types
- Add relevant unit tests
- Add `change-case` package to generate variable names in PascalCase, camelCase, CONSTANT_CASE etc.

**Related issue(s)**
See also https://github.com/asyncapi/shape-up-process/issues/47


src/generators/java/JavaRenderer.ts
```diff
@@ -0,0 +1,97 @@
...
+        return 'double';
+      case 'binary':
+        return 'byte[]';
+      case 'array': {
+        const types = model.items ? this.renderType(model.items) : 'Object';
+        return `Array<${types}>`;
+      }
+      default:
+        return 'Object';
+    }
```
jonaslagoni, line 65: Type does not have that many possibilities? Remember we still only use JSON Schema types

src/helpers/FormatHelpers.ts
```diff
@@ -1,11 +1,53 @@
...
+  }
+
+  /**
+   * Transform into upper case string with an underscore between words.
+   * @param {string} value to transform
+   * @returns {string}
+   */
+  static toConstantCase(value: string): string {
+    return constantCase(value);
+  };
```
jonaslagoni, line 47: Why these wrappers? 🤔

src/generators/java/renderers/EnumRenderer.ts
```diff
@@ -0,0 +1,70 @@
...
+  protected renderItems(type: "String" | "Integer" = "String"): string {
+    const enums = this.model.enum || [];
+    const items = enums.map(e => {
+      const name = FormatHelpers.toConstantCase(type === "String" ? e : `number ${e}`);
+      return `${name}("${e}")`;
+    }).join(", ");
+    return `${items}`;
+  }
+
+  protected renderHelpers(enumName: string, type: "String" | "Integer" = "String"): string {
```
jonaslagoni, line 43: Why only `"String" | "Integer"` ?

jonaslagoni (changes requested):
Just have a few comments for clarification 😄 Remember to write some docs as well :)

src/generators/java/JavaRenderer.ts
magicmatatjahu: Using `type` with `format` fields you can specify complex type. At the moment we go only with `type` and we should also go with format in near future.

src/generators/java/renderers/EnumRenderer.ts
magicmatatjahu: As I see in examples in net, Java cannot handle enums like in TS by `{key} = {value}` syntax, and also you can specify only some primitives in enum like String, Integer (I remember you can also use Boolean).

jonaslagoni: You can with [constructors](https://stackoverflow.com/questions/40547499/how-to-make-key-value-like-enum-in-java) 😄

magicmatatjahu: I also used constructors :P but then you have more complex type. QuickType for example, create enums only for strings, and then if you have defined strings, boolean, integers in enums, QT creates extra type (and treat it as union) to handle boolean and integers :P

src/helpers/FormatHelpers.ts
magicmatatjahu: I used them as plain functions, but I realized that in every renderer I use `FormatHelpers` class, so  putted them inside it.

jonaslagoni: What I mean is, is there a reason for there to be a wrapper around the external functions?

magicmatatjahu: Nope, but I want to use them using `FormatHelpers` not with imports in each renderer :)

magicmatatjahu: Or I can change it to 

```ts
static toConstantCase = constantCase;
```

jonaslagoni: That would be much better IMO 🎉

magicmatatjahu:
@jonaslagoni I added/changed everything what you want. I also make enums more generic - if we will have more that one type in `type` field, then I use `Object` type as value type in enum.

jonaslagoni (approved):
awesome 👍

PR #29, opened 2021-02-03
chore: fixes anonymous id is preserved in session

jonaslagoni:
**Description**
This PR fixes that if two schemas are simplified in the same session the same counter is used throughout.


src/simplification/Simplifier.ts
```diff
@@ -0,0 +1,82 @@
+
+import { CommonModel, Schema } from '../models';
+import simplifyProperties from './SimplifyProperties';
+import simplifyEnums from './SimplifyEnums';
+import simplifyTypes from './SimplifyTypes';
+import simplifyItems from './SimplifyItems';
```
magicmatatjahu, line 6: Maybe you should add those function inside Simplifier as static methods?

jonaslagoni: I would rather have them separated completely tbh instead of everything inside one class 😄

magicmatatjahu: Do you use in other place these function? I think you can include them inside this class :) Also you can make something like `private static {staticMethodName}` and then use inside another static methods `this.{staticMethodName}` 😄

jonaslagoni: I mean I use them in tests separately. Do you suggest adding everything from the separate functions and add them as static functions inside the `Simplifier` class? 🤔

magicmatatjahu: If it is a problem for you, you can leave it, but functions are only used inside `simplify` function

jonaslagoni: Yes, but IMO separate files for these simplification rules are the best solution. The class file in the end would be way too bloated.

magicmatatjahu: OK 😄

magicmatatjahu (approved):

PR #32, opened 2021-02-03
chore: references are now parsed in the JSON Schema processor

jonaslagoni:
**Description**
This PR solves the issue of parsing references before the simplification stage.


src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -47,13 +49,21 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
 
     /**
      * Process a draft 7 schema
+     * 
      * @param input to process as draft 7
      */
-    private processDraft7(input: any) : CommonInputModel {
-        const schema = Schema.toSchema(input);
+    private async processDraft7(input: any) : Promise<CommonInputModel> {
+        const refParser = new $RefParser;
```
magicmatatjahu, line 56: Is new parser needed for every parsing?

jonaslagoni: Ehhh, goood question. I guess not? 😄

jonaslagoni: let me try

magicmatatjahu: Ahh, it should be new 😄 I remember one issue in parser-js that some state in refParser was reassigned in asynchronus functions, so we should stay with it.

magicmatatjahu (approved):

PR #37, opened 2021-02-04
chore: added the simplifier for the extend keyword

jonaslagoni:
**Description**
This PR introduces the simplifier for the extend keyword to allow for inheritance.
- Changed the extend keyword to be an array of strings
- Adapted all tests which should extend
- Introduces Simplifier options, at the moment it is not possible to customize the input process and therefore the simplifier. This is for a later issue.

This is blocked by 
- https://github.com/asyncapi/generator-model-sdk/pull/36
- https://github.com/asyncapi/generator-model-sdk/pull/35


src/models/SimplificationOptions.ts
```diff
@@ -0,0 +1,7 @@
+
+/**
+ * Options passed along in the simplification stage.
+ */
+export class SimplificationOptions {
```
magicmatatjahu, line 5: I don't think so that it should be as class. Please change to interface.

src/processors/InputProcessor.ts
```diff
@@ -30,7 +30,7 @@ export class InputProcessor {
    * @param input to process
    * @param type of processor to use
    */
-  process(input: any): Promise<CommonInputModel> {
+  process(input: any) : Promise<CommonInputModel> {
```
magicmatatjahu, line 33: Blank space is necessary? 😄

src/simplification/Simplifier.ts
```diff
@@ -4,8 +4,16 @@ import simplifyProperties from './SimplifyProperties';
 import simplifyEnums from './SimplifyEnums';
 import simplifyTypes from './SimplifyTypes';
 import simplifyItems from './SimplifyItems';
+import simplifyExtend from './SimplifyExtend';
+import { SimplificationOptions } from '../models/SimplificationOptions';
 export default class Simplifier {
+  options: SimplificationOptions;
   anonymCounter = 1;
+
+  constructor(options: SimplificationOptions = new SimplificationOptions()){
```
magicmatatjahu, line 13: Please change options to interface.

src/simplification/SimplifyExtend.ts
```diff
@@ -0,0 +1,30 @@
+import { CommonModel } from "models";
+import { Schema } from "models/Schema";
+import Simplifier from "./Simplifier";
+type output = {newModels: CommonModel[] | undefined; extendingSchemas: string[] | undefined};
```
magicmatatjahu, line 4: I know that at the moment it's not important but then we should make something like 

```ts
type CommonOutput = { newModels?: CommonModel[] }
type ExtendOutput = CommonOutput & { extendingSchemas?: string[] }
```

src/simplification/Simplifier.ts
```diff
@@ -76,6 +85,15 @@ export default class Simplifier {
           model.enum = enums;
         }
       }
+      if(this.options.allowInheritance){
+        const simplifiedExtends = simplifyExtend(schema, this);
+        if(simplifiedExtends.newModels !== undefined){
```
magicmatatjahu, line 90: if `newModels` has signature `{ newModels: CommonModel[] }` why we check undefined case? 😄

src/simplification/SimplifyExtend.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Find out which common models we should extend
+ * 
+ * @param schema to find extends of
+ */
+export default function simplifyExtend(schema: Schema | boolean, simplifier : Simplifier) : output {
+  let models : CommonModel[] | undefined;
+  let extendingSchemas : string[] | undefined;
```
magicmatatjahu, line 13: You can assign empty array to those properties and then change code from this:

```ts
let models : CommonModel[] | undefined;
let extendingSchemas : string[] | undefined;

extendingSchemas = [...(extendingSchemas || []), rootSimplifiedModel.$id];
models = [...(models || []), ...simplifiedModels];
```

to 

```ts
let extendingSchemas : string[] = [];

extendingSchemas = [...extendingSchemas, rootSimplifiedModel.$id];
models = [...models), ...simplifiedModels];
```

magicmatatjahu (changes requested):

src/simplification/Simplifier.ts
jonaslagoni: It does not? `CommonModel[] | undefined` 
I use undefined instead of checking length of array, since that makes more sense 😄

src/models/SimplificationOptions.ts
jonaslagoni: Then I cant have default values? 🤔

src/simplification/Simplifier.ts
magicmatatjahu: ahh, sorry... I did not notice it 😄

src/models/SimplificationOptions.ts
magicmatatjahu: Yes, in Simplify class's constructor.

magicmatatjahu:
```ts
class Simplifier {
  private static defaultOptions: SimplificationOptions = {}

  constructor(
    options: SimplificationOptions = Simplifier.defaultOptions,
  ) {
    this.options = { ...Simplifier.defaultOptions, ...options }
  }
}
```

src/processors/InputProcessor.ts
jonaslagoni: Haha I guess not, but noticed I had it somewhere else 🤷

src/simplification/SimplifyProperties.ts
```diff
@@ -52,7 +52,10 @@ export default function simplifyProperties(schema: Schema | boolean, simplifier
       }
     }
     //If we encounter combination schemas ensure we recursively find the properties
-    handleCombinationSchemas(schema.allOf);
+    if(simplifier.options.allowInheritance === false){
```
magicmatatjahu, line 55: it should be changed to `simplifier.options.allowInheritance !== false`

magicmatatjahu (approved):

PR #39, opened 2021-02-09
chore: add customisation flow

magicmatatjahu:
**Description**

- Add types for presets
- Migrate JS, TS and Java to presets - create default presets to every language
- Add tests

**Related issue(s)**
See also https://github.com/asyncapi/shape-up-process/issues/55


fmvilas:
I love the presets idea. Very modular and simple to use 👏

derberg:
I just second @fmvilas, I like the presets idea. It looks much cleaner and clearer than in quicktype.

One question to `User should be able to concatenate/merge given presets in order to obtain the desired model shape`. Just to make sure about `concatenate/merge`. I assume you really mean `concatenate` only or will there be some `merging` involved too?

magicmatatjahu:
@derberg I will change this section with concatenation.  In merge, I mean that flow will be executed from `defaultPresets` with order of given presets, like in example: methods from SOME_PRESET_FROM_NPM preset -> methods from ANOTHER_PRESET preset -> class.property in last preset - so I have in mind concatenation :)

derberg:
@magicmatatjahu thanks for the explanation. I just wanted to make sure we will not do any magic merge of the methods with the same name, merge or overwrite

magicmatatjahu:
@derberg Nope, it will be simple chain of execution, you can think about it like interceptors https://en.wikipedia.org/wiki/Interceptor_pattern

src/models/Preset.ts
```diff
@@ -0,0 +1,50 @@
+import { AbstractRenderer } from "../generators/AbstractRenderer";
+import { CommonInputModel } from "./CommonInputModel";
+import { CommonModel } from "./CommonModel";
+
+// TODO: Change any type to correct one
```
jonaslagoni, line 5: 🤔

jonaslagoni (changes requested):
Looks great over all, got a few comments 😄

src/models/Preset.ts
magicmatatjahu: Yes, I should write also why 😄 I don't want to make big PR and in one of next I would like to fixing typing to improve DX, so at the moment I wanna to stay with it.

src/generators/java/JavaGenerator.ts
```diff
@@ -38,12 +36,14 @@ export class JavaGenerator extends AbstractGenerator {
   }
 
   async renderClass(model: CommonModel, inputModel: CommonInputModel): Promise<string> {
-    const renderer = new ClassRenderer(model, inputModel, this.options);
-    return renderer.render();
+    const presets = this.getPresets("class");
```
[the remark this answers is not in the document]
magicmatatjahu: Enum in which case? `this.getPresets` or `renderClass`? If first, then no, because `getPresets` is generic function and for example in GoLang we haven't class but struct 😄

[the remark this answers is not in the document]
jonaslagoni: Was the first one I meant 👍

src/models/Preset.ts
jonaslagoni: Maybe just add an issue pointing to the todo, so we have it in the backlog 😄

src/generators/AbstractRenderer.ts
```diff
@@ -19,10 +23,29 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
...
+  }
+
+  async runPreset(
+    functionName: string,
+    params: object = {},
+  ): Promise<string> {
+    let content = "";
+    for (const [preset, options] of this.presets) {
+      if (typeof preset[functionName] === "function") {
+        content = await preset[functionName]({ ...params, renderer: this, content, options, model: this.model, inputModel: this.inputModel, });
```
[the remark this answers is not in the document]
magicmatatjahu: `, ` doesn't make problems in code but ok, I'll remove it for you 😘

src/models/Preset.ts
magicmatatjahu: Added :) https://github.com/asyncapi/generator-model-sdk/issues/50

jonaslagoni (approved):

PR #41, opened 2021-02-10
chore: add functionality to handle circular schemas

jonaslagoni:
**Description**
This PR adds support for handling circular schemas.
- Adds a wrapper callback around all functionality which tries to recursively iterate schemas by reusing existing instances.
- Fixes a wrong interpretation of when to split out new objects for - ` test/processors/JsonSchemaInputProcessor/commonInputModel/combination_schemas.json`


src/simplification/Simplifier.ts
```diff
@@ -28,11 +35,19 @@ export default class Simplifier {
...
+        }
+      }
+      return false;
+    };
     if(simplifiedModel.length > 0){
       //Get the root model from the simplification process which is the first element in the list
       const schemaSimplifiedModel = simplifiedModel[0];
       //Only if the schema is of type object and contains properties, split it out
-      if(schemaSimplifiedModel.type !== undefined && schemaSimplifiedModel.type.includes("object") && schemaSimplifiedModel.properties !== undefined){
+      if(schemaSimplifiedModel.type !== undefined && schemaSimplifiedModel.type.includes("object") && !containsAllTypes(schemaSimplifiedModel)){
```
jonaslagoni, line 50: This actually solves a wrong interpretation for `test/processors/JsonSchemaInputProcessor/commonInputModel/combination_schemas.json` where object is not split out correctly.

src/simplification/SimplifyItems.ts
```diff
@@ -1,67 +1,72 @@
...
-  let models : CommonModel[] | undefined;
-  if(typeof schema !== "boolean"){
-    const addToModels = (model : CommonModel[] = []) => { models = [...(models || []), ...model]; }
-    const mergeWithItem = (model : CommonModel) => {
-      commonItems = CommonModel.mergeCommonModels(commonItems, model, schema);
-    }
-    const addToItemsAndModels = (out: output) => {
-      if(out.newModels !== undefined){
-          addToModels(out.newModels);
+export default function simplifyItems(input: Schema | boolean, simplifier : Simplifier) : Output {
```
magicmatatjahu, line 12: I understand why you make these callback in every function 😸 but it's not good for optimalization. When we'll have a lot of models in input, then it could be a problem. Maybe we should make callback outside the main function and then pass as one of argument `seenSchemas`, something like:

```ts
function callback(schema: Schema | boolean, simplifier : Simplifier, Map<any, Output> = new Map()) {
  ...logic
}

export default function simplifyItems(input: Schema | boolean, simplifier : Simplifier) : Output {
  const seenSchemas: Map<any, Output> = new Map();
  return callback(input, simplifier, seenSchemas);
}
```

We only compile one time `callback` function and it saves a lot of memory/cpu.

We should do it here and in other places with `callback` function. I don't know if it should be done in this PR or in the next one, but in my opinion we should optimize it sooner or later - better sooner 😄

magicmatatjahu (commented):
Great work! Two comments :) I don't approve and request, I want to start discussion.

src/simplification/Simplifier.ts
```diff
@@ -28,11 +35,19 @@ export default class Simplifier {
   simplifyRecursive(schema : Schema | boolean) : CommonModel[] {
     let models : CommonModel[] = [];
     let simplifiedModel = this.simplify(schema);
+    const containsAllTypes = (model : CommonModel) => {
+      if(model.type !== undefined){
+        if(Array.isArray(model.type)){
+          return model.type.length === 6;
```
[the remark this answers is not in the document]
jonaslagoni: Honestly It might need its own utility function with docs, will fix it 👍

src/simplification/SimplifyItems.ts
jonaslagoni: Lets do that yea! would also make it more clear and it should be split up more as the code evolve probably 👍

jonaslagoni: Why not remove the function completely and change it from: 

```ts
export default function simplifyItems(input: Schema | boolean, simplifier : Simplifier) : Output { 
```

to

```ts
export default function simplifyItems(input: Schema | boolean, simplifier : Simplifier, seenSchemas: Map<any, Output> = new Map()) : Output { 
```

😄

src/simplification/Simplifier.ts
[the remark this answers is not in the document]
magicmatatjahu: and also someone can make union type with 6 types :D So we should check that array contains number, string, array, boolean, null, object.

src/simplification/SimplifyItems.ts
magicmatatjahu: Yes, this second is better :)

src/models/Schema.ts
```diff
@@ -51,74 +51,80 @@ export class Schema extends CommonSchema<Schema | boolean>{
...
      * @returns CommonModel instance of the object
      */
-    static toSchema(object: Object) : Schema | boolean{
-        if(typeof object === "boolean") return object;
-        let schema = new Schema();
-        schema = Object.assign(schema, object);
-        schema = CommonSchema.transformSchema(schema, Schema.toSchema) as Schema;
+    static toSchema(object: Object): Schema | boolean {
+        const seenSchemas: Map<any, Schema> = new Map();
+        const callback = (object: Object): Schema | boolean => {
```
magicmatatjahu, line 56: Also here we can remove callback and add `seenSchemas` as second argument of `toSchema` function

jonaslagoni: Good catch!

src/simplification/Simplifier.ts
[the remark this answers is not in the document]
jonaslagoni: This function will change as soon as I do a bit of refactoring tomorrow :)

jonaslagoni:
@magicmatatjahu we need to force the indentation in eslint, this is getting out of hand with indentation.

magicmatatjahu (approved):

PR #56, opened 2021-02-12
chore: ensure unit test independance for SimplifyAdditionalProperties

jonaslagoni:
**Description**
This PR introduces mocking to SimplifyAdditionalProperties unit test to ensure there are no ripple effects when output changes from other modules.


magicmatatjahu (approved):

src/simplification/Utils.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * check if CommonModel is a separate model or a simple model.
+ */
+export function isModelObject(model: CommonModel) : boolean {
+  // This check should be done instead, needs a refactor to allow it though:
+  // this.extend !== undefined || this.properties !== undefined
+  if (model.type !== undefined) {
+    // If all possible JSON types are defined, don't split it even if it does contain object.
+    if (Array.isArray(model.type) && model.type.length === 6) {
```
magicmatatjahu, line 11: Maybe this is time ti change that? 😄

jonaslagoni: #59 addresses this if merged, with the current implementation this is not possible because of the recursive nature where each CommonModel does not have access to those properties yet 😄

PR #59, opened 2021-02-12
chore: simplify model splitting 

jonaslagoni:
**Description**
The current simplifier code is a bit weird because it in its recurse path tries to split out new models into separate ones. This behavior can be tricky to understand. This refactor splits models at the end where the model have already been simplified. 

This PR introduces:
1. Removes the need for separate simplified functions to return new models on the go.
2. Fixes a bug in the simplification of extend.

Blocked by unit test separation #58 #57 #56


src/simplification/Simplifier.ts
```diff
@@ -16,52 +16,27 @@ export class Simplifier {
   options: SimplificationOptions;
   anonymCounter = 1;
   seenSchemas: Map<Schema, CommonModel> = new Map();
-
+  iteratedModels: {[key: string]: CommonModel} = {};
```
magicmatatjahu, line 19: `iteratedModels: Record<string, CommonModel> = {};`

src/simplification/Simplifier.ts
```diff
@@ -73,57 +48,80 @@ export class Simplifier {
...
 
       const required = simplifyRequired(schema);
       if (required !== undefined) {
         model.required = required;
       }
     }
+    this.ensureModelsAreSplit(model);
+    const modelsToReturn = Object.values(this.iteratedModels);
+    //Add models which have not been iterated before
+    if (isModelObject(model) && this.iteratedModels[`${model.$id}`] === undefined) {
```
magicmatatjahu, line 83: For what you use string template, for avoid solar errors?

src/simplification/Simplifier.ts
```diff
@@ -73,57 +48,80 @@ export class Simplifier {
...
-        }
+        model.enum = enums;
       }
 
       const required = simplifyRequired(schema);
       if (required !== undefined) {
         model.required = required;
       }
     }
+    this.ensureModelsAreSplit(model);
```
magicmatatjahu, line 80: You should return from this function `iteratedModels`, why? As I see in other simplifier functions, you pass `this` value and then perform again `simplify` function, so you write data to `iteratedModels` in every execution of `simplify` function and it could duplicate models in output, am I right?

src/simplification/Simplifier.ts
magicmatatjahu: also, you should remove it and return this data from `ensureModelsAreSplit` function.

src/simplification/SimplifyProperties.ts
```diff
@@ -2,7 +2,7 @@
 import { CommonModel } from '../models/CommonModel';
 import { Schema } from '../models/Schema';
 import { Simplifier } from './Simplifier';
-type Output = { newModels: CommonModel[] | undefined; properties: { [key: string]: CommonModel } | undefined };
+type Output = { [key: string]: CommonModel } | undefined;
```
magicmatatjahu, line 5: `type Output = Record<string, CommonModel>| undefined;`

src/simplification/Simplifier.ts
```diff
@@ -73,57 +48,80 @@ export class Simplifier {
...
+    */
+    const splitModels = (model: CommonModel) : CommonModel => {
+      if (isModelObject(model)) {
+        const switchRootModel = new CommonModel();
+        switchRootModel.$ref = model.$id;
+        this.iteratedModels[`${model.$id}`] = model;
+        return switchRootModel;
+      }
+      return model;
+    };
```
magicmatatjahu, line 110: This function should be moved to outside of parent function but in `Simplifier` scope.

```ts
private ensureModelsAreSplit(model: CommonModel) { ... }

/**
    * This function splits up a model if needed and add the new model to the list of models.
    * 
    * @param model check if it should be split up
    * @param models which have already been split up
    */
    const splitModels = (model: CommonModel): CommonModel => {
      if (isModelObject(model)) {
        const switchRootModel = new CommonModel();
        switchRootModel.$ref = model.$id;
        this.iteratedModels[`${model.$id}`] = model;
        return switchRootModel;
      }
      return model;
    };
```

magicmatatjahu (changes requested):
Good! But you must do some improvements :)

src/simplification/Simplifier.ts
jonaslagoni: Yes :)

src/simplification/Simplifier.ts
jonaslagoni: No, because as soon as a model have been split it cannot be split again by the nature of the algorithm 😄

src/simplification/Simplifier.ts
jonaslagoni: I cant, since in the end I want to return all the models even from the recursive calls. Thats why this map is here to ensure that all separate models are kept and returned

magicmatatjahu (approved):

PR #61, opened 2021-02-15
chore: add required simplification

magicmatatjahu:
**Description**

Add simplification process for `required` field.


README.md
```diff
@@ -12,4 +12,5 @@ We have split out the different simplification processes into separate small fun
 - [Simplification of enums](./docs/SimplifyEnums.md)
 - [Simplification of items](./docs/SimplifyItems.md)
 - [Simplification of properties](./docs/SimplifyProperties.md)
-- [Simplification of additional properties](./docs/SimplifyAdditionalProperties.md)
+- [Simplification of properties](./docs/SimplifyAdditionalProperties.md)
```
jonaslagoni, line 15: This should be additional properties

src/simplification/SimplifyRequired.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type Output = string[] | undefined;
+
+/**
+ * Find the required array for a simplified version of a schema
+ * 
+ * @param schema to find the simplified required array for
+ * @param seenSchemas already seen schemas, this is to avoid circular schemas
+ */
+export default function simplifyRequired(schema: Schema | boolean, seenSchemas: Set<any> = new Set()): Output {
```
jonaslagoni, line 11: Just gonna mention it, that others might find it weird that we all of the sudden use a `Set` over map like the other simplification rules, also in regard to you returning undefined instead of the previous already simplified property. Not saying one is better then the other, but should this be the same across the board? 🤔

test/simplification/required.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+      "else": {
+          "properties": { 
+            "testProp3": { 
+              "type": "string"
+            } 
+          },
+          "required": ["testProp3"],
+      }
+    }
```
jonaslagoni, line 55: Just a suggestion, but I find it more easy to read both test wise and input if the larger schemas reside in an external file.

src/models/Schema.ts
```diff
@@ -40,7 +40,7 @@ export class Schema extends CommonSchema<Schema | boolean> {
     contentMediaType?: string; //Enum?
     definitions?: { [key: string]: Schema | boolean; };
     description?: string;
```
jonaslagoni, line 42: Did you remove the `required?: string[];` from schema?

jonaslagoni (changes requested):

src/models/Schema.ts
magicmatatjahu: Nope, I only changed type for `default`.

src/simplification/SimplifyRequired.ts
magicmatatjahu: I don't need to return previous simplified property, because I only return array of values, not reference, due to this fact I changed Map to Set.

test/simplification/required.spec.ts
magicmatatjahu: but simplification of required is not complex like other simplification :) and also I don't want use mocks, but check output value.

src/simplification/SimplifyRequired.ts
magicmatatjahu: I added comment why I don't need Map :P

src/models/Schema.ts
jonaslagoni: My question goes towards, it should be removed since it is now part of the CommonSchema right?

src/simplification/SimplifyRequired.ts
jonaslagoni: That is what I mean 😄 If someone else then us wants contribute to the simplifier code and looks through the separate functions then all the others does one behavior and this one does it differently. Question is do we want the same behavior across the board or do we not care about that? 🤔

src/models/Schema.ts
magicmatatjahu: We shouldn't alter JSON Draft7 Schema type. I haven't a problem with `required` field inside schema.

jonaslagoni: But it inherits from CommonSchema which now has the property, same as all the other properties inside CommonSchema are no longer defined inside the Schema class 😄

src/simplification/SimplifyRequired.ts
magicmatatjahu: I think we should go with use case behaviour. In one place Set will be better, on another Map.

src/models/Schema.ts
magicmatatjahu: You're right.

src/simplification/SimplifyRequired.ts
jonaslagoni: Alright 👍

jonaslagoni (approved):
LGTM 👍

PR #65, opened 2021-02-17
chore: infer name for object schema

magicmatatjahu:
**Description**

- reflect names of `declarations`, `definitions`, `properties` by propertyName and save them in `x-modelgen-inferred-name`, then infer this name to `$id` CommonModel's property
- update tests

**Related issue(s)**
See also https://github.com/asyncapi/shape-up-process/issues/45


src/simplification/Simplifier.ts
```diff
@@ -64,14 +65,19 @@ export class Simplifier {
...
         model.$id = schemaId;
       } else if (schema.$id !== undefined) {
         model.$id = schema.$id;
       }
 
+      const name = simplifyName(schema, model.$id);
+      if (name) {
+        model.name = name;
+      }
+
```
jonaslagoni, line 80: What is the different between model name and model id? 🤔 Why have both?

test/blackbox/AsyncAPI.spec.ts
```diff
@@ -1,7 +1,8 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import { JsonSchemaInputProcessor } from '../../src/processors/JsonSchemaInputProcessor';
-describe('AsyncAPI JSON Schema file', function() {
+
+describe.skip('AsyncAPI JSON Schema file', function() {
```
jonaslagoni, line 5: This needs to be unskipped

test/processors/AsyncAPIInputProcessor.spec.ts
```diff
@@ -3,7 +3,7 @@ import * as path from 'path';
 import {parse} from '@asyncapi/parser';
 import {AsyncAPIInputProcessor} from '../../src/processors/AsyncAPIInputProcessor'
 
-describe('AsyncAPIInputProcessor', function() {
+describe.skip('AsyncAPIInputProcessor', function() {
```
jonaslagoni, line 6: This needs to be unskipped

jonaslagoni (changes requested):
Most of the things looks good I however got a few things that needs to be addressed first:
- Does it really make sense to have `originalSchema` anymore? since it is no longer the original but a modified version?
- Does it make sense to have  `$id` as part of the CommonModel anymore since your new name kinda replaces this right? This also means you can remove the anonymous schema id naming functionality.
- Skipped tests need to be unskipped
- I would love to see unit tests for this function instead of relying on integration tests to show it.

src/simplification/Simplifier.ts
magicmatatjahu: > Does it make sense to have $id as part of the CommonModel anymore since your new name kinda replaces this right? 

We had a discussion and as I remember we decided that we should have `$id` as a separate field and name should be written in a wrapper of Schema (which I didn't do because it creates an additional wrapper for one field and it creates not needed complexity).

> This also means you can remove the anonymous schema id naming functionality.

Nope, anonymous is needed, when you have allOf, oneOf etc schemas -> without title or separate extension for naming you cannot name this schema.

jonaslagoni: 🤔 The reason why it should be moved to another class wrapper for Schema was because we  could keep our originalSchema as is. IMO this naming functionality should completely remove our current very basic anonymous naming including allOf/oneOf etc and infer naming of models more correctly. I dont see why we need to have both?

magicmatatjahu:
@jonaslagoni Thanks for review!

> Does it really make sense to call it originalSchema anymore? since it is no longer the original but a modified version?

I know, but we have an architectural problem. As you know we have this problem with additional iteration in `toSchema` function, and I don't know why. If I add name extension after `toSchema`, then I had a problem naming, because in one test as I remember in extension I had  a `test` name and should be `street_name` etc.

Also extension shouldn't be treated as changing schema -> this same we make inside our parser-js.

src/simplification/Simplifier.ts
magicmatatjahu: > IMO this naming functionality should completely remove our current very basic anonymous naming including allOf/oneOf etc and infer naming of models more correctly. I dont see why we need to have both?

How do you want to avoid anonymSchema when you have an object model inside oneOf, and then infer this model to `models` array?

jonaslagoni: One example is to use the path as the name - similar to what is suggested in https://github.com/asyncapi/parser-js/issues/176 

Pretty sure we discussed this as well where I said you should be careful about using the property names alone as names. I actually thought I made an issue here in the rep but I guess we took it in slack 🤔

test/blackbox/AsyncAPI.spec.ts
magicmatatjahu: AsyncAPI should be skipped, because we don't test it 😄

magicmatatjahu:
@jonaslagoni I removed `name` and save name to `$id` and also added missing test for new function.

test/blackbox/AsyncAPI.spec.ts
jonaslagoni: I mean, running the file through the processor is kinda a test in it self, but fair 😄

jonaslagoni: Ahh you added expect, nwm 😄

test/processors/JsonSchemaInputProcessor.spec.ts
```diff
@@ -123,4 +124,79 @@ describe('JsonSchemaInputProcessor', function() {
...
+                    },
+                    {
+                        type: "string",
+                    },
+                ]
+            }
+            const expected = JsonSchemaInputProcessor.reflectSchemaName(schema) as any;
+
+            // properties
+            expect(expected.properties.prop['x-modelgen-inferred-name']).toEqual('prop');
```
jonaslagoni, line 182: this is gonna be a problem if the same property name is used in multiple places, I would not recommend to use the prop name as is but rather path to that property. For example in this case I would do `RootPropertiesProp`. 

Or would you rather handle this in the simplifier? Not sure that is better.. What would the id then be? `propX`?

test/processors/JsonSchemaInputProcessor.spec.ts
```diff
@@ -123,4 +124,79 @@ describe('JsonSchemaInputProcessor', function() {
...
+            expect(expected.properties.allOfCase.allOf[0]['x-modelgen-inferred-name']).toEqual('allOfCase_allOf_0');
+            expect(expected.properties.allOfCase.allOf[1]['x-modelgen-inferred-name']).toEqual('allOfCase_allOf_1');
+
+            // patternProperties
+            expect(expected.patternProperties.patternProp['x-modelgen-inferred-name']).toEqual('patternProp');
+
+            // dependencies
+            expect(expected.dependencies.dep['x-modelgen-inferred-name']).toEqual('dep');
```
jonaslagoni, line 190: Same as properties.

test/processors/JsonSchemaInputProcessor.spec.ts
```diff
@@ -123,4 +124,79 @@ describe('JsonSchemaInputProcessor', function() {
...
+            }
+            const expected = JsonSchemaInputProcessor.reflectSchemaName(schema) as any;
+
+            // properties
+            expect(expected.properties.prop['x-modelgen-inferred-name']).toEqual('prop');
+            expect(expected.properties.allOfCase.allOf[0]['x-modelgen-inferred-name']).toEqual('allOfCase_allOf_0');
+            expect(expected.properties.allOfCase.allOf[1]['x-modelgen-inferred-name']).toEqual('allOfCase_allOf_1');
+
+            // patternProperties
+            expect(expected.patternProperties.patternProp['x-modelgen-inferred-name']).toEqual('patternProp');
```
jonaslagoni, line 187: Are you sure you would use regex as a name?

test/processors/JsonSchemaInputProcessor.spec.ts
jonaslagoni: Also, with the current implementation of property name inference you would have two schemas with the same name.

jonaslagoni (changes requested):
So my requirement for this infer name is that name is unique for all schemas, this might be wrong, but then I dont see how we can use this in replacement of the anon schema naming?

test/processors/JsonSchemaInputProcessor.spec.ts
magicmatatjahu: Nope, I also handle such a case. I will update test.

test/processors/JsonSchemaInputProcessor.spec.ts
magicmatatjahu: Good point!

test/processors/JsonSchemaInputProcessor.spec.ts
magicmatatjahu: Nope, I also handle such a case. I will update test.

test/processors/JsonSchemaInputProcessor.spec.ts
```diff
@@ -197,6 +211,7 @@ describe('JsonSchemaInputProcessor', function() {
             // anyOf
             expect(expected.anyOf[0]['x-modelgen-inferred-name']).toEqual('anyOf_0');
             expect(expected.anyOf[1]['x-modelgen-inferred-name']).toEqual('anyOf_1');
+            expect(expected.anyOf[1].properties.prop['x-modelgen-inferred-name']).toEqual('anyOf_1_prop');
```
jonaslagoni, line 214: What about
```

properties: {
    prop: {
        type: "object",
        properties: {
            prop: {
                type: "object",
            }
        }
    },
```

jonaslagoni (changes requested):

test/processors/JsonSchemaInputProcessor.spec.ts
magicmatatjahu: It will be `prop_prop`

src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -74,14 +78,146 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
...
 
   /**
-     * Simplifies a JSON Schema into a common models
-     * 
-     * @param schema to simplify to common model
-     */
-  static convertSchemaToCommonModel(schema: Schema | boolean) : {[key: string]: CommonModel} {
+   * Reflect name from given schema and save it to `x-modelgen-inferred-name` extension.
+   * 
+   * It should be removed when we'll simplify current solution for processing schema to CommonModel.
```
jonaslagoni, line 83: What do you mean by this line?

src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -74,14 +78,146 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
...
+   * 
+   * @param schema to process
+   * @param namesStack
+   * @param name
+   * @param isRoot
```
jonaslagoni, line 88: need a definition of the arguments. `isRoot` is pretty obvious, but the other two could use a description IMO

jonaslagoni (changes requested):
Just have two comments, looking great overall!

src/processors/JsonSchemaInputProcessor.ts
magicmatatjahu: I don't remember. I wrote this in previous week 😄

src/processors/JsonSchemaInputProcessor.ts
magicmatatjahu: 🆗

docs/customization.md
```diff
@@ -0,0 +1,7 @@
+# Customization
+
+The AsyncAPI Model SDK uses for customization **presets**.
+
+## Preset
+
+Each **preset** is a JS object which extends the rendered model with additional content.
```
jonaslagoni, line 7: Why is this added in this pr? 😄

jonaslagoni (changes requested):

docs/customization.md
magicmatatjahu: Mistake 😄 I forgot to checkout branch and by mistake pushed new docs.

jonaslagoni (approved):
🥳

Issue #68, opened 2021-02-18
should render additionalProperties 


jonaslagoni:
#### Reason/Context
Currently the `CommonModel` does include additionalProperties, however they are not rendered as properties in the generator 

Questions:
- How should we handle `additionalProperties` if we already have a property called `additionalProperties`?

magicmatatjahu:
@jonaslagoni You pinged me about problems with `addProperties` etc in slack, so I give you some tips here, also to check by another people from community:

For `additionalProperties`:
- in TS and JS we should render for additionalProperties additional accessor for set and get, but not native, but something like `getAdditionalProperty(key: string): {type}` and `setAdditionalProperty(key: string, value: {type}): void`.
- in Java we can make something similar `public {type} getAddtionalProperty(String key) { this.additionalProperties.get(key) }` and similar `setAdditionalProperty`.

For `patternProperties` situation is very similar to `additionalProperties`. We have an example:

```ts
{
  "type": "object",
  "patternProperties": {
    "^S_": { "type": "string" },
    "^I_": { "type": "integer" }
  },
  "additionalProperties": false
}
```

- in TS and JS we should render:

```ts
patternProperties: Record<string, any>;

getPatternProperty(key: string): {type | undefined} {
  return patternProperties[key];
}

setPatternProperty(key: string, value: any): void {
  // here we must check patterns
  if (/^S_/.test(key) && typeof value === "string") {
    this.patternProperties[key] = value;
  }
  if (/^I_/.test(key) && typeof value === "number") {
    this.patternProperties[key] = value;
  }
  // maybe throw error
}
```

- for Java very similar like for TS - additional get and set for patternProperties and inside them check pattern before saving data.

`additionalProperties` and `patternProperties` you can render in 2 ways:
- using `additionalContent` preset for class [here](https://github.com/asyncapi/generator-model-sdk/blob/master/src/generators/typescript/renderers/ClassRenderer.ts#L79) or
- adjust `renderProperties` in appropriate renderer

PR #71, opened 2021-02-18
chore: add additionalProperties preset hooks

jonaslagoni:
**Description**
This PR adds additionalProperties preset hooks.

**Related issue(s)**
solves #68


jonaslagoni:
@magicmatatjahu how should we handle additionalProperties in Javascript renderer? 🤔

jonaslagoni:
@magicmatatjahu I have fully added the typescript rendering can you check it out in regards to presets and the setup there? Also the class encapsulation for the additionProperties should they have presets?

src/models/Preset.ts
```diff
@@ -21,15 +21,21 @@ export interface PropertyArgs {
...
 }
 
+export interface AdditionalPropertiesArgs {
+  parentModel: CommonModel;
+  additionalProperties: CommonModel;
+}
 export interface ClassPreset<R extends AbstractRenderer, O extends object = any> extends CommonPreset<R, O> {
   ctor?: (args: PresetArgs<R, O>) => Promise<string> | string;
   property?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
+  additionalProperties?: (args: PresetArgs<R, O> & AdditionalPropertiesArgs) => Promise<string> | string;
```
magicmatatjahu, line 31: `additionalProperties` should have only `PresetArgs` arguments, why? `parentModel ` is not needed - it's my fault and in next PR I will remove it - we have `model` property for that (it is the same 😅). Also `additionalProperties` field is hardcoded for `additionalProperties` inferred from `model` so you have also possibility to use it by `model`. `additionalProperties` preset should have also possibility to render `patternProperties` etc, so hardcoded for only one field is very bad idea.

```ts
additionalProperties?: (args: PresetArgs<R, O>) => Promise<string> | string;
```

src/models/Preset.ts
```diff
@@ -21,15 +21,21 @@ export interface PropertyArgs {
...
   getter?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
   setter?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
 }
 
 export interface InterfacePreset<R extends AbstractRenderer, O extends object = any> extends CommonPreset<R, O> {
   property?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
+  additionalProperties?: (args: PresetArgs<R, O> & AdditionalPropertiesArgs) => Promise<string> | string;
```
magicmatatjahu, line 38:
```suggestion
  additionalProperties?: (args: PresetArgs<R, O>) => Promise<string> | string;
```

src/generators/typescript/TypeScriptRenderer.ts
```diff
@@ -87,6 +92,10 @@ ${lines.map(line => ` * ${line}`).join('\n')}
     return `${name}${signature};`;
   }
 
+  async runAdditionalPropertiesPreset(additionalProperties: CommonModel, parentModel: CommonModel): Promise<string> {
+    return this.runPreset('additionalProperties', { additionalProperties, parentModel });
+  }
+
```
magicmatatjahu, line 98:
```suggestion
  async runAdditionalPropertiesPreset(): Promise<string> {
    return this.runPreset('additionalProperties');
  }

```

src/generators/typescript/TypeScriptRenderer.ts
```diff
@@ -77,6 +77,11 @@ ${lines.map(line => ` * ${line}`).join('\n')}
       const rendererProperty = await this.runPropertyPreset(propertyName, property, this.model);
       content.push(rendererProperty);
     }
+    
+    if (this.model.additionalProperties !== undefined && this.model.additionalProperties instanceof CommonModel) {
+      const additionalProperty = await this.runAdditionalPropertiesPreset(this.model.additionalProperties, this.model);
```
magicmatatjahu, line 82: `additionalProperties` preset should also render another properties like patternProperties - also passed by options to preset - so this part should look like:

```ts
// from this
if (this.model.additionalProperties !== undefined && this.model.additionalProperties instanceof CommonModel) {
  const additionalProperty = await this.runAdditionalPropertiesPreset(this.model.additionalProperties, this.model);
  content.push(additionalProperty);
}

// to 
const additionalProperties = await this.runAdditionalPropertiesPreset(this.model.additionalProperties, this.model);
if (additionalProperties) {
  content.push(additionalProperties);
}
```

we shouldn't hardcoded only for `additionalProperties` from model.

src/generators/javascript/renderers/ClassRenderer.ts
```diff
@@ -36,6 +36,12 @@ ${this.indent(this.renderBlock(content, 2))}
       content.push(this.renderBlock([getter, setter]));
     }
 
+    if (this.model.additionalProperties !== undefined && this.model.additionalProperties instanceof CommonModel) {
+      const getter = 'getAdditionalProperty(key){ return _additionalProperties[key]; }';
+      const setter = 'setAdditionalProperty(key, value) { _additionalProperties[key] = value; }';
+      content.push(this.renderBlock([getter, setter]));
+    }
+
```
magicmatatjahu, line 44: it should be rendered inside `additionalContent` preset:

```ts
{
  class: {
    additionalContent({ renderer, model, content }) {
      if (!(this.model.additionalProperties instanceof CommonModel)) return content;
    
      const getter = 'getAdditionalProperty(key){ return _additionalProperties[key]; }';
      const setter = 'setAdditionalProperty(key, value) { _additionalProperties[key] = value; }';
      this.renderBlock([content, getter, setter]);
     }
  }
}
```

src/generators/javascript/JavaScriptRenderer.ts
```diff
@@ -36,6 +36,11 @@ ${content}
       content.push(rendererProperty);
     }
 
+    if (this.model.additionalProperties !== undefined && this.model.additionalProperties instanceof CommonModel) {
```
magicmatatjahu, line 39: Please add it like in TS -> to `additionalContent` preset

src/generators/java/renderers/ClassRenderer.ts
```diff
@@ -52,6 +52,12 @@ ${this.indent(this.renderBlock(content, 2))}
       content.push(this.renderBlock([getter, setter]));
     }
 
+    if (this.model.additionalProperties !== undefined && this.model.additionalProperties instanceof CommonModel) {
```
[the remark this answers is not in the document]
magicmatatjahu: Like in TS -> to `additionalContent` preset

magicmatatjahu (changes requested):
Good, but I have an another idea how to use `additionalProperties` :) It's a more generic.

src/generators/typescript/TypeScriptRenderer.ts
magicmatatjahu: We don't need additional arguments in `additionalProperties` preset :)

src/generators/java/renderers/ClassRenderer.ts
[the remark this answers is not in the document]
jonaslagoni: What do you mean by this? I don't use `additionalContent` in TypeScript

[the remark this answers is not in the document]
magicmatatjahu: I mean that this part should be moved to `additionalContent` preset like I suggest in another comment for TS :)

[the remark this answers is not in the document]
jonaslagoni: I guess you mean this for JS - https://github.com/asyncapi/generator-model-sdk/pull/71#discussion_r581883512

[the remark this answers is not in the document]
magicmatatjahu: Yeah, sorry 😄

src/generators/javascript/renderers/ClassRenderer.ts
jonaslagoni: I think you mean the extra functions that are not really encapsulations right? 🤔

magicmatatjahu: Yes, for that `additionalContent` exists - for extra function. Encapsulation is a problem, but you can create separate presets to make your encapsulation

```ts
new TypeScriptGenerator({ presets: [ADDITIONALPROPETIES, PATTERN_PROPERTES] })
```

magicmatatjahu: We can make separate presets for everything, argument, type signature etc. but then it would be hard to use.

jonaslagoni:
Gonna provide a clean implementation here, gonna use most of the comments already given to solve it. Thanks for the review @magicmatatjahu 🙇

Issue #53, opened 2021-02-11
Remove Eslint warnings


jonaslagoni:
#### Reason/Context
Currently we ignore warnings but it needs to be addressed. Especially the complexity check. After that we need to force no warnings.

magicmatatjahu:
Partially done in https://github.com/asyncapi/generator-model-sdk/pull/60

jonaslagoni:
Completely done in https://github.com/asyncapi/generator-model-sdk/pull/81

PR #72, opened 2021-02-18
chore: removes eslint warnings for enums simplificiation

jonaslagoni:
**Description**
Removes complexity of enum simplification.

```
/Users/lagoni/Documents/AsyncAPI/generator-model-sdk/src/simplification/SimplifyEnums.ts
  10:25  warning  Refactor this function to reduce its Cognitive Complexity from 28 to the 15 allowed  sonarjs/cognitive-complexity
```

**Related issue(s)**
part of #53


src/simplification/SimplifyEnums.ts
```diff
@@ -9,56 +9,84 @@ type Output = any[] | undefined;
...
+}
+
+/**
+ * Ensure enums in not are never included.
+ * 
+ * @param schema currently searching in
+ * @param existingEnums which have already been found
+ * @param seenSchemas already seen schemas and their respectable output
+ */
+function ensureNotEnumsAreRemoved(schema: Schema, existingEnums: any[], seenSchemas: Map<any, Output>) {
```
magicmatatjahu, line 51: `ensureNotEnums` is enought :)

magicmatatjahu (approved):

src/simplification/SimplifyEnums.ts
```diff
@@ -9,56 +9,84 @@ type Output = any[] | undefined;
  */
 export default function simplifyEnums(schema: Schema | boolean, seenSchemas: Map<any, Output> = new Map()): Output {
   if (typeof schema !== 'boolean') {
```
magicmatatjahu, line 11: Also I think that we should go with this approach with handling boolean schemas:

```ts
if (typeof schema === 'boolean') {
  return udnefined;
}

// ...rest of logic
```

For me this is better to read and understand. Of course I can do that in follow up PR after yours.

jonaslagoni: If we want consistency I think we should do that in another PR yes 😄

src/simplification/SimplifyEnums.ts
jonaslagoni: True 😄

PR #76, opened 2021-02-19
chore: removes eslint warnings from Simplifier

jonaslagoni:
**Description**
Removes eslint warnings from simplifier.

```
/Users/lagoni/Documents/AsyncAPI/generator-model-sdk/src/simplification/Simplifier.ts
  32:3  warning  Refactor this function to reduce its Cognitive Complexity from 25 to the 15 allowed  sonarjs/cognitive-complexity
```

**Related issue(s)**
related to #53


src/simplification/Simplifier.ts
```diff
@@ -86,6 +50,52 @@ export class Simplifier {
     return [model, ...modelsToReturn];
   }
 
+  /**
+   * Function to simplify all model properties from schema.
+   * 
+   * @param model to simplify properties to 
+   * @param schema to simplify
+   */
+  private simplifyModelProperties(model: CommonModel, schema: Schema) {
```
magicmatatjahu, line 59: I think that `simplifyModel` is enough. By `simplifyModelProperties` I have in mind `object` type and this is not a true - we also simplify here items, enums etc.

magicmatatjahu (changes requested):

src/simplification/Simplifier.ts
jonaslagoni: Agreed 😄

magicmatatjahu (approved):

PR #78, opened 2021-02-19
chore: removes eslint warnings from item simplification

jonaslagoni:
**Description**
Refactors the item simplification to remove eslint warnings.

- This reduced the number of calls to merge since we make more checks in regards to when to call the function.
- Adapted tests to always return undefined because we mock functions

**Related issue(s)**
related to #53


magicmatatjahu (dismissed):

jonaslagoni:
So the error occurred with circular schemas being returned twice. So I added a check and removed the model if it already existed (since it will be re-added right)

src/simplification/SimplifyItems.ts
```diff
@@ -12,50 +13,66 @@ type Output = CommonModel | undefined;
...
+ * Go through schema(s) and combine the simplified items together.
+ * 
+ * @param schema to go through
+ * @param currentModel the current output
+ * @param simplifier the simplifier to use
+ * @param seenSchemas schemas which we already have outputs for
+ */
+function combineSchemas(schema: (Schema | boolean) | (Schema | boolean)[] | undefined, currentModel: Output, simplifier : Simplifier, seenSchemas: Map<any, Output>) {
+  if (schema === undefined) return;
+  if (typeof schema === 'boolean') return;
```
magicmatatjahu, line 56:
```suggestion
  if (typeof schema !== 'object') return;
```

src/simplification/SimplifyItems.ts
```diff
@@ -12,50 +13,66 @@ type Output = CommonModel | undefined;
  * @param seenSchemas already seen schemas and their corresponding output, this is to avoid circular schemas
  */
 export default function simplifyItems(schema: Schema | boolean, simplifier : Simplifier, seenSchemas: Map<any, Output> = new Map()) : Output {
-  let tempOutput: Output;
   if (typeof schema !== 'boolean') {
-    if (seenSchemas.has(schema)) return seenSchemas.get(schema)!;
+    if (seenSchemas.has(schema)) return seenSchemas.get(schema);
+    const tempOutput: Output = new CommonModel();
```
magicmatatjahu, line 18: Why `tempOutput` name?

src/simplification/SimplifyItems.ts
```diff
@@ -12,50 +13,66 @@ type Output = CommonModel | undefined;
...
-        mergeWithItem(out);
-      }
-    };
-    const handleCombinationSchemas = (schemas: (Schema | boolean)[] = []) => {
-      schemas.forEach((itemSchema) => {
-        addToItems(simplifyItems(itemSchema, simplifier, seenSchemas));
-      });
-    };
 
+    //Check if any items have been sat, if so lets merge them
```
magicmatatjahu, line 21:
```suggestion
    //Check if any items have been set, if so lets merge them
```

src/simplification/Simplifier.ts
```diff
@@ -44,9 +44,13 @@ export class Simplifier {
       this.simplifyModel(model, schema);
     }
     this.ensureModelsAreSplit(model);
+    //Ensure current model is not part of the iterated list since we could have circular schemas
+    if (this.iteratedModels[`${model.$id}`] !== undefined) {
+      delete this.iteratedModels[`${model.$id}`];
+    }
```
magicmatatjahu, line 50: Can I trust you? With next PRs, for me logic is more and more magic 😄

magicmatatjahu (changes requested):
Good job! I left two suggestions and two questions :)

src/simplification/SimplifyItems.ts
jonaslagoni: 🤷 you don't like it?  😄

src/simplification/SimplifyItems.ts
jonaslagoni: `schema` can be undefined :smile:

magicmatatjahu: but then you have also false for undefined :P

src/simplification/Simplifier.ts
jonaslagoni: Never trust anyone 😈 

The problem is at line 51 we create a copy of all the values of the `iteratedModels` and if we have a circular schema that we would return the same model twice `return [model1, ...[model2, model1]];`

Unfortunately, I don't see any other way around this based on the nature of JSON Schema 😟

src/simplification/SimplifyItems.ts
magicmatatjahu: `temp` prefix is always used in small function and not for returned values, here you return this var. If you have no idea for a different name, leave it :)

src/simplification/SimplifyItems.ts
jonaslagoni: AHH, magic 🔥

src/simplification/SimplifyItems.ts
jonaslagoni: True, I will change it.

src/simplification/Simplifier.ts
magicmatatjahu: I understand the logic, but it seems more and more complex and I see that it (probably) will cause a lot of errors in the future. We must to think about another way to handle scheme processing.

jonaslagoni: The only other way to do this is by not using recursion and that would be a rewrite from the ground up and basically reengineered 😄

magicmatatjahu: Yeah, by `We must to think about another way to handle scheme processing` I mean reengineered. What we have learned is ours 😄

magicmatatjahu (approved):

Issue #43, opened 2021-02-10
Iterate pattern properties


jonaslagoni:
#### Reason/Context
<s>Pattern properties should be merged with additional properties</s>
Pattern properties should be included the same way that additionalProperties are. Meaning it should be moved from Schema to CommonSchema.

jonaslagoni:
There are different ways to achieve this so given the following JSON Schema:
```
{
  "type": "object",
  "properties": {
    "builtin": { "type": "number" }
  },
  "patternProperties": {
    "^S_": { "type": "string" },
    "^I_": { "type": "integer" }
  },
  "additionalProperties": { "type": "string" }
}
```
should it generate a model (gonna use Java as an example here) such as:
```
public class ClassName {
  private builtin: number; 
  private Map<String, string> S_PatternProperties = new HashMap();
  private Map<String, int> I_PatternProperties = new HashMap();
  private Map<String, string> additionalProperties = new HashMap();
}
```
or
```
public class ClassName {
  private builtin: number; 
  private Map<String, Object> additionalProperties = new HashMap();
}
```

jonaslagoni:
Based on multiple discussion we should go with something like:
```
public class ClassName {
  private builtin: number; 
  private Map<String, string> S_PatternProperties = new HashMap();
  private Map<String, int> I_PatternProperties = new HashMap();
  private Map<String, string> additionalProperties = new HashMap();
}
```
Now we still have a problem with duplicate properties in cases where `properties` would contain a property called `S_PatternProperties` which we need to figure out how to handle.

jonaslagoni:
Gonna close this issue since we currently interpret patternProperties, all that is missing are https://github.com/asyncapi/modelina/issues/154

PR #80, opened 2021-02-19
chore: patternProperties should be included in the output

jonaslagoni:
**Description**
This PR ensures that patternProperties are included in the output. 

**Related issue(s)**
resolves #43


jonaslagoni:
This is a wrong implementation since it should be moved to CommonModel

Issue #67, opened 2021-02-18
Fix simplification of tuple 


magicmatatjahu:
#### Describe the bug

When I fixing rendering the tuple for TS, I found the bug.

At the moment given schema:

```json
{ "type": "array", "items": [{ "type": "string" }, { "type": "number" }] }
```

produces given CommonModel:

```ts
CommonModel {
  originalSchema: Schema { type: 'array', items: [ [Schema], [Schema] ] },
  type: [ 'string', 'number' ]
}
```

and this is not true, because it's a tuple, not an array of unions. We should fix it.

magicmatatjahu:
Also if I write something like:

```json
{ "type": "array", "items": { "oneOf": [ { "type": "string" }, { "type": "number" } ] } }
```

then this is true union, but I give this same CommonModel, like in tuple case, so user/dev cannot determine which type should be rendered for model ans must for that check the originalSchema, which is not acceptable.

jonaslagoni:
This might get fixed by an upcoming pr, there are a bug in the type simplifier.

jonaslagoni:
The simplifier does semi-require the root schema to be an object... I'll try spend a few hours to figure out how we can handle this better.

jonaslagoni:
Current implementation (with all the PR's merged) the CommonModel would look like 
```
{
  originalSchema: {
    type: "array",
    items: [
      {
        type: "string",
      },
      {
        type: "number",
      },
    ],
  },
  type: "array",
  items: {
    type: [
      "string",
      "number",
    ],
    originalSchema: {
      type: "array",
      items: [
        {
          type: "string",
        },
        {
          type: "number",
        },
      ],
    },
  },
}
```
Which is as expected right?

magicmatatjahu:
@jonaslagoni No, this is not a true, because this schema indicates that you have an array of union, not a tuple. We should have schema like this:

```ts
"items": [
  { type: "string" },
  { type: "number" },
]
```

magicmatatjahu:
We conclude that this a feature request, but also a bug :)

PR #82, opened 2021-02-19
chore: added tests for non-object roots

jonaslagoni:
**Description**
This PR adds test to the simplifier to ensure non-object roots function correctly

**Related issue(s)**
Related to #67


test/simplification/simplify.spec.ts
```diff
@@ -18,4 +18,22 @@ describe('Simplification', function() {
...
   });
+
+  test('should support array roots', function() {
+    const actualModels = simplify({ "type": "array", "items": [{ "type": "string" }, { "type": "number" }] });
+    const expectedSchemaString = fs.readFileSync(path.resolve(__dirname, './simplify/expected/array_root.json'), 'utf8');
+    const expectedModel = JSON.parse(expectedSchemaString);
+    expect(actualModels).not.toBeUndefined();
+    expect(actualModels[0]).toEqual(expectedModel);
+  });
+  test('should support string roots', function() {
```
magicmatatjahu, line 29:
```suggestion
  test('should support primitive roots', function() {
```

test/simplification/simplify/expected/array_root.json
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    "originalSchema": {
+      "type": "array",
+      "items": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "number"
+        }
```
magicmatatjahu, line 27: This is not a true, because this schema indicates that you have an array of union, not a tuple. We should have schema like this: 

```js
"items": [
  { type: "string" },
  { type: "number" },
]
```

magicmatatjahu (changes requested):
Sorry, I found a bug 😄

test/simplification/simplify/expected/array_root.json
magicmatatjahu: Tuple will be integrated as feature in the future.

magicmatatjahu (approved):

magicmatatjahu (approved):

PR #87, opened 2021-02-22
docs: ensure doc consistancy and correctness

jonaslagoni:
**Description**
This PR re-writes some of the documentation for how to the library works.


jonaslagoni:
@smoya and @derberg I really want your opinion on the docs and if anything needs to be changed as they are now 😄

docs/input processing.md
```diff
@@ -0,0 +1,40 @@
+# Input processing
+The input process is about processing any input into our internal model representation `CommonInputModel`. This most likely have to be done using different methods based on which kind of input it is.
+
+As of now two inputs are supported:
+- JSON Schema Draft 7
+- AsyncAPI version 2.0.0
```
fmvilas, line 6: Is it AsyncAPI (the whole file) or AsyncAPI Schema Object?

README.md
```diff
@@ -40,24 +40,21 @@ Run this command to install the SDK in your project:
...
-The transformation process starts from checking what type of input data is provided and based on that it is processed differently. Currently JSON Schema Draft 7 and AsyncAPI version 2.0.0 are supported. It is also in this stage that references are resolved and schemas are named. Each input data must be converted to JSON Schema or be a subset of JSON Schema (like AsyncAPI spec) to go to the next process - the simplification stage.
+The process of creating data models from input data consists of 2 processes, the input and generation process.
 
-### The simplification process
+### The input process
 
-In order to simplify the model generation process as much as possible, AsyncAPI Model SDK simplifies the transformed JSON schema into `CommonModel`s, which contain information about the type of model it is, what properties it might have, etc, so finally we have a bare minimum schema. Think of it like converting data validation rules (JSON Schema) to a data definition.
+The input process ensures that any supported input are handles correctly, the basics is that any input needs to be converted into our internal model representation `CommonModel`. The following inputs which are supported:
```
smoya, line 49: Few lang typos here:
```suggestion
The input process ensures that any supported input is handled correctly, the basics are that any input needs to be converted into our internal model representation `CommonModel`. The following inputs are supported:
```

I would like to suggest using https://app.grammarly.com so most of those will be fixed. I'm not a native English speaker either so I usually do the same mistakes.

docs/input processing.md
jonaslagoni: The whole file. At the moment you cannot say `.payload()` directly, you need the underlying JSON schema for it to be processed - `.payload()._json`

README.md
jonaslagoni: Thanks for the review 🙇 It is now installed just to catch the most obvious errors 😄

README.md
```diff
@@ -102,6 +99,25 @@ interface Address {
   state?: "Texas" | "Alabama" | "California" | "other";
 }
 ```
+## Supported input
+These are the supported inputs.
+### AsyncAPI input
+The library expects the `asyncapi` property for the document to be sat in order to understand the input format.
+- Generate from a [parsed AsyncAPI document](https://github.com/asyncapi/parser-js)
```
smoya, line 106: Is there any other option for parsing rather than JS? If so, would it be worth mentioning a few or just indicate the JS is an example of them?.

README.md
```diff
@@ -102,6 +99,25 @@ interface Address {
   state?: "Texas" | "Alabama" | "California" | "other";
 }
 ```
+## Supported input
+These are the supported inputs.
+### AsyncAPI input
+The library expects the `asyncapi` property for the document to be sat in order to understand the input format.
```
smoya, line 105:
```suggestion
The library expects the `asyncapi` property for the document to be set in order to understand the input format.
```

README.md
magicmatatjahu: @smoya `model-sdk` is written in JS and will be used in generator (also written in JS), so at the moment we cannot use parsed schema from another parser, I mean written in another language.

magicmatatjahu: Only solution to use package with parser written in another language is first parse schema, save parsed schema to file, and then read content from file and put in our `model-sdk`.

docs/input processing.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Internal model representation
+
+<img src="./assets/class_diagram.png"
+     alt="Class diagram" />
+
+As seen on the class diagram the `InputProcessor` is our main point of entry for processing input data. It uses the defined other input processors (`AsyncAPIInputProcessor`, `JsonSchemaInputProcessor`, ...) by calling `shouldProcess` function of each and if the function returns true processing it using the `process` function. If no processes returns true it defaults to `JsonSchemaInputProcessor`.
```
smoya, line 13: I would suggest following https://sembr.org here. In this case, just by adding line breaks after periods everything becomes more readable for humans (especially for reviewing/editing).

docs/input processing.md
```diff
@@ -0,0 +1,40 @@
...
+- `$id` is the id/name of the model.
+- `type` this is the different types for the model. All types from JSON Schema are used, no custom types are used.
+- `enum` defines the different enums for the model.
+- `items` defines the type for `array` models as `CommonModel`.
+- `properties` defines the properties and its expected types as `CommonModel`.
+- `additionalProperties` are used to define if any extra properties are allowed, also defined as a  `CommonModel`.
+- `$ref` is a reference to another `CommonModel` by using`$id` as a simple string.
+- `required` list of required properties.
+- `extend` list of other `CommonModel`s this model extends, is an array of `$id` strings.
+- `originalSchema` is the actual input which this model represent.
```
smoya, line 27: I'm a bit concerned about the possibility of this becoming out of date earlier than expected. Maintaining this here rather than in code. Is there a way we can move all this to become always up-to-date with code generation? Perhaps documenting code?

docs/simplification.md
```diff
@@ -1,8 +1,14 @@
...
+The library simplify the JSON Schema from data validation rules to data definitions (`CommonModel`(s)). We do this because JSON Schema files can be extremely complex and can in many cases be simplified down to a bare minimum representation. Now keep in mind that there is a difference between validating input and the underlying structure of the data. In our case we do not care directly about the validation of input but rather how we display the underlying structure correctly.
 
-The AsyncAPI Model SDK supports simplification of:
+This document is expected to be a supplement to the code to better understand how the simplifier works.
+
+## Simplifier
+The main functionality is located in the `Simplifier` class. This class ensures to recursively create (or retrieve from a cache) a `CommonModel` representation of a Schema. We have tried to keep the the functionality split out into separate functions to reduce complexity and ensure it is easier to maintain. This main function also ensures to split any created models into separate ones if needed.
```
smoya, line 8: Minor: Little typo there

```suggestion
The main functionality is located in the `Simplifier` class. This class ensures to recursively create (or retrieve from a cache) a `CommonModel` representation of a Schema. We have tried to keep the functionality split out into separate functions to reduce complexity and ensure it is easier to maintain. This main function also ensures to split any created models into separate ones if needed.
```

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
 - [Required](#determining-the-required-properties-for-the-model)
 - [Extend](#determining-the-extend-for-the-model)
 
+<br/>
```
smoya, line 21: I'm not sure this line break provides any improvement here. Also, if needed, you could add another empty line here as per markdown standards.

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
+
 ## Determining the type for the model
 
-In order to determine all the possible types a schema can be, we both infer and use existing definitions of types, however we need to define a precedence for JSON Schema keywords for in which order they types are inferred or determined.
+In order to determine all the possible types a model can be based on schema, we both infer and use existing definitions of types, however we need to define a precedence for JSON Schema keywords for in which order they types are inferred or determined.
```
smoya, line 25: I think I understand what this sentence says, but took me a while to figure out. Would something like the following still mean the same you wanted to say?

```suggestion
In order to determine all the possible schema types a model can be based on, we both infer and use existing definitions of types. 
Precedence for JSON Schema keywords matters, meaning we need to evaluate them in a particular order.
```

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
 
-### Precedence
+**Precedence of JSON Schema keywords**
 
 The precedence of keywords are in which order we infer or determine types in. Notice it goes form left to right meaning first we set the type using the `type` keyword and then it moves right applying them one by one. The following are the precedence for determining types:
```
smoya, line 29: Most of this would be already mentioned in https://github.com/asyncapi/generator-model-sdk/pull/87/files#r580499801 so I think it can partially be removed.

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
 
-`type` --> `allOf` --> `oneOf` --> `anyOf` --> `then` --> `else` --> `enum` --> `const` --> `not`
+<p align="center">type --> allOf --> oneOf --> anyOf --> then --> else --> enum --> const --> not</p>
```
smoya, line 31: Totally subjective: I saw how this is rendered and not pretty convinced it improves readability in my case.

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
 
-#### Enum and const
+**Enum and const**
 
 `enum` and `const` are ONLY used to infer the type if it has not already been defined. Type is inferred from the provided value(s).
 
-#### Not
-
-`not` if defined always overwrites any existing inference of types.
+**Not**
```
smoya, line 41: I think you can use an h3 here (`###` prefix) so it will render as you expect.

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
 
-#### Non-mentioned keywords
+if defined removes any defined/inferred types defined in `not`.
```
smoya, line 43:
```suggestion
If defined, removes any defined/inferred types defined in `not`.
```

docs/simplification.md
```diff
@@ -12,89 +18,83 @@ The AsyncAPI Model SDK supports simplification of:
...
 
-Any keywords not mentioned as a title are all cumulating the type.
+<br/>
 
 ## Determining the enums for the model
 
-In order to determine all the possible enums a schema can be we both infer and use existing definitions of types, however we need to define a precedence for JSON Schema keywords and in which order they are applied.
+In order to determine the possible enums a model can be we both infer and use existing definitions, however we need to define a precedence for JSON Schema keywords and in which order they are applied.
```
smoya, line 49: Same as https://github.com/asyncapi/generator-model-sdk/pull/87/files#r580500164. I feel it sounds a bit redundant.

smoya:
@jonaslagoni Great job here! 
I did some reviewing. However and if it sounds ok to you, I would like to do another pass in case you apply some of the suggestions so I can read it closer from the user perspective again.

docs/input processing.md
jonaslagoni: Hmm, yea you are right. Let me see if jsdoc provides some functionality for us when we extend classes.

README.md
smoya: > Only solution to use package with parser written in another language is first parse schema, save parsed schema to file, and then read content from file and put in our `model-sdk`.

Considering that https://github.com/asyncapi/parser-go is gonna become a thing at some point (with a thing I mean something used consciously), perhaps make sense to move to the model you suggest @magicmatatjahu.
However, the work of outputting to a file would be a task to be done by the parsers (in their code) rather than by this generator, right?

magicmatatjahu: @smoya Yes exactly. Custom parser should save parsed schema inside file, and then someone should read content from file and put it in model-sdk as I described. Using internally another parser without file-workaround is also possible, but then we must use web-assembly, and it will be very complex 😄

smoya: I think I like the idea of files as input.

magicmatatjahu: But we must have in mind that if schema has circular references, then after saving it to file we will have missing data in some parts due to lack of circular refs.

jonaslagoni: I too like the idea of file inputs @smoya would you mind creating a feature request for this? 🤔

docs/simplification.md
jonaslagoni: The reason for the centering was that I felt like everything blended together otherwise 🤔

docs/simplification.md
jonaslagoni: @smoya maybe the following is better?
```suggestion
In order to determine all the possible types for a model, we both infer and use existing definitions of types. 
Precedence for JSON Schema keywords needs to be predetermined to ensure consistency.
```

docs/input processing.md
jonaslagoni: @fmvilas In the long run I would like for us to be able to provide it with any input from the parsed AsyncAPI document and it would generate all the models for that. Example operation, message, definition etc.

docs/simplification.md
smoya: @jonaslagoni Yup, looks really good to me 👍

docs/simplification.md
smoya: Fair enough. just felt a bit weird to me but it seems fine otherwise :) Thanks for your reasoning!

README.md
smoya: Done for this repository: https://github.com/asyncapi/generator-model-sdk/issues/97

Ideally a new issue should be created on each parser repository (JS and Go as per today). If you think it looks ok, I will proceed and create them.

docs/input processing.md
fmvilas: That makes sense. So then, from an AsyncAPI document, it generates all the models?

jonaslagoni: Yes, it goes through all defined messages and generates those - https://github.com/asyncapi/generator-model-sdk/blob/6480dce4649750e83118ff74172f7ce5aee0bc27/src/processors/AsyncAPIInputProcessor.ts#L23

docs/input processing.md
jonaslagoni: @smoya updated with link to generated docs.

API.md
```diff
@@ -110,10 +113,25 @@ This class is the wrapper for simplified models and the rest of the context need
...
+Common internal representation for a model.
 
 **Kind**: global class  
 **Extends**: [<code>CommonSchema&lt;CommonModel&gt;</code>](#CommonModel)  
+**Properties**
+
+| Name | Type | Description |
+| --- | --- | --- |
+| $id | <code>string</code> | define the id/name of the model. |
+| type | <code>string</code> \| <code>Array.&lt;string&gt;</code> | this is the different types for the model. All types from JSON Schema are used with no custom ones added. |
```
smoya, line 125: Is there any reason to use html tags instead of markdown here? E.g., you can achieve the same result in a more readable and simpler way:

```suggestion
| type | `string` \| `Array.<string>` | this is the different types for the model. All types from JSON Schema are used with no custom ones added. |
```

jonaslagoni: This is generated by the command `npm run docs` so I have no control over the API, I just generated it manually for you to see the changes it would do. Normally the CI system auto-generates this when PRs are merged

smoya: Oh didn't know about it 👍  thks

docs/input processing.md
```diff
@@ -0,0 +1,35 @@
+# Input processing
```
magicmatatjahu, line 1: Change name of file to `input_processing.md` :) Expect that everything is good.

docs/input processing.md
```diff
@@ -0,0 +1,35 @@
...
+The input process is about processing any input into our internal model representation `CommonInputModel`. This most likely have to be done using different methods based on which kind of input it is.
+
+As of now two inputs are supported:
+- JSON Schema Draft 7
+- AsyncAPI version 2.0.0
+
+## Internal model representation
+
+<img src="./assets/class_diagram.png"
+     alt="Class diagram" />
```
smoya, line 11: You can use Markdown syntax here.
```suggestion
![Class diagram](./assets/class_diagram.png)
```

smoya (dismissed):
Great job here! LGTM now 🚀 🌕

jonaslagoni:
Sorry @smoya, had a merge conflict, thanks for the review 🙇

smoya:
Ouch, seems like you have another one 😅  @jonaslagoni

jonaslagoni:
@smoya luckily not a conflict 😄

smoya (approved):
Fantastic work! 🚀

PR #88, opened 2021-02-22
docs: document the customisation flow

magicmatatjahu:
**Description**

Add docs for `presets` - customisation process in our `model-sdk`.

**Related issue(s)**
See https://github.com/asyncapi/shape-up-process/issues/55


docs/customisation.md
```diff
@@ -0,0 +1,184 @@
...
+    additionalContent(...options) { /* logic */ },
+  },
+}
+```
+
+Each language has different model types, which results in different implementable methods in a single preset. For more information, please check the [preset's shape](#presets-shape) section.
+
+## Custom preset
+
+Below is a custom preset written for TypeScript language, which adds a description to each interface's property and to the self as a JavaScript comment.
```
fmvilas, line 29:
```suggestion
Below is a custom preset written for TypeScript language, which adds a description to each interface's property and to the model itself as a JavaScript comment.
```

fmvilas (changes requested):

docs/customisation.md
```diff
@@ -0,0 +1,184 @@
...
+}
+
+const generator = new TypeScriptGenerator({ defaultPreset: DEFAULT_PRESET });
+```
+
+> **NOTE**: Each default preset must have implemented each method for a given model type. Keep this in mind when overriding the default preset.
+
+## Preset's shape
+
+By each preset user can implement two basic methods:
```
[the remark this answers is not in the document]
magicmatatjahu: Yes, this is true, because each model type, like class, enum, interface has by default these two method and then appropriate model - like class - has another methods like setter, getter etc.

I'll change this sentence to `For each model type, you can implement two basic methods:`

docs/customisation.md
```diff
@@ -0,0 +1,184 @@
...
+
+- `self` - the method for extending the whole model shape.
```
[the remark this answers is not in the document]
magicmatatjahu: I wanna leave the `extending` word, because as in example showed in document you can also extending model shape, not only replacing.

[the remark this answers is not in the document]
magicmatatjahu: So I removed the `whole` word, it causes misunderstanding.

docs/customisation.md
```diff
@@ -0,0 +1,184 @@
...
+
+- `self` - the method for extending the model shape.
+- `additionalContent` - the method which adds additional content to the model.
+
+Each customization method receives the following arguments:
+
+- `model` - an instance of the [`CommonModel`](../src/models/CommonModel.ts) class, which described rendered data model.
+- `inputModel` - an instance of the [`CommonInputModel`](../src/models/CommonInputModel.ts) class.
+- `renderer` - an instance of the class with common helper functions to render appropriate model type.
+- `content` - rendered content from previous preset.
```
jonaslagoni, line 126:
```suggestion 
- `content` - rendered content based on the language and preset.
```
Maybe?

docs/customisation.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### **Enum**
+
+| Method | Description | Additional arguments |
+|---|---|---|
+| `item` | A method to extend enum's item. | an `item` containing the value of enum's item and `parentModel` as a [`CommonModel`](../src/models/CommonModel.ts) instance. |
+
+#### **Type**
+
+There are no additional methods.
```
jonaslagoni, line 184: I am just gonna put the same comment @smoya left for my PR regarding quick outdated docs. Can we add jsdoc to make these explanations?

jonaslagoni (changes requested):

docs/customisation.md
magicmatatjahu: You cannot change language in rendering process and `content` is exactly rendered content from previous preset.

docs/customisation.md
magicmatatjahu: I also thought about that, but the problem is that types for presets are a TS interfaces. Also another options was pointing to source code of preset types, but there we have a problem, because types use generics, so it hard to read if you are not familiar with TS. In the future I have a plan for each generator to have a separate document with options/presets etc and then it will be easier to maintain.

docs/customisation.md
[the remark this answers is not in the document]
fmvilas: Yeah, much better. Thanks!

docs/customisation.md
magicmatatjahu: I tried to add jsdoc to interfaces and it doesn't work.

smoya: Totally unfamiliar with TS jsdoc but, are we sure it is not a supported feature? I saw some issues where seems it's a thing - E.g. https://github.com/microsoft/TypeScript/issues/18358

magicmatatjahu: Yes, TS has supports for jsdoc, but interfaces is missed after transpilation to JS and even if I changed script related to render the `API.md` to `./node_modules/.bin/jsdoc2md lib/index.js -f lib/**/*.ts > API.md` then I have empty file.

magicmatatjahu: We transpile TS to JS and from JS we render docs.

smoya: I guess that's why things like https://github.com/SoftwareBrothers/better-docs#how-it-works exist. Also, you can generate docs from TS from it afaik. Not sure what's the real effort of this, but I guess it's a topic non purely related with this PR.

magicmatatjahu: We use https://github.com/jsdoc2md/jsdoc-to-markdown so at the moment we don't want to switch :)

fmvilas: @smoya maybe you can create an issue so we don't forget about better-docs package.

fmvilas (approved):

fmvilas:
Great job with the documentation, Maciej! 🚀

docs/customisation.md
smoya: > We use https://github.com/jsdoc2md/jsdoc-to-markdown so at the moment we don't want to switch :)
Fair enough 👍 

> @smoya maybe you can create an issue so we don't forget about better-docs package.

Yup, makes totally sense. Thanks for pointing that out @fmvilas

magicmatatjahu: @smoya Maybe only in https://github.com/asyncapi/generator-react-sdk. In other place we use pure JS.

fmvilas: Let's just leave it there and if we prefer this one over the jsdoc-to-markdown maybe we can just replace it everywhere.

smoya (approved):
LGTM! 🎉

magicmatatjahu:
@fmvilas @smoya Thanks for review! :)

README.md
```diff
@@ -103,6 +104,10 @@ interface Address {
 }
 ```
 
+## Customisation
+
+The AsyncAPI Model SDK uses **presets** objects to extend the rendered model. For more information, check [customisation document](./docs/customisation.md).
```
jonaslagoni, line 109:
```suggestion 
The AsyncAPI Model SDK uses **preset** objects to extend the rendered model. For more information, check [customisation document](./docs/customisation.md).
```

docs/customisation.md
```diff
@@ -0,0 +1,184 @@
+# Customisation
+
+The AsyncAPI Model SDK uses **presets** objects to extend the rendered model.
```
jonaslagoni, line 3:
```suggestion 
The AsyncAPI Model SDK uses **preset** objects to extend the rendered model.
```

jonaslagoni (changes requested):
Just a small change 😄

jonaslagoni (approved):

fmvilas (approved):

PR #89, opened 2021-02-22
chore: removes eslint warnings from property simplification

jonaslagoni:
**Description**
This PR removes all eslint warnings.

```
/home/runner/work/generator-model-sdk/generator-model-sdk/src/simplification/SimplifyProperties.ts
Warning:   14:25  warning  Refactor this function to reduce its Cognitive Complexity from 35 to the 15 allowed  sonarjs/cognitive-complexity
```
**Related issue(s)**
related to #53

### DIFF
--- a/src/simplification/SimplifyProperties.ts
+++ b/src/simplification/SimplifyProperties.ts
@@ -16,56 +16,71 @@ export default function simplifyProperties(schema: Schema | boolean, simplifier 
     if (seenSchemas.has(schema)) return seenSchemas.get(schema);
     const output: Output = {};
     seenSchemas.set(schema, output);
-    const addToProperty = (propName: string, propModel: CommonModel) => {
-      if (output !== undefined) {
-        //If a simplified property already exist, merge the two
-        if (output[`${propName}`] !== undefined) {
-          output[`${propName}`] = CommonModel.mergeCommonModels(output[`${propName}`], propModel, schema);
-        } else {
-          output[`${propName}`] = propModel;
-        }
-      }
-    };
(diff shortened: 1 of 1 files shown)


src/simplification/SimplifyProperties.ts
```diff
@@ -16,56 +16,71 @@ export default function simplifyProperties(schema: Schema | boolean, simplifier
...
+ * 
+ * @param schema to go through
+ * @param currentModel the current output
+ * @param simplifier the simplifier to use
+ * @param seenSchemas schemas which we already have outputs for
+ * @param rootSchema the root schema we are combining schemas for
+ */
+function combineSchemas(schema: (Schema | boolean) | (Schema | boolean)[] | undefined, currentModel: Output, simplifier : Simplifier, seenSchemas: Map<any, Output>, rootSchema : Schema) {
+  if (schema === undefined) return;
+  if (typeof schema === 'boolean') return;
```
magicmatatjahu, line 73:
```suggestion
  if (typeof schema !== 'object') return;
```

magicmatatjahu (changes requested):

src/simplification/SimplifyProperties.ts
jonaslagoni: We have to check for undefined because it can be that type. It is easier in the main functionality to allow such value in this sub-function.

magicmatatjahu: As I wrote in another PR. For undefined you have also `false`.

magicmatatjahu (approved):

PR #90, opened 2021-02-23
chore: export class/functions from package

magicmatatjahu:
**Description**

As in title.


magicmatatjahu:
Will be merged here https://github.com/asyncapi/generator-model-sdk/pull/85

PR #95, opened 2021-02-23
fix: render TS class properties in proper way

magicmatatjahu:
**Description**

Render in proper way properties in TS class:
- add `_` prefix for private fields in class.
- render appropriate object shape for constructor argument.


jonaslagoni (approved):
👍

PR #103, opened 2021-02-24
fix: update JavaRenderer and remove unnecessary code from presets

magicmatatjahu:
**Description**

- Fix issue https://github.com/asyncapi/generator-model-sdk/issues/98
- PascalCase model's name.
- Remove `parentModel` argument - we have `model` and it's the same
- Add missed two function in JavaRenderer: `renderComments` and fix `renderAnnotation`.
- Render type by `format` property if exists in JavaRenderer.
- Update unit tests 

**Related issue(s)**
Resolves https://github.com/asyncapi/generator-model-sdk/issues/98


jonaslagoni (approved):
LGTM 👍

PR #105, opened 2021-02-24
feat: add support function to get all CommonModel dependencies

jonaslagoni:
**Description**
This PR adds a helper function to CommonModel which returns all references to other CommonModels. This is useful when you want to add imports and split the models into separate files.


magicmatatjahu (approved):
LGTM

PR #109, opened 2021-02-24
fix: switch to merging allOf keywords instead of extending them

### LINKED ISSUE #107: Renderer does not render extended classes which is simplified by default
By default, we simplify JSON schema into multiple classes such that:
```json
{
  "allOf": [
    {
      "type": "object",
      "properties": {
        "street_address": {
          "type": "string"
        },
        "country": {
          "type": "string"
        }
      }
    },
    {
      "type": "object",
      "properties": {
        "street_address2": {
          "type": "string"
        },
        "country2": {
          "type": "string"
        }
      }
    }
  ]
}
```
should render 3 classes, `root` which extends `rootAllOf1` and `rootAllOf2`, and then the later two classes are then separated. But since the renderer is not ready for this we need to switch to defaulting to merging the `allOf` schemas instead. So no schemas are left behind.

jonaslagoni:
**Description**
This PR switches the default option from extending to merging.

**Related issue(s)**
fixes #107

### DIFF
--- a/src/simplification/Simplifier.ts
+++ b/src/simplification/Simplifier.ts
@@ -12,7 +12,7 @@ import simplifyName from './SimplifyName';
 
 export class Simplifier {
   static defaultOptions: SimplificationOptions = {
-    allowInheritance: true
+    allowInheritance: false
   }
 
   private anonymCounter = 1;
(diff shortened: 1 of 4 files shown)


magicmatatjahu (approved):
LGTM

PR #112, opened 2021-02-25
feat: add common presets for Java

magicmatatjahu:
**Description**

Integrating our `model-sdk` into `java-spring-template` in this task https://github.com/asyncapi/shape-up-process/issues/49, I had to create some presets that add description to models and annotations from `javax.validation.constraints` and `com.fasterxml.jackson` packages. They are generic enough to add them as PR.

- Add common presets for Java:
  - `JAVA_DESCRIPTION_PRESET` - adds description for classes (self and for getters) and for enums.
  - `JAVA_CONSTRAINTS_PRESET` - adds to getters the annotations related to `javax.validation.constraints` package.
  - `JAVA_JACKSON_PRESET` - add to getters the annotations related to `com.fasterxml.jackson` package.
  - `JAVA_COMMON_PRESET` - add inside class `hashCode`, `toString`, `equal` functions.
- Add unit tests for presets

**Related issue(s)**
https://github.com/asyncapi/shape-up-process/issues/49


magicmatatjahu:
@jonaslagoni Could you look on this PR? Also I think that PR can be useful for followup issues https://github.com/asyncapi/generator-model-sdk/issues/128 I added preset to render description and examples from JSON as well, then we can reuse it in another generators. Also this PR block this https://github.com/asyncapi/java-spring-template/pull/121 PR - I don't want to duplicate generic presets in template.

src/generators/java/presets/ConstraintsPreset.ts
```diff
@@ -0,0 +1,57 @@
...
+      const exclusiveMaximum = property.getFromSchema('exclusiveMaximum');
+      if (exclusiveMaximum !== undefined) {
+        annotations.push(renderer.renderAnnotation('Max', exclusiveMaximum - 1));
+      }
+    
+      const minItems = property.getFromSchema('minItems');
+      const maxItems = property.getFromSchema('maxItems');
+      if (minItems !== undefined || maxItems !== undefined) {
+        annotations.push(renderer.renderAnnotation('Size', { min: minItems, max: maxItems }));
+      }
```
jonaslagoni, line 52: I feel like I have to leave a comment here 😄 In complicated JSON Schemas (when keywords for combination or conditional are used), this will not give the desired output. 

I know you added it because the java-spring-template had them. But they should be simplified, and even that process is difficult and needs to be well thought out how you merge such formatting rules.

IMO rendering these formatting rules when they have not been simplified will in the worse cases not give a useable output for the user. In the common cases yes it will work, but if there is just one combination or conditional keyword described contradicting formatting rules this will not provide a useable outcome.
i.e. 
```
{
  "type": "string",
  "maxLength": 50,
  "allOf": [
    { "maxLength": 2 }
    { "maxLength": 5 }
  ]
}
```
Should only work with strings between 0 and 5 characters but with this implementation up to 50 is allowed.

jonaslagoni (commented):
In general, I have one comment which is about the use of `getFromSchema` in the standard presets. 

IMO we should limit the amount of usage if at all use it since it does not provide desirable outcomes in complex schemas.

I don't mind approving it now since the majority of use-case will work... But I will have to ask you to create issues for the bugs that will be introduced. That way we have a trace of what needs to be changed to fix them and keep track of them 🙂 


Other than that it looks great 👍

src/generators/java/presets/ConstraintsPreset.ts
magicmatatjahu: Yes, I know about this problem, but as you wrote in main comment, I will create followup issue for tracking such a cases. For "simple" metadata fields, not related to any type, like `description` etc `getFromSchema` function should work without any problems, but with combined schema and types like `number`, `string` etc we should simplify it. At the moment any template doesn't support nested metadata, so right now it's not a big problem.

magicmatatjahu: Here is a issue https://github.com/asyncapi/generator-model-sdk/issues/130

jonaslagoni (approved):
👍

PR #114, opened 2021-02-25
fix: remove customizations property from CommonInputModel class

magicmatatjahu:
**Description**

As in title: remove `customizations` property from `CommonInputModel` class.


jonaslagoni (approved):
👍

PR #116, opened 2021-02-26
docs: add documentation for generators

magicmatatjahu:
**Description**

Add documentation for generators:
- document how to pass options to generator.
- document options for generators. 
- document how to write custom generator.

**Related issue(s)**
Resolves https://github.com/asyncapi/generator-model-sdk/issues/46
See also https://github.com/asyncapi/generator-model-sdk/issues/45


docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+| `renderTypes` | Boolean | Renders or not signature for types. | `true` |
+| `modelType` | String | It indicates which model type should be rendered for `object` type. Its value can be either `interface` or `class`. | `class` |
+
+## Custom generator
+
+Creating a new generator which will become a part of the officially supported languages in AsyncAPI Model SDK is pretty simple. The minimum set of required actions to create a new generator are:
+
+- source code must be included in [generators](../src/generators) folder.
+- must extend the abstract [`AbstractGenerator`](../src/generators/AbstractGenerator.ts) class,
+- must define [`Preset`](./customisation.md)'s shape for language,
```
jonaslagoni, line 46:
```suggestion
- must define [`Preset`](./customisation.md)'s shape for the language,
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+Below is a list with additional options available for a given generator.
+
+### [TypeScript](../src/generators/typescript/TypeScriptGenerator.ts)
+
+| Option | Type | Description | Default value |
+|---|---|---|---|---|
+| `renderTypes` | Boolean | Renders or not signature for types. | `true` |
+| `modelType` | String | It indicates which model type should be rendered for `object` type. Its value can be either `interface` or `class`. | `class` |
+
```
jonaslagoni, line 39: The table is not showing up properly when rendered - https://github.com/asyncapi/generator-model-sdk/blob/a5817b374172a673f5dd567d71823d741eb997f4/docs/generators.md

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+- source code must be included in [generators](../src/generators) folder.
+- must extend the abstract [`AbstractGenerator`](../src/generators/AbstractGenerator.ts) class,
+- must define [`Preset`](./customisation.md)'s shape for language,
+- must define language options by passing an interface describing additional options to the first generic argument of [`AbstractGenerator`](../src/generators/AbstractGenerator.ts). Interface must also be extended by `CommonGeneratorOptions` interface,
+- must define default options as static class's field, which must be extended by `defaultGeneratorOptions`,
+- default options must include `defaultPreset` property,
+- must implement `render` function,
+- must define **renderer**s classes for available model types in given language. **Renderer** is an instance of the class with common helper functions to render appropriate model type and must be extended by [`AbstractRenderer`](../src/generators/AbstractRenderer.ts) class - [example](../src/generators/typescript/renderers/ClassRenderer.ts).
```
jonaslagoni, line 51: Instead of listing the requirements, what about a here is how you do it 1,2,3? 🤔

jonaslagoni (changes requested):

docs/generators.md
magicmatatjahu: No matter what to do first, as some items can be done sooner or later, so for me the listing is actually better that list :)

magicmatatjahu:
@jonaslagoni Thanks very much for review! Suggestions are applied :)

docs/generators.md
jonaslagoni: It might just be me that don't fancy this way of describing how to implement something 😄 Hope you dont mind, just gonna cc some others to get some more opinions 🙂 @derberg @smoya

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+- [JavaScript](../src/generators/javascript/JavaScriptGenerator.ts),
+- [TypeScript](../src/generators/typescript/TypeScriptGenerator.ts),
+- [Java](../src/generators/java/JavaGenerator.ts).
+
+## Generator's options
+
+Each generator extends default options for the generator. It means that the generator can also have additional options.
+
+Options is passed as first argument to generator's constructor. Check the example:
```
derberg, line 15:
```suggestion
Options are passed as the first argument to the generator's constructor. Check the example:
```

README.md
```diff
@@ -41,27 +41,29 @@ Run this command to install the SDK in your project:
...
 ## How it works
 
 The process of creating data models from input data consists of 2 processes, the input and generation process.
 
 ### The input process
 
 The input process ensures that any supported input is handled correctly, the basics are that any input needs to be converted into our internal model representation `CommonModel`. The following inputs are supported:
-- [JSON Schema Draft 7](#JSON-Schema-input), this is the default inferred input if we cannot find a another input processor.
+
+- [JSON Schema Draft 7](#JSON-Schema-input), this is the default inferred input if we cannot find an another input processor.
```
smoya, line 53: Technically the `a` or `an` article is not needed. You are free to remove it.

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+Default options contains:
+
+| Option | Type | Description | Default value |
+|---|---|---|---|
+| `indentation` | Object | Options for indentation. | - |
+| `indentation.type` | String | Type of indentation. Its value can be either `SPACES` or `TABS`. | `SPACES` |
+| `indentation.size` | String | Size of indentation. | 2 |
+| `defaultPreset` | Object | Default preset for generator. For more information, read [customisation](./customisation.md) document. | _Implemented by generator_ |
```
derberg, line 28:
```suggestion
| `defaultPreset` | Object | Default preset for generator. For more information, read [customization](./customisation.md) document. | _Implemented by generator_ |
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+| `presets` | Array | Array contains **presets**. For more information, read [customisation](./customisation.md) document. | `[]` |
```
derberg, line 29:
```suggestion
| `presets` | Array | Array contains **presets**. For more information, read [customization](./customisation.md) document. | `[]` |
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+Below is a list with additional options available for a given generator.
```
derberg, line 31:
```suggestion
Below is a list of additional options available for a given generator.
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+### [TypeScript](../src/generators/typescript/TypeScriptGenerator.ts)
+
+| Option | Type | Description | Default value |
+|---|---|---|---|
+| `renderTypes` | Boolean | Renders or not signature for types. | `true` |
+| `modelType` | String | It indicates which model type should be rendered for `object` type. Its value can be either `interface` or `class`. | `class` |
```
derberg, line 38:
```suggestion
| `modelType` | String | It indicates which model type should be rendered for the `object` type. Its value can be either `interface` or `class`. | `class` |
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Custom generator
+
+Creating a new generator which will become a part of the officially supported languages in AsyncAPI Model SDK is pretty simple. The minimum set of required actions to create a new generator are:
```
derberg, line 42:
```suggestion
Creating a new generator that will become a part of the officially supported languages in AsyncAPI Model SDK is pretty simple. The minimum set of required actions to create a new generator are:
```

`is pretty simple` do we need this 😄 pretty subjective 😄

docs/generators.md
smoya, line 15:
```suggestion
Options are passed as the first argument to the generator constructor. Check the example:
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+Each generator extends default options for the generator. It means that the generator can also have additional options.
+
+Options is passed as first argument to generator's constructor. Check the example:
+
+```ts
+const generator = new TypeScriptGenerator({ ...options });
+```
+
+Default options contains:
```
smoya, line 21:
```suggestion
Default options contain:
```

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+- source code must be included in [generators](../src/generators) folder.
+- must extend the abstract [`AbstractGenerator`](../src/generators/AbstractGenerator.ts) class,
+- must define [`Preset`](./customisation.md)'s shape for the language,
+- must define language options by passing an interface describing additional options to the first generic argument of [`AbstractGenerator`](../src/generators/AbstractGenerator.ts). The interface must also be extended by `CommonGeneratorOptions` interface,
+- must define default options as static class's field, which must be extended by `defaultGeneratorOptions`,
+- default options must include `defaultPreset` property,
+- must implement `render` function,
+- must define **renderer**s classes for available model types in a given language. **Renderer** is an instance of the class with common helper functions to render appropriate model type and must be extended by [`AbstractRenderer`](../src/generators/AbstractRenderer.ts) class - [example](../src/generators/typescript/renderers/ClassRenderer.ts).
```
derberg, line 51: something broken here with bold

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+
+Check the [generator implementation](../src/generators/typescript/TypeScriptGenerator.ts) for `TypeScript` language to see how it should looks like.
```
derberg, line 53:
```suggestion
Check the [generator implementation](../src/generators/typescript/TypeScriptGenerator.ts) for `TypeScript` language to see how it should look like.
```

derberg (commented):
@jonaslagoni bullet points are fine if there is not order required, 
more important is to have a ref to an example where possible.

docs/generators.md
```diff
@@ -0,0 +1,53 @@
...
+| `defaultPreset` | Object | Default preset for generator. For more information, read [customisation](./customisation.md) document. | _Implemented by generator_ |
+| `presets` | Array | Array contains **presets**. For more information, read [customisation](./customisation.md) document. | `[]` |
+
+Below is a list with additional options available for a given generator.
+
+### [TypeScript](../src/generators/typescript/TypeScriptGenerator.ts)
+
+| Option | Type | Description | Default value |
+|---|---|---|---|
+| `renderTypes` | Boolean | Renders or not signature for types. | `true` |
```
smoya, line 37: `Renders or not signature for types` is hard to understand IMHO. Since it is a boolean type, its clear that it enables/disables something. What about something like: 

```suggestion
| `renderTypes` | Boolean | Render signature for types. | `true` |
```

docs/generators.md
smoya, line 51: To be consistent with the style used across all docs, I think we should capitalize the first character of each bullet.

smoya (commented):
So good documentation work here 💯 👏   
I just wrote some minors I consider nice to fix. I also encourage you to pass the doc through Grammarly, which helps to catch some typos (at least it does help me a lot)

README.md
magicmatatjahu: I use grammarly and sometimes it highlights such as these errors.

docs/generators.md
magicmatatjahu: I have problem here :) `CustomiZation` is sometimes highlighted as error, so I changed to version with `S`, but I see that between these two versions aren't any differences https://grammarist.com/spelling/customise-customize/

docs/generators.md
magicmatatjahu: Like above.

docs/generators.md
magicmatatjahu: In renderer view everything is fine https://github.com/asyncapi/generator-model-sdk/blob/9c05e429e32e7ceca8af2e30f6d48137d18547d2/docs/generators.md#custom-generator So I leave it as it is.

docs/generators.md
magicmatatjahu: I wanna to leave this version with `S` because in another docs we also use this version. Is it ok for you?

docs/generators.md
magicmatatjahu: Yeah, it's better :)

docs/generators.md
smoya: the file is called `customisation.md` with `s`. Meaning we use the British form. Do we have any agreement on what to use? (US VS British English forms)

magicmatatjahu: No, we don't (as I know) - I used `s` because grammarly highlighted me the errors for `z`.

magicmatatjahu:
@derberg @smoya Thanks for review mates!

@derberg I added reference to `TypeScript` generator example - making reference for every point is not needed - everything is in one file, like here - https://github.com/asyncapi/generator-model-sdk/blob/master/src/generators/typescript/TypeScriptGenerator.ts

docs/generators.md
magicmatatjahu: I left only second sentence here and added at the end of file this

> If you created a generator then you can contribute it to the AsyncAPI Model SDK and it will become the official supported generator.

It is ok for you?

docs/generators.md
smoya: I kinda agree @jonaslagoni. Getting an idea of the requirements is hard by just reading those points. The reality is that some requirements look way more complicated than in reality. For example:

> Must define language options by passing an interface describing additional options to the first generic argument of [`AbstractGenerator`](../src/generators/AbstractGenerator.ts). The interface must also be extended by `CommonGeneratorOptions` interface

I think this should be part of the library api documentation by its own. Perhaps just showing an example of the usage here would be simpler for the user?

magicmatatjahu: @smoya Example is mentioned below the list 

> Check the [generator implementation](../src/generators/typescript/TypeScriptGenerator.ts) for `TypeScript` language to see how it should look like.

Link: https://github.com/asyncapi/generator-model-sdk/blob/master/src/generators/typescript/TypeScriptGenerator.ts

Making inline example like:

```ts
export interface TypeScriptOptions extends CommonGeneratorOptions<TypeScriptPreset> {
  ...
}

export class TypeScriptGenerator extends AbstractGenerator<TypeScriptOptions> {...}
```

is unnecessary in my POV - you have linked example for that. I don't like to write documentation like for idiots -> make this in this way, this in another way etc 😄 I mean, creating guidelines for every case etc. Also I agree with you that `some requirements look way more complicated than in reality`, if you extend custom generator by `AbstractGenerator` your IDE automatically throws errors that you must define `render` function or pass interface as generic argument - and this interface must be extended by CommonGeneratorOptions etc. Some people which will want to contribute, probably won't understand TS, but this is a problem not related to guidelines.

jonaslagoni (dismissed):

docs/generators.md
derberg: your grammarly is just set for British English, I recommend switching to US English as it is just a standard in technical writing.

for me, this is not a blocker for this PR because we are in general not consistent in this matter. I would just, for now, recommend always follow US English as this is standard worldwide

derberg (dismissed):

docs/generators.md
magicmatatjahu: Ok, I changed to `customiZation` everywhere :) Thanks!

jonaslagoni (dismissed):

docs/generators.md
magicmatatjahu: Done :)

derberg (approved):
:shipit:

PR #118, opened 2021-02-27
fix: `integer` type should also be implied

jonaslagoni:
**Description**
This PR introduces the following:
- Improving test coverage
- fixed `integer` type was not implied.


src/simplification/SimplifyTypes.ts
```diff
@@ -81,40 +81,37 @@ function inferTypeFromValue(value: any) {
...
  * @param currentOutput the current output
  * @param seenSchemas schemas which we already have outputs for
  */
 function inferNotTypes(schema: Schema | boolean, currentOutput: Output, seenSchemas: Map<any, Output>) {
-  if (schema === undefined || currentOutput === undefined) return;
-  if (!Array.isArray(currentOutput)) return;
-  if (currentOutput.length === 0) currentOutput.push('object', 'string', 'number', 'array', 'boolean', 'null');
-  if (typeof schema === 'boolean') return;
-  if (schema.not) {
+  if (typeof schema === 'object' && schema.not && Array.isArray(currentOutput)) {
```
magicmatatjahu, line 113: I have only problem here. ^think that `schema.not` should be checked after rewrite `currentOutput` with these 7 types, like in previous code.

jonaslagoni: Hmm, yea I can see it changes the behavior 🤔 But this is actually the correct one - "if not keyword is defined and no types have been implied, all types must be included".

magicmatatjahu (approved):
LGTM

PR #119, opened 2021-02-27
chore: add unit tests to renderers

jonaslagoni:
**Description**
This PR introduces the following changes:
- It removes unused switch case in TypeScript render, CommonModel type cannot be `bigint`
- Added unit tests for JavaRendere, JavaScriptRendere and TypeScriptRenderer

PR #117, opened 2021-02-27
fix: some keywords are not iterated for AsyncAPI input

jonaslagoni:
**Description**
This PR introduces two things.
- Added unit test for reflecting schema name based on our parser.
- Fixed a bug where properties, dependencies, and definitions were never iterated.

PR #120, opened 2021-02-27
chore: add test coverage badges

jonaslagoni:
**Description**
This PR is more a test to see if this is something we would like to visually display.

**Related issues**
blocked by the following issues.
- #117
- #118 
- #119 
- #121


derberg:
did you consider using maybe something like https://coveralls.io/ ? I bet they generate badges that we do not have to store in the repo

jonaslagoni:
@derberg that would be optimal, do we have an account?

derberg:
nope, but I can get one 😄 if we want to use this and not something else. I just know this one as I saw it while contributing to redocly project

jonaslagoni:
Pretty easy to integrate I see https://github.com/nickmerwin/node-coveralls/issues/126#issuecomment-242963700

derberg:
ok, I logged it and enabled asyncapi org, turned it on for this repo only for now

derberg:
there is official action for it https://github.com/marketplace/actions/coveralls-github-action

jonaslagoni:
> ok, I logged it and enabled asyncapi org, turned it on for this repo only for now

Thaaank you 😘 

> there is official action for it https://github.com/marketplace/actions/coveralls-github-action 

Even better 😄

jonaslagoni:
@derberg finally got some time to solve this. Even though that the check fails, I assume it is because the workflow is not yet added (merged) to the default branch (master)... 🤔 I cannot find any information about it though... I have the basic setup without any modification (besides the node version is changed).

.github/workflows/coverall.yml
```diff
@@ -0,0 +1,20 @@
+on: ["push", "pull_request"]
```
derberg, line 1: I didn't use this action so not sure what it is doing, but if it is suppose to put some updates on a PR, then it needs to be `pull_request_target` to make it work for forks

and about `push`, do we really need it? in theory PR should be fine?

.github/workflows/coverall.yml
```diff
@@ -0,0 +1,20 @@
...
+name: Test Coveralls
```
derberg, line 2: maybe better name so it is clear in check section what is done, maybe `Check test coverage`?

derberg (commented):
I left some comments.
I do not know why it doesn't work, but for sure the reason is not that it is not in the master

the coveralls action has poor documentation. Try adding [this](https://github.com/coverallsapp/coveralls-node-demo/blob/master/.github/workflows/workflow.yml#L22-L23) debug flag to see if maybe more logs will explain

.github/workflows/coverall.yml
jonaslagoni: Well... I think I broke it somehow, cause I cannot get it to run the workflow anymore 😩

jonaslagoni: Lol, I have to solve conflicts from master before workflow changes take effect 🤯

jonaslagoni: @derberg I am not sure if we should change to `pull_request_target`. Currently, I am using this trigger, but PR's do not trigger the workflow on this PR 🤔 I am not sure if that means that the coverage bot wont add comments on the PRs as wanted...

jonaslagoni: It wont as stated in the readme https://github.com/coverallsapp/github-action#coveralls-github-action
> When running on pull_request events, a comment will be added to the PR with details about how coverage will be affected if merged.

jonaslagoni: > and about push, do we really need it? in theory PR should be fine?

We need push on the master branch for when PRs are merged. Otherwise, the repo has no stats to compare PR's with.

derberg: > @derberg I am not sure if we should change to pull_request_target. Currently, I am using this trigger, but PR's do not trigger the workflow on this PR 🤔 I am not sure if that means that the coverage bot won't add comments on the PRs as wanted...

> It wont as stated in the readme https://github.com/coverallsapp/github-action#coveralls-github-action

we have to, we cannot trust their readme to much 😄 
the thing is, they add a comment, to do that they need a token, then need write-token, and on `pull_request` it is not available on a fork, fork has access to read-only token.

of course it means that you need to first merge the workflow to see if it works 🤷🏼 
This action is strange, you do not need to provide any Coveralls token to talk to it, just GH token. For me it means they are probably analyzing the scopes of the token, and if you do not have `write` you cannot call their API. I hope I explained my view good? Makes sense?

> We need push on the master branch for when PRs are merged. Otherwise, the repo has no stats to compare PR's with.

but you do not push on the master, not in this PR. Even `push` means that workflow will trigger after merge, and that is it, it will just run, not commit/push or anything like that, so nothing will be saved on `master`. Why did you remove https://github.com/asyncapi/generator-model-sdk/pull/120/commits/700f3773fad8a4c90284478cc655681dd29eee86#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L61 ? was it not needed?

jonaslagoni: > For me it means they are probably analyzing the scopes of the token, and if you do not have write you cannot call their API. I hope I explained my view good? Makes sense?

Remember that they have a bot that comments, and with public repositories, it should not be a problem. We just need to ensure that coveralls get notified of coverage changes in PRs. And that can only be done using `pull_request`.

> but you do not push on the master, not in this PR. Even push means that workflow will trigger after merge, and that is it, it will just run, not commit/push or anything like that, so nothing will be saved on master

We need the coverall job to run when PRs are merged for coverall to add the updated coverage report.

> Why did you remove 700f377#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L61 ? was it not needed?

Because we should never manually use the library, we use it though github action.

derberg:
looks like something happened https://coveralls.io/github/asyncapi/generator-model-sdk
although the commit message is strange

.github/workflows/coverall.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Setup Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    - name: npm install, run test
+      run: |
+        npm install
+        npm run test
+    - name: Coveralls
+      uses: coverallsapp/github-action@master
```
derberg, line 23: don't fix to master
```suggestion
      uses: coverallsapp/github-action@v1.1.2
```

derberg (commented):
one small change needed, then let us merge and see how it goes

derberg (approved):

derberg:
btw, have a look at sonar, it indicates some code duplication

jonaslagoni:
> btw, have a look at sonar, it indicates some code duplication

👍

jonaslagoni:
!@all-contributors please add @derberg for review

PR #151, opened 2021-04-26
fix: false schemas no longer crash the library

### LINKED ISSUE #143: Support false validation rules
#### Reason/Context
Currently, whenever types are simplified and come across a schema of type `false` it throws an error and the library won't generate anything which is far from ideal.

https://github.com/asyncapi/generator-model-sdk/blob/cc42b3ae7753044c173cbdd16cfa4f9fb7cb2c34/src/simplification/SimplifyTypes.ts#L16

#### Description
I suggest we just ignore that specific model entirely.

jonaslagoni:
**Description**
This PR ensure false schemas are handled more gracefully.

**Related issues**
Fixes #143

### DIFF
--- a/src/simplification/Simplifier.ts
+++ b/src/simplification/Simplifier.ts
@@ -39,7 +39,7 @@ export class Simplifier {
       }
     }
     //If it is a false validation schema return no common model
-    if (typeof schema === 'boolean' && !schema) {
+    if (schema === false) {
       return [];
     }
     const model = new CommonModel();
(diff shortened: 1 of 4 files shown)


src/simplification/Simplifier.ts
```diff
@@ -38,6 +38,10 @@ export class Simplifier {
         return [cachedModel];
       }
     }
+    //If it is a false validation schema return no common model
+    if (typeof schema === 'boolean' && !schema) {
```
magicmatatjahu, line 42:
```suggestion
    if (schema === false) {
```

magicmatatjahu (changes requested):

magicmatatjahu (approved):

PR #153, opened 2021-04-26
feat: simplify patternProperties

jonaslagoni:
**Description**
This PR moves the patternProperties property from Schema to CommonModel to allow rendering of pattern properties.

Beside this main feature:
- Fixed unnecessary ignore of lint rule.
- Added better docs
- Added missing tests for additionalProperties merge 

**Related issue(s)**
Solves #43

### DIFF
--- a/src/models/CommonSchema.ts
+++ b/src/models/CommonSchema.ts
@@ -7,7 +7,8 @@ export class CommonSchema<T> {
     enum?: any[];
     items?: T | T[];
     properties?: { [key: string]: T; };
-    additionalProperties?: boolean | T;
+    additionalProperties?: T;
+    patternProperties?: { [key: string]: T; };
     $ref?: string;
     required?: string[];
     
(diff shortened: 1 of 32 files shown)


src/models/CommonSchema.ts
```diff
@@ -7,7 +7,8 @@ export class CommonSchema<T> {
     enum?: any[];
     items?: T | T[];
     properties?: { [key: string]: T; };
-    additionalProperties?: boolean | T;
+    additionalProperties?: T;
```
magicmatatjahu, line 10: `additionalProperties` can be also false/true. Why you changed it?

jonaslagoni: Cause it was a mistake. For the `CommonModel` it should not be possible to be a boolean, it should be a `CommonModel`. For `Schema` it can be `Schema | Boolean` which it still can because of the generic type.

src/simplification/SimplifyPatternProperties.ts
```diff
@@ -0,0 +1,80 @@
...
+      combineSchemas(schema.allOf, output, simplifier, seenSchemas, schema);
+    }
+    combineSchemas(schema.oneOf, output, simplifier, seenSchemas, schema);
+    combineSchemas(schema.anyOf, output, simplifier, seenSchemas, schema);
+
+    //If we encounter combination schemas ensure we recursively find the properties
+    combineSchemas(schema.then, output, simplifier, seenSchemas, schema);
+    combineSchemas(schema.else, output, simplifier, seenSchemas, schema);
+
+    return !Object.keys(output).length ? undefined : output;
```
magicmatatjahu, line 36:
```suggestion
    return Object.keys(output).length ? output : undefined;
```

magicmatatjahu (approved):
LGTM! Only one change :)

src/simplification/SimplifyPatternProperties.ts
jonaslagoni: Gonna change it for https://github.com/asyncapi/generator-model-sdk/blob/2f1c8f57d843d7f97452276acfba26e879f3def7/src/simplification/SimplifyProperties.ts#L39 as well then.

magicmatatjahu (approved):

PR #155, opened 2021-04-26
feat: ensure unions of enums are rendered for properties for TS

jonaslagoni:
**Description**
This PR ensures the type for properties are rendered with enums if sat.

**Related issue(s)**
solves https://github.com/asyncapi/generator-model-sdk/issues/133

### DIFF
--- a/src/generators/typescript/TypeScriptRenderer.ts
+++ b/src/generators/typescript/TypeScriptRenderer.ts
@@ -23,6 +23,9 @@ export abstract class TypeScriptRenderer extends AbstractRenderer<TypeScriptOpti
     if (Array.isArray(model)) {
       return model.map(t => this.renderType(t)).join(' | ');
     }
+    if (model.enum !== undefined) {
+      return model.enum.map(value => typeof value === 'string' ? `"${value}"` : value).join(' | ');
+    }
     if (model.$ref !== undefined) {
       return model.$ref;
     }
--- a/test/generators/typescript/TypeScriptGenerator.spec.ts
+++ b/test/generators/typescript/TypeScriptGenerator.spec.ts
@@ -14,16 +14,16 @@ describe('TypeScriptGenerator', function() {
       }
     };
     const expected = `export class Address {
-private _state?: "Texas" | "Alabama" | "California" | "other";
+  private _state?: "Texas" | "Alabama" | "California" | "other";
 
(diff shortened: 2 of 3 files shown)


czlowiek488:
It may be enough and I will use it but I don`t like it.
In perfect world I as a user should be able to generate mutual enum if I reference to it by $ref in schema.
Also I would love to see generic interface when I use allOf with $ref.

czlowiek488:
I belive that if you can create mutual enum, it would be nice to add way specify somehow const using values from referenced enum.

jonaslagoni:
Hmm...

@czlowiek488 just to make it perfectly clear what you are looking for, please correct me if I am wrong :smile: 

Given the input of the following JSON Schema:
```
{
  $id: "Address",
  "type": "object",
  "properties": {
    "prop1": { "$ref": "#/definitions/enum1" },
    "prop2": { "$ref": "#/definitions/enum2" },
  },
  "definitions": {
    "enum1": { "enum": ["VAL1", "VAL2"] },
    "enum2": { "enum": ["VAL3", "VAL4"] },
  }
}
```
You expect the output of the class to be:

```
export class Address {
  private _prop1?: Enum1;
  private _prop2?: Enum2;
  ...
}
```
Instead of what it currently will generate:

```
export class Address {
  private _prop1?: "VAL1" | "VAL2";
  private _prop2?: "VAL3" | "VAL3";
  ...
}
```

czlowiek488:
@jonaslagoni Yes, exactly.

czlowiek488:
Also it would be awesome if I could make use of enum to specify a value in interface.
```
{
  $id: "Address",
  "type": "object",
  "properties": {
    "prop1": { 
        "const": "$ref": "#/definitions/enum1/VAL1"
    },
  },
  "definitions": {
    "enum1": { "enum": ["VAL1", "VAL2"] },
  }
}
```

```
export interface Address {
 prop1: Enum1.VAL1;
}
```

czlowiek488:
@jonaslagoni After a bit of thinking, it may be good to have a way to specify enums keys and values to achieve something like here.
```
export enum Data {
dataA = "data A value",
...
}
```

magicmatatjahu (approved):

jonaslagoni:
@czlowiek488 would you mind creating an issue with your use case?

PR #170, opened 2021-04-29
feat: add a logging interface

### LINKED ISSUE #152: Add logging interface
#### Reason/Context
We should provide a logging interface for users of the library to get notified on internal logs.

Use-case:
- In case two simplified models need to be merged some properties such as `$id` are overwritten, in these special cases I would like to inform the user that a clash between properties occurred and the output might not contain the expected output. 
- As a user I would like to see which models are generated

#### Description
I suggest we provide a logging interface so the users of the library can decide which logs they would like to do something with.

This could for example include 4 levels:
- Debug
- Info
- Warning
- Error

jonaslagoni:
**Description**
This PR adds a logging interface to the library to allow logging for different scenarios.

**Related issue(s)**
fixes #152

### DIFF
--- a/src/utils/LoggingInterface.ts
+++ b/src/utils/LoggingInterface.ts
@@ -14,11 +14,8 @@ export interface ModelLoggingInterface {
  * 
  * This class acts as a forefront for any external loggers which is why it also implements the interface itself.
  */
-export class Logger implements ModelLoggingInterface {
-  private static instance?: Logger;
-  private logger?: ModelLoggingInterface;
-
-  private constructor() { this.logger = undefined; }
+export class LoggerClass implements ModelLoggingInterface {
+  private logger?: ModelLoggingInterface = undefined;
 
   debug(message?: any, ...optionalParams: any[]): void {
     if (this.logger) {
@@ -49,12 +46,6 @@ export class Logger implements ModelLoggingInterface {
   setLogger(logger?: ModelLoggingInterface) {
     this.logger = logger;
   }
-
-  public static getInstance(): Logger {
-    if (!Logger.instance) {
-      Logger.instance = new Logger();
-    }
-    return Logger.instance;
-  }
 }
 
+export const Logger: LoggerClass = new LoggerClass();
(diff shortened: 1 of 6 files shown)


src/utils/LoggingInterface.ts
```diff
@@ -0,0 +1,60 @@
...
+  error(message?: any, ...optionalParams: any[]): void;
+}
+
+/**
+ * Logger class for the model generation library
+ * 
+ * This class acts as a forefront for any external loggers which is why it also implements the interface itself.
+ */
+export class Logger implements ModelLoggingInterface {
+  private static instance?: Logger;
```
magicmatatjahu, line 18: I don't think so this is a good idea. We should create default logger by in constructor, and avoid logic like `Logger.getInstance().*`. As developer I rather prefer `Logger.*`

magicmatatjahu (changes requested):

src/utils/LoggingInterface.ts
jonaslagoni: Hmm, good point, let me refactor it.

src/utils/LoggingInterface.ts
```diff
@@ -0,0 +1,51 @@
...
+   * Sets the logger to use for the model generation library
+   * 
+   * @param logger to add
+   */
+  setLogger(logger?: ModelLoggingInterface) {
+    this.logger = logger;
+  }
+}
+
+export const Logger: LoggerClass = new LoggerClass();
```
magicmatatjahu, line 51: Maybe by default we should set `Logger` as `this.logger`? 

```js
const Logger: LoggerClass = new LoggerClass();
Logger.setLogger(Logger);
export { Logger };
```

jonaslagoni: That would create an infinite loop 😄 The logger was designed for others to add their own logger implementation based on their own use case. We don't need to set the logger to itself. Maybe the naming of the class is a bit confusing 🤔?

magicmatatjahu: Ahh ok, I didn't check the infinity loop 😄 then LGTM.

magicmatatjahu (approved):

Issue #172, opened 2021-05-03
Create a CLI


jonaslagoni:
#### Reason/Context
Because this library should not be directly associated with the generator but a separate library we should introduce a CLI for the library.

#### Description
Be aware that this CLI might be part of a larger [CLI](https://github.com/asyncapi/cli/) and should be implemented with extensibility in mind.

#### Related issues
https://github.com/asyncapi/generator-model-sdk/issues/132

magicmatatjahu:
It should be done in the [cli](https://github.com/asyncapi/cli) project :)

derberg:
better create a new one as first we need a discussion on commands structure. 'asyncapi generate' should support generator and modelina as one feature, and depending on different flags you just get what you need

voxpelli:
I kind of would like having a small CLI for just this module as well, but maybe it shouldn't live in this repo.

This module has potentials for use outside of the AsyncAPI ecosystem and would benefit everyone if it isn't too tightly coupled to the AsyncAPI ecosystem (I eg. created #360 now as the API:s I currently am consuming are Swagger API:s, though will eventually hopefully move to AsyncAPI API:s)

jonaslagoni:
Gonna close this one down as it has now been integrated into the AsyncAPI CLI. If that is not enough and we need a small custom CLI for this library alone, then please create another issue 🙂

PR #173, opened 2021-05-03
test: adds blackbox testing

### LINKED ISSUE #101: Rendered models should be compiled as a blackbox test
#### Reason/Context
At the moment we do not directly test the correctness of the output. So when we receive PRs changing the rendering in any way we have no way of knowing whether the output can actually compile.

We already encountered this issue and having something in place to handle this logic would be awesome. 

#### Description
I would suggest we implement a GitHub action per language we output to. For now, this means we would create a workflow for Java, Typescript, and JS.

Questions that need to be answered:
- Should this workflow run for each PR?
- Should we create one global input file which covers all possible values the output can be?
- Say we scale to 10 languages, how much time would that affect the "wait time" for checks to complete?
- What about different versions of languages, should that be tested?

jonaslagoni:
**Description**
This PR is an initial take on adding blackbox testing to the repository. 
- Adds 1 workflow file where all languages are added as a job (`.github/workflows/blackbox-testing.yml`)
- Since we have no CLI directly (#172), we will have to write the output to files on our own, hence why there is 3 `renderer.js` files. These will be removed once we solve that issue which will make it more maintainable.
- Since we are to re-create the output without the CI I have added the generation of the output as scripts to the package. `ci:blackbox:<language>:test` to test the output, `ci:blackbox:<language>:generate` to generate the output and `ci:blackbox:<language>` to execute both scripts. This will require additional tools installed on the developer's machine to run, such as JDK, etc once we get more outputs.

See this to inspect the output: https://github.com/jonaslagoni/generator-model-sdk/actions/runs/812886851

Questions before moving forward:
1. What file should we test with? In this case I simply added our JSON Schema representation for AsyncAPI 2.0.0 which is a somewhat extreme case which we want to support in the end, but currently fails 👀 (https://github.com/asyncapi/generator-model-sdk/issues/176)
    > For now we use the schema payload of https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml because of #199 
2. Do you think this setup is maintainable long-term?
3. Should this workflow run on every PR? 
    > Yes, for now.

**Related issue(s)**
Fixes #101


jonaslagoni:
cc @magicmatatjahu @derberg @smoya

test/CI-blackbox/java/renderer.js
```diff
@@ -0,0 +1,22 @@
+
+const {JavaGenerator} = require('../../../lib');
+const fs = require('fs');
```
derberg, line 3: you can do this, just a hint
```suggestion
const fsp = require('fs').promises;
```

test/CI-blackbox/java/renderer.js
```diff
@@ -0,0 +1,22 @@
...
+const path = require('path');
+/**
+ * Render all schema models
+ * @returns 
+ */
+async function renderModels() {
+  const input = JSON.parse(fs.readFileSync(path.resolve(__dirname, '../test.json')));
+  const generator = new JavaGenerator();
+  const generatedModels = await generator.generate(input);
+  const outputDir = path.resolve(__dirname, './output');
```
derberg, line 13: how this will work when you run tests on local if the output will be already there? wouldn't there be a risk for some false/positives? In generator I did this https://github.com/asyncapi/generator/blob/master/test/integration.test.js#L18

derberg: I guess that in general this part about creating output, you could move to a helper as I can see you have the same in every test

derberg (commented):
I left few comments, looks good so far

> Should this workflow run on every PR?

It should run on every PR

> What file should we test with? In this case I simply added our JSON Schema representation for AsyncAPI 2.0.0 which is a somewhat extreme case which we want to support in the end, but currently fails 👀 (#176)

Maybe long term we should run it against a selection of schemas stored on schemastore https://www.schemastore.org/json/?

> Do you think this setup is maintainable long-term?

LGTM. I'm just curious why you run it in such custom way, with custom triggers and not as jest tests as I do in case of similar tests for generator? In the end you are not really assessing if the file was created.

---
last but not least, why you do not think it is necessary to check if the content of the file is correct? what other tests cover this?

jonaslagoni:
Thanks for the review @derberg!

> Maybe long term we should run it against a selection of schemas stored on schemastore https://www.schemastore.org/json/?

Could be awesome yea 👍 

> LGTM. I'm just curious why you run it in such custom way, with custom triggers and not as jest tests as I do in case of similar tests for generator? In the end you are not really assessing if the file was created.

While I don't check whether the output is as expected I do test that it can be compiled/interpreted/transpiled depending on the language. For Java I compile the output file as my way of checking if it was rendered properly. For TS I first transpile the output file for then to run it with node. For JS I just run the output file with node.

> LGTM. I'm just curious why you run it in such custom way, with custom triggers and not as jest tests as I do in case of similar tests for generator? In the end you are not really assessing if the file was created.
>
> last but not least, why you do not think it is necessary to check if the content of the file is correct? what other tests cover this?

The reason why I did not want to compare output is for the very reason of extending it later to something like schemastore. It would be a nightmare trying to keep the expected output synchronized... We already have tests for the generator to ensure output is as expected for smaller schemas. That together with the unit tests should catch potential problems together with this blackbox testing.

The reason why I use the custom `renderer`s is as stated only until a CLI can take over. Then those files will be removed completely.

test/CI-blackbox/java/renderer.js
jonaslagoni: Yea you are right on both points, let me see if I can't make them use the same function.

derberg:
@jonaslagoni 

>Maybe long term we should run it against a selection of schemas stored on schemastore https://www.schemastore.org/json/?

Just recently I had a discussion with JSON Schema folks about their policy for accepting validators on the list of tools. What poped up during the conversation was that they have some tests they run against validators. Don't know the details but maybe they are similar to our TCK, so using them here would make a lot of sense, I mean those test json schema files.

Side note, we should get here https://json-schema.org/implementations.html on the list too!

> While I don't check whether the output is as expected I do test that it can be compiled/interpreted/transpiled depending on the language. For Java I compile the output file as my way of checking if it was rendered properly. For TS I first transpile the output file for then to run it with node. For JS I just run the output file with node.

yeah, but why not use jest as a runner here but instead have 10 custom npm scripts 😄 

> It would be a nightmare trying to keep the expected output synchronized

Why? :trollface: 

but no, seriously, why? you can fix your test to a given version of the document. Output can easily be compared with snapshot testing. Refreshing those is also not a big deal. Are you sure it can be a problem, I mean you already tried? not a nightmare in the generator so far.

> CLI can take over

is CLI right behind the corner? I think it is not something that will show up quickly. More important is to have it in AsyncAPI CLI first. Not sure it makes sense to spend time on CLI for the model generator now. I'm all yours if it comes to grow a wider community around it, separate from AsyncAPI, but first, let's see if people actually will use it in other contexts. Just a suggestion.

jonaslagoni:
@derberg 

> Just recently I had a discussion with JSON Schema folks about their policy for accepting validators on the list of tools. What poped up during the conversation was that they have some tests they run against validators. Don't know the details but maybe they are similar to our TCK, so using them here would make a lot of sense, I mean those test json schema files.
> 
> Side note, we should get here https://json-schema.org/implementations.html on the list too!

Yep, I read it, I think you are referring to https://github.com/asyncapi/generator-model-sdk/issues/177. But yes, maybe we can extend this to the blackbox testing 🤔 

> Why? :trollface:
> 
> but no, seriously, why? you can fix your test to a given version of the document. Output can easily be compared with snapshot testing. Refreshing those is also not a big deal. Are you sure it can be a problem, I mean you already tried? not a nightmare in the generator so far.

Having a snapshot of more than 1 schema would IMO be a nightmare because the number of changes it would introduce in a simple PR would just clutter the changes.

I am all for having at least one file tested against some expected output. But when we are talking about supporting a wide range of files it simply becomes too much. It would have the same ripple effect as we have now with the simplifier, and it is a big problem https://github.com/asyncapi/generator-model-sdk/issues/64.

derberg:
> Yep, I read it, I think you are referring to #177. But yes, maybe we can extend this to the blackbox testing

you, cool, you spy 😄 

> Having a snapshot of more than 1 schema would IMO be a nightmare because the number of changes it would introduce in a simple PR would just clutter the changes.

not sure what your mean really 😄 changes in snapshots should be required only if you change the structure of the object that you use for testing. Link in generator you need to update snapshot tests only when you actually make modifications to the dummy.yml. In other cases, if snapshot test fail, it means you really need to look into it. 

I'm not much aware about your issues with simplified or processor tests. I might just misunderstand your concerns.

jonaslagoni:
@derberg 
> not sure what your mean really 😄 changes in snapshots should be required only if you change the structure of the object that you use for testing. Link in generator you need to update snapshot tests only when you actually make modifications to the dummy.yml. In other cases, if snapshot test fail, it means you really need to look into it.
> 
> I'm not much aware about your issues with simplified or processor tests. I might just misunderstand your concerns.

No, the structure of the object used for testing is not the only thing that will force it. The difference between this library and the generator is that in the generator you force a specific version of the template. This means that your output never really changes or should not change.

With this library, we are a template (kinda) and this means that any changes to the output, in any way, will require us to update and inspect the snapshots. Having > 1 snapshots that frequently needs to be updated would be annoying in the long run. 

The files we test with will not change, but the generated output will, especially with the stage the library is in at the moment.

That's why I am suggesting we just test whether the output can be "compiled". This way it will not cause us any headache but provide a solid safetynet. Then we can extend something like https://github.com/asyncapi/generator-model-sdk/blob/master/test/blackbox/AsyncAPI.spec.ts with a snapshot.

derberg:
@jonaslagoni I get your point. Just my view on the subject is that templates also need snapshot tests. I have them in the template for generator templates and would expect long term we have them everywhere really as there is simply no other and better way to do it imho. But yeah very good snapshot testing is something that is a requirement for me for 1.0.0, but when you constantly change as you are way before 1.0.0, then snapshot testing can just make you cry, like literally 😄

package.json
```diff
@@ -68,7 +68,17 @@
...
+    "ci:blackbox": "npm run ci:blackbox:js && npm run ci:blackbox:ts && ci:blackbox:java",
+    "ci:blackbox:js": "npm run ci:generate:blackbox:js && npm run ci:test:blackbox:js",
+    "ci:generate:blackbox:js": "node ./test/CI-blackbox/js/renderer.js",
+    "ci:test:blackbox:js": "node ./test/CI-blackbox/js/output/output.js",
+    "ci:blackbox:ts": "npm run ci:generate:blackbox:ts && npm run ci:test:blackbox:ts",
+    "ci:generate:blackbox:ts": "node ./test/CI-blackbox/ts/renderer.js",
+    "ci:test:blackbox:ts": "tsc ./test/CI-blackbox/ts/output/output.ts --outfile ./test/CI-blackbox/ts/output/output.js && node ./test/CI-blackbox/ts/output/output.js",
+    "ci:blackbox:java": "npm run ci:generate:blackbox:java && npm run ci:test:blackbox:java",
+    "ci:generate:blackbox:java": "node ./test/CI-blackbox/java/renderer.js",
+    "ci:test:blackbox:java": "javac ./test/CI-blackbox/java/output/output.java && java -cp ./test/CI-blackbox/java/output/ output"
```
magicmatatjahu, line 81: Maybe having these scripts we should write single nodejs script and run it like:

```bash
npm run ci:blackbox -- js,ts,java
```

where `js,ts,java` are a languages

What do you think? Then we can add new language in this nodejs script.

jonaslagoni: Could be an alternative 🤔 I will look into that!

jonaslagoni: Added a small CLI to reduce code

test/CI-blackbox/renderer.js
```diff
@@ -0,0 +1,52 @@
+
+const {JavaGenerator, JavaScriptGenerator, TypeScriptGenerator} = require('../../lib');
+const fs = require('fs');
```
magicmatatjahu, line 3:
```js
const fs = require('fs').promises;
```

test/CI-blackbox/renderer.js
```diff
@@ -0,0 +1,52 @@
...
+const path = require('path');
+const languagesToInclude = process.argv.slice(2);
+/**
+ * Processes the input and renders the language the models
+ */
+async function processInput() {
+  const input = JSON.parse(fs.readFileSync(path.resolve(__dirname, './test.json')));
```
magicmatatjahu, line 10: If you put `input` inside function, you can also put `languagesToInclude` or move `input` outside the function.

test/CI-blackbox/renderer.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {*} models 
+ * @param {*} outputPath path to output
+ */
+async function renderModels(models, outputPath) {
+  const outputDir = path.resolve(__dirname, path.dirname(outputPath));
+  await fs.promises.mkdir(outputDir, { recursive: true });
+  const output = [];
+  for (const generatedModel of models) {
+    output.push(generatedModel.result);
+  }
```
magicmatatjahu, line 47: You can use map here, but it's also good for me :)

magicmatatjahu (commented):
Only few comments for CLI :) Great work!

test/CI-blackbox/renderer.js
jonaslagoni: Good call 👍

test/CI-blackbox/renderer.js
jonaslagoni: Gonna keep everything inside `processInput` 🙂

test/CI-blackbox/java/renderer.js
jonaslagoni: So I moved everything to a single renderer. Regarding the naming, there is not since everything is rendered to a single file (`output.(js | java | ts)` and overwrite that file completely `false/positives` should not be possible.

jonaslagoni:
Now the question becomes, do we wanna merge this before or after the bugs are fixed? 🤔 https://github.com/asyncapi/generator-model-sdk/issues/176

jonaslagoni:
Alright I have added a simpler input file which for now is the schema payload of https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml because of #199 we cant use the direct AsyncAPI file.

I added some extra documentation explaining the pipeline, think this PR is ready for the final review 🙂

package.json
```diff
@@ -68,7 +68,17 @@
...
+    "ci:blackbox": "npm run ci:blackbox:js && npm run ci:blackbox:ts && npm run ci:blackbox:java",
+    "ci:blackbox:js": "npm run ci:blackbox:js:generate && npm run ci:blackbox:js:test",
+    "ci:blackbox:js:generate": "node ./test/CI-blackbox/renderer.js js",
+    "ci:blackbox:js:test": "node ./test/CI-blackbox/js/output.js",
+    "ci:blackbox:ts": "npm run ci:blackbox:ts:generate && npm run ci:blackbox:ts:test",
+    "ci:blackbox:ts:generate": "node ./test/CI-blackbox/renderer.js ts",
+    "ci:blackbox:ts:test": "tsc -t es5 ./test/CI-blackbox/ts/output.ts && node ./test/CI-blackbox/ts/output.js",
+    "ci:blackbox:java": "npm run ci:blackbox:java:generate && npm run ci:blackbox:java:test",
+    "ci:blackbox:java:generate": "node ./test/CI-blackbox/renderer.js java",
+    "ci:blackbox:java:test": "javac ./test/CI-blackbox/java/*.java"
```
magicmatatjahu, line 81: Do we really need these scripts? If we add another language then what, new scripts? 😅 Maybe we should use this trick with `npm run blackbox -- {languages}`?

magicmatatjahu:
> 1. What file should we test with? In this case I simply added our JSON Schema representation for AsyncAPI 2.0.0 which is a somewhat extreme case which we want to support in the end, but currently fails 👀 (#176)
For now we use the schema payload of https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml because of #199

We can start with testing simple cases and then add complex parts.

> 2. Do you think this setup is maintainable long-term?

Due to additional npm scripts, no -> see my comment.

> 3. Should this workflow run on every PR?

For me, yes.

package.json
jonaslagoni: Really didnt want to add my own sub-process spawn in the renderer... What if we refactored it to 
```
    "ci:blackbox:generator": "node ./test/CI-blackbox/renderer.js",
    "ci:blackbox:ts:test": "tsc -t es5 ./test/CI-blackbox/ts/output.ts && node ./test/CI-blackbox/ts/output.js",
    "ci:blackbox:java:test": "javac ./test/CI-blackbox/java/*.java"
```

magicmatatjahu (approved):

test/CI-blackbox/renderer.js
```diff
@@ -0,0 +1,78 @@
...
+        break;
+        case 'ts':
+          generator = new TypeScriptGenerator();
+          outputPath = './ts/output.ts';
+        break;
+        default:
+          throw new Error(`Does not understand language input ${language}`);
+      }
+      const generatedModels = await generator.generate(input);
+      await renderModels(generatedModels, outputPath);
```
magicmatatjahu, line 40: You can put it in one switch :)

jonaslagoni: True, It just looks weird when JS and TS works differently from Java 😄

jonaslagoni:
@derberg do you want to review it as well? 🤔

derberg:
@jonaslagoni nah, I saw @magicmatatjahu did review

I'll just still super curious why you picked custom scripts over jest as a runner

> yeah, but why not use jest as a runner here but instead have 10 custom npm scripts 😄

you never replied and I'm super curious what is the reason :D

package.json
magicmatatjahu: It's better that previous, but as Łukasz wrote, maybe we should switch to Jest runner? Probably it will simplify scripts only for TS and JS models, not for Java, but maybe... what do you think? 😄

jonaslagoni:
> I'll just still super curious why you picked custom scripts over jest as a runner

My reason for not going with a Jest test was because it would be "too" complex for what it currently should achieve (generate -> compile/transpile) which basically lay in that I didn't want to add my own sub-process spawn.

But if we think Jest runner is gonna be more convenient even now then I can look into changing it 🙂

jonaslagoni:
@derberg @magicmatatjahu is that thumbs up as in "okay fair reasoning, let's keep it as it is now" or "Good idea to look into changing it to runners" 😆

derberg:
@jonaslagoni 😆 for me 👍🏼 means it is up to you. You know the best if you want to invest some more time in Jest do it, if you really feel it is not worth it and will actually cause more problems, then don't. The fact that I think it should be in Jest instead of custom scripts does not mean this is a better solution 😄 It is just my personal feeling but I did not try it on my own and don't know it that would be better, I just do not like to grow custom scripts if not needed 😆

magicmatatjahu:
As derberg said, if Jest will create problems, please don't invest time for that, but if you want, go ahead. It can be also done as followup as little POC :)

jonaslagoni:
@derberg @magicmatatjahu was actually not too bad in terms of implementing the process in Jest runners. I feel like this is better 👍 

While excluding black box testing from the coverage it has decreased below the threshold, however, I feel like this should be the correct behavior, and coverage should only cover specific tests. Thoughts?

docs/development.md
```diff
@@ -0,0 +1,23 @@
...
+3. Make sure all tests pass `npm test`
+4. Make sure code is well formatted and secure `npm run lint`
+
+## CI Blackbox testing
+In each PR there will be running a CI pipeline which ensures that this [dummy.json](../test/CI-blackbox/dummy.yml) file can produce a valid output file. This means all generated models will be written to files and then transpiled/compiled depending on the output language. 
+
+If any of those CI jobs fail this is how to reproduce the output.
+
+For each output language (`ts`, `js`, `java`) there will be the following scripts:
+1. `ci:blackbox:<language>:generate` which generates the output
```
derberg, line 16: I think you need to update this

derberg:
> While excluding black box testing from the coverage it has decreased below the threshold, however, I feel like this should be the correct behavior, and coverage should only cover specific tests. Thoughts?

I tried to quickly have a look and tbh I do not know why it dropped. It happened on a run when you merged the latest master.

jonaslagoni:
> > While excluding black box testing from the coverage it has decreased below the threshold, however, I feel like this should be the correct behavior, and coverage should only cover specific tests. Thoughts?
> 
> I tried to quickly have a look and tbh I do not know why it dropped. It happened on a run when you merged the latest master.

@derberg It dropped because I exclude the blacbox tests from coverage. The question is, should it be excluded 🧐?

derberg:
I think so, it is not relevant for coverage imho

docs/development.md
magicmatatjahu: Probably `CI Blackbox testing` and `Requirements` should be updated.

package.json
```diff
@@ -56,7 +56,8 @@
   "scripts": {
     "start": "tsc --watch",
     "build": "tsc",
-    "test": "cross-env CI=true jest --coverage",
+    "test": "cross-env CI=true jest --coverage --testPathIgnorePatterns ./test/blackbox",
+    "test:blackbox": "cross-env CI=true jest --testTimeout=100000 ./test/blackbox",
```
magicmatatjahu, line 60: You can add timeout also as 3rd argument in `test()` function :)

```js
test('should be generated in TypeScript', async function() {}, {timeout})
```

test/blackbox/utils/Utils.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Exec a command and if any errors occur reject the promise.
+ * 
+ * @param absolutePathToFile 
+ * @param generator  
+ */
+ export async function generateModels(absolutePathToFile: string, generator: AbstractGenerator) {
+  const inputFileContent = await fs.readFile(absolutePathToFile);
+  const input = JSON.parse(String(inputFileContent));
+  return await generator.generate(input);
```
magicmatatjahu, line 17: If you have async function which returns some result from promise, then last `await` is unnecessary, because finally you return Promise. It's only a tip, `return await` is not a problem for me :)

test/blackbox/utils/Utils.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export async function renderModels(generatedModels: OutputModel[], outputPath: string) {
+  const outputDir = path.resolve(__dirname, path.dirname(outputPath));
+  await fs.rm(outputDir, { recursive: true, force: true });
+  await fs.mkdir(outputDir, { recursive: true });
+  const output = generatedModels.map((model) => {
+    return model.result;
+  });
+  const outputFilePath = path.resolve(__dirname, outputPath);
+  await fs.writeFile(outputFilePath, output.join('\n'));
```
magicmatatjahu, line 66: If you have async function which returns some result from promise, then last `await` is unnecessary, because finally you return Promise. It's only a tip, `return await` is not a problem for me :)

jonaslagoni: Gonna leave it for this function, more clean to look at when it returns void.

package.json
jonaslagoni: Don't want to add that for each test 😄 Think it is better to keep it global. It is also only the CI that complains about timing out.

jonaslagoni: I could maybe do
```
jest.setTimeout();
```

jonaslagoni: Changed 😄

docs/development.md
jonaslagoni: I completely forgot about the docs, sorry 😅 

I have changed it.

derberg:
did you notice sonar complaining on code duplication https://sonarcloud.io/component_measures?id=asyncapi_generator-model-sdk&pullRequest=173&metric=new_duplicated_lines_density&view=list ?

test/blackbox/Dummy.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  describe('should be able to generate and transpile TS', () => {
+    test('class', async function() {
+      const generator = new TypeScriptGenerator();
+      const generatedModels = await generateModels(fileToGenerate, generator);
+      expect(generatedModels).not.toHaveLength(0);
+      const renderOutputPath = path.resolve(__dirname, './output/ts/output.ts');
+      await renderModels(generatedModels, renderOutputPath);
+      const transpiledOutputPath = path.resolve(__dirname, './output/ts/output.js');
+      const transpileAndRunCommand = `tsc -t es5 ${renderOutputPath} && node ${transpiledOutputPath}`;
```
magicmatatjahu, line 30: We can use es6. Who currently use the es5 in node project?! 😄

magicmatatjahu (dismissed):
LGTM :)

magicmatatjahu:
@derberg Note that the duplications are not accidental in this case and separating them into separate functions will only create an additional difficult understandable abstraction.

test/blackbox/Dummy.spec.ts
jonaslagoni: True, let's expand the tests later 😄

derberg:
I'm not questioning your code folks 😄 if duplication is intentional I bet you had a good reason. What is important though is to solve the alert. Either refactor or silent the rule with clear description. If alert is there forever, everyone gets used to and stop taking care

magicmatatjahu (approved):
LGTM

PR #183, opened 2021-05-12
refactor: add support functions for CommonModel

jonaslagoni:
**Description**
This PR introduces some more functions to `CommonModel`, which are going to be used as part of the further refactoring of the simplifier.

Furthermore, I added functionality to ensure that recursive `CommonModel`s are handles correctly when merging. With the current simplifier, this is not a problem so it is not a `fix:`.

### DIFF
--- a/src/models/CommonModel.ts
+++ b/src/models/CommonModel.ts
@@ -133,9 +133,11 @@ export class CommonModel extends CommonSchema<CommonModel> {
   }
   /**
    * Merge two common model properties together 
+   * 
    * @param mergeTo 
    * @param mergeFrom 
    * @param originalSchema 
+   * @param alreadyIteratedModels 
    */
   private static mergeProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
     const mergeToProperties = mergeTo.properties;
--- a/src/models/CommonModel.ts
+++ b/src/models/CommonModel.ts
@@ -28,7 +28,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
    */
   getFromSchema<K extends keyof Schema>(key: K) {
     let schema = this.originalSchema || {};
-    if (typeof schema === 'boolean') schema = {};
+    if (typeof schema === 'boolean') return undefined;
     return schema[`${key}`];
   }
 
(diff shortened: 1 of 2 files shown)


src/models/CommonModel.ts
```diff
@@ -20,6 +20,102 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+      if (additionalPropertiesRef !== undefined) {
+        dependsOn.push(additionalPropertiesRef);
+      }
+    }
+    if (this.extend !== undefined) {
+      for (const extendedSchema of this.extend) {
+        dependsOn.push(extendedSchema);
+      }
+    }
+    if (this.items instanceof CommonModel) {
```
magicmatatjahu, line 103: I don't remember, but do we support `items` as array? Should we also iterate on `items` array?

src/models/CommonModel.ts
```diff
@@ -188,73 +296,16 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+    if (mergeFrom.$id !== undefined && mergeTo.$id === undefined) {
       mergeTo.$id = mergeFrom.$id;
     }
-    if (mergeFrom.$ref !== undefined) {
+    if (mergeFrom.$ref !== undefined && mergeTo.$ref === undefined) {
       mergeTo.$ref = mergeFrom.$ref;
     }
-    if (mergeFrom.extend !== undefined) {
+    if (mergeFrom.extend !== undefined && mergeTo.extend === undefined) {
       mergeTo.extend = mergeFrom.extend;
```
magicmatatjahu, line 306: You can make here trick like:

```ts
mergeTo.$id = mergeTo.$id || mergeFrom.$id;
```

Also by this you don't need to write ifs

magicmatatjahu (commented):
Few comments :) Great job! Also, please change `chore:` to `refactor:` :)

src/models/CommonModel.ts
jonaslagoni: Good point, yes! This is actually a bug, gonna create another issue to solve this.

jonaslagoni: https://github.com/asyncapi/generator-model-sdk/issues/186

src/models/CommonModel.ts
jonaslagoni: Good call, looks cleaner :)

magicmatatjahu (approved):

jonaslagoni:
Thanks 🎉

PR #184, opened 2021-05-12
refactor: introduce new interpreter

jonaslagoni:
**Description**
This PR adds the initial refactoring of the simplification process and renames it to interpretation.

It changes the entire perspective, to instead of adding simplification functions per `CommonModel` property, we process the JSON Schema keywords which then affects the `CommonModel` in some way.

**It adds**
- Interpreter, the new interpretation, which is quite similar to the old but different in how it simplifies the models.
- Properties interpretation, to show how the general principle works.

- https://github.com/asyncapi/generator-model-sdk/blob/master/src/simplification/Simplifier.ts
- https://github.com/asyncapi/generator-model-sdk/blob/master/src/simplification/SimplifyProperties.ts
- And all [related tests](https://github.com/asyncapi/generator-model-sdk/tree/master/test/simplification) have been [refactored](https://github.com/jonaslagoni/generator-model-sdk/tree/feature/initial_simplifier/test/newsimplification) to have next to no dependencies by using the jest mocking framework.

**Questions**
- Should we just use the general naming of `simplification`? I think it makes more sense to call it `JSON Schema draft 7 simplifications`. Where `simplification` refers to the process of interpreting how validation rules should be converted to `CommonModels`. Other words we can use instead of `simplification`? `interpretation`? `processing`? Not sure what fits best 🤷 
    - We are gonna rename it to `interpretation`.

**Todo**

- [x] Refactor docs
- [x] Solve code smells

**Related issue(s)**
Related to: https://github.com/asyncapi/generator-model-sdk/issues/64
Blocked by: https://github.com/asyncapi/generator-model-sdk/pull/183


jonaslagoni:
@czlowiek488 pinging you in case you want to have a look at the refactoring of the simplifier and have some comments.

src/models/CommonModel.ts
```diff
@@ -20,6 +20,102 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+  }
+
+  /**
+   * Set the types of the model
+   * 
+   * @param types to set the model type to
+   */
+  setType(types : string | string[]) {
+    if (Array.isArray(types)) {
+      if (types.length === 0) {
```
czlowiek488: You may want to unnest if`s like this one with something similar to `if(!Array.isArray(types) { this.types = types; return; }`

src/models/CommonModel.ts
```diff
@@ -20,6 +20,102 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+  }
+
+  /**
+   * Adds types to the existing model types.
+   * 
+   * Makes sure to only keep a single type incase of duplicates.
+   * 
+   * @param types which types we should try and add to the existing output
+   */
+  addToTypes(types: string[] | string) {
```
czlowiek488: I do not like this `|` in arguments, it is not following Single Responsibility Principle.

src/models/CommonModel.ts
```diff
@@ -20,6 +20,102 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+      this.type = types;
+    } else if (!Array.isArray(this.type) && this.type !== types) {
+      this.type = [this.type, types];
+    } else if (Array.isArray(this.type) && !this.type.includes(types)) {
+      this.type.push(types);
+    }
+  }
+
+  /**
+   * Checks if given property name is required in object
```
czlowiek488: imho, you duplicate documentation here, you have typescript types and jsdoc types. If you actually need this for something else I am fine with it.

src/models/CommonModel.ts
```diff
@@ -20,6 +20,102 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+      }
+    }
+    if (this.items instanceof CommonModel) {
+      const itemsRef = (this.items as CommonModel).$ref;
+      if (itemsRef !== undefined) {
+        dependsOn.push(itemsRef);
+      }
+    }
+    if (this.properties !== undefined && Object.keys(this.properties).length) {
+      Object.entries(this.properties).forEach(([, propertyModel]) => {
```
czlowiek488: I guess you could just use `Object.values` instead of `Object.entries`. Also imo it is more readable if you instead of `forEach` use `map`, and then push it with spread like `dependsOne.push(...result)`.

src/models/CommonModel.ts
```diff
@@ -41,7 +137,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
    * @param mergeFrom 
    * @param originalSchema 
    */
-  private static mergeProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema) {
+  private static mergeProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
```
czlowiek488: There is so many arguments here that I would consider passing them in a object. If you would do it this way, you dont have to care about order while developing.

src/models/CommonModel.ts
```diff
@@ -64,14 +160,14 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
-  private static mergeAdditionalProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema) {
+  private static mergeAdditionalProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
     const mergeToAdditionalProperties = mergeTo.additionalProperties;
     const mergeFromAdditionalProperties = mergeFrom.additionalProperties;
     if (mergeFromAdditionalProperties !== undefined) {
-      if (mergeToAdditionalProperties === undefined) {
-        mergeTo.additionalProperties = mergeFromAdditionalProperties;
+      if (mergeToAdditionalProperties !== undefined) {
+        mergeTo.additionalProperties = CommonModel.mergeCommonModels(mergeToAdditionalProperties, mergeFromAdditionalProperties, originalSchema, alreadyIteratedModels);
       } else {
```
czlowiek488, line 194: I saw above that you are using `return` instead of `else` in cases like here. I think it should be consistent.

src/models/CommonModel.ts
```diff
@@ -90,7 +186,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
         for (const [pattern, patternModel] of Object.entries(mergeFromPatternProperties)) {
           if (mergeToPatternProperties[`${pattern}`] !== undefined) {
-            mergeToPatternProperties[`${pattern}`] = CommonModel.mergeCommonModels(mergeToPatternProperties[`${pattern}`], patternModel, originalSchema);
+            mergeToPatternProperties[`${pattern}`] = CommonModel.mergeCommonModels(mergeToPatternProperties[`${pattern}`], patternModel, originalSchema, alreadyIteratedModels);
```
czlowiek488, line 217: I guess that there is DRY rule violation.

src/models/CommonModel.ts
```diff
@@ -106,12 +202,14 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
-  private static mergeItems(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema) {
+  private static mergeItems(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
     const merge = (models: CommonModel | CommonModel[] | undefined): CommonModel | undefined => {
       if (Array.isArray(models)) {
         if (models.length > 0) {
-          let mergedItemsModel: CommonModel = models[0];
-          models.forEach((model) => { mergedItemsModel = CommonModel.mergeCommonModels(mergedItemsModel, model, originalSchema); });
+          let mergedItemsModel: CommonModel | undefined = undefined;
+          models.forEach((model) => { 
+            mergedItemsModel = CommonModel.mergeCommonModels(mergedItemsModel, model, originalSchema, alreadyIteratedModels);
```
czlowiek488: Why do not you just use the last element of an array? It is unclear why this iteration take place.

src/models/CommonModel.ts
```diff
@@ -171,13 +269,15 @@ export class CommonModel extends CommonSchema<CommonModel> {
    * @param mergeFrom CommonModel to merge values from
    * @param originalSchema schema to use as original schema
    */
-  static mergeCommonModels(mergeTo: CommonModel | undefined, mergeFrom: CommonModel, originalSchema: Schema): CommonModel {
+  static mergeCommonModels(mergeTo: CommonModel | undefined, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()): CommonModel {
     if (mergeTo === undefined) return mergeFrom;
```
czlowiek488, line 302: I would consider opening `{}` even for simple if statements.

src/models/CommonModel.ts
```diff
@@ -188,73 +288,16 @@ export class CommonModel extends CommonSchema<CommonModel> {
     }
 
     // Which values are correct to use here? Is allOf required?
```
czlowiek488: Do not you think that this question should be resolved?

test/models/CommonModel.spec.ts
```diff
@@ -147,6 +147,18 @@ describe('CommonModel', function() {
     });
   });
   describe('mergeCommonModels', function() {
+    test('should handle recursive models', function() {
```
czlowiek488: I do not know what are you testing here. I expected assertion.

test/models/CommonModel.spec.ts
```diff
@@ -479,6 +492,31 @@ describe('CommonModel', function() {
       });
     });
   });
+
+  describe('addToTypes', function() {
+    test('should add multiple types', function() {
```
czlowiek488, line 558: You need some many tests because you violated Single Responsibility Principle.

src/models/CommonModel.ts
```diff
@@ -134,10 +134,10 @@ export class CommonModel extends CommonSchema<CommonModel> {
   /**
    * Merge two common model properties together 
    * 
-   * @param mergeTo 
-   * @param mergeFrom 
-   * @param originalSchema 
-   * @param alreadyIteratedModels 
+   * @param mergeTo CommonModel to merge properties into
```
czlowiek488: Personally I do not like this. Imo names of variables should be understandable enough.

src/newsimplification/Simplifier.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private anonymCounter = 1;
+  private seenSchemas: Map<Schema, CommonModel> = new Map();
+  private iteratedModels: Record<string, CommonModel> = {};
+  
+  constructor(
+    readonly options: SimplificationOptions = Simplifier.defaultOptions,
+  ) {
+    this.options = { ...Simplifier.defaultOptions, ...options };
```
czlowiek488, line 19: It is ok for now. But if you would add nested options you would need to change it anyway.

src/newsimplification/Simplifier.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.combineSchemas(schema.else, model, schema);
+  }
+
+  /**
+   * Go through schema(s) and combine the simplified models together.
+   * 
+   * @param schema to go through
+   * @param currentModel the current output
+   */
+  combineSchemas(schema: (Schema | boolean) | (Schema | boolean)[] | undefined, currentModel: CommonModel, rootSchema: Schema) {
```
czlowiek488, line 94: Single Responsibility Principle! Instead of `| undefined` you should use `?`.

src/newsimplification/Simplifier.ts
```diff
@@ -0,0 +1,150 @@
...
+  private iteratedModels: Record<string, CommonModel> = {};
+  
+  constructor(
+    readonly options: SimplificationOptions = Simplifier.defaultOptions,
+  ) {
+    this.options = { ...Simplifier.defaultOptions, ...options };
+  }
+
+  /**
+   * Simplifies a schema into instances of CommonModel.
```
czlowiek488, line 23: I think that it is good idea to have here a detailed description what simplify is actually doing. I do not get this, it is too unclear for me.

src/newsimplification/Simplifier.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+}
+
+/**
+ * This is the default wrapper for the simplifier class which always create a new instance of the simplifier. 
+ * 
+ * @param schema to simplify
+ */
+export function simplify(schema: Schema | boolean, options?: SimplificationOptions): CommonModel[] {
```
czlowiek488, line 147: I guess that it is too much for encapsulation. I am not sure but I think that some of people would say that you violeted KISS rule here. Now you have multiple ways to actually use simplifier.

src/newsimplification/SimplifyProperties.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Simplifier function for finding the simplified version of properties for a model.
+ * 
+ * @param schema the schema to simplify properties for
+ * @param model the model to simplify properties into
+ * @param simplifier the simplifier instance 
+ */
+// eslint-disable-next-line sonarjs/cognitive-complexity
+export default function simplifyProperties(schema: Schema | boolean, model: CommonModel, simplifier : Simplifier) {
+  if (typeof schema !== 'boolean' && schema.properties !== undefined) {
```
czlowiek488, line 13: You can easly unnest this if statement.

test/newsimplification/Simplifier.spec.ts
```diff
@@ -0,0 +1,147 @@
+import {simplify, Simplifier} from '../../src/newsimplification/Simplifier';
+import {isModelObject, simplifyName} from '../../src/newsimplification/Utils';
+import simplifyProperties from '../../src/newsimplification/SimplifyProperties';
+import simplifyAllOf from '../../src/newsimplification/SimplifyAllOf';
+import { CommonModel, Schema } from '../../src/models';
+
+let mockedIsModelObjectReturn = false;
+jest.mock('../../src/newsimplification/Utils', () => {
```
czlowiek488, line 7: Jest is commonly used in frontend but it is heavy, consider using mocha for backend tests.

test/newsimplification/Simplifier.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    const schema: any = true;
+    const models = simplify(schema);
+    expect(models).toHaveLength(1);
+    expect(models[0].type).toEqual(['object', 'string', 'number', 'array', 'boolean', 'null', 'integer']);
+  });
+  test('should set id of model if object', function() {
+    const schema = { type: 'object' };
+    const models = simplify(schema);
+    expect(models).toHaveLength(1);
+    expect(simplifyName).toHaveBeenNthCalledWith(1, schema);
```
czlowiek488, line 45: I have no idea from simplifyName is coming from. I do not like this.

test/newsimplification/Utils.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      model.type = ['object', 'string'];
+      expect(isModelObject(model)).toEqual(true);
+    });
+    test('should return false if contains all types', function () {
+      const model = new CommonModel();
+      model.type = ['object', 'string', 'number', 'array', 'boolean', 'null', 'integer'];
+      expect(isModelObject(model)).toEqual(false);
+    });
+    test('should return false if type is not defined', function () {
+      const model = new CommonModel();
```
czlowiek488, line 33: I prefer to see AAA pattern(rule?) in tests.

src/models/CommonModel.ts
```diff
@@ -57,10 +57,10 @@ export class CommonModel extends CommonSchema<CommonModel> {
    * 
    * @param types which types we should try and add to the existing output
    */
-  addToTypes(types: string[] | string) {
+  addTypes(types: string[] | string) {
     if (Array.isArray(types)) {
       types.forEach((value) => {
-        this.addToTypes(value);
+        this.addTypes(value);
```
czlowiek488, line 64: I highly recommend you to stop doing mutation. For this project it may too late.

src/models/CommonModel.ts
```diff
@@ -84,6 +84,23 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+   * If the property already exist the two are merged.
+   * 
+   * @param propertyName 
+   * @param propertyModel 
+   * @param schema schema to the corresponding property model
+   */
+  addProperty(propertyName: string, propertyModel: CommonModel, schema: Schema) {
+    if (this.properties === undefined) this.properties = {};
+    if (this.properties[`${propertyName}`] !== undefined) {
+      this.properties[`${propertyName}`] = CommonModel.mergeCommonModels(this.properties[`${propertyName}`], propertyModel, schema);
```
czlowiek488, line 100: Can not you replace 
```
[`${propertyName}`]
```
with
```
[propertyName]
```
?
I am not sure why are you parsing it to a string if it is already a string...

src/models/CommonModel.ts
```diff
@@ -147,11 +164,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
         mergeTo.properties = mergeFromProperties;
       } else {
         for (const [propName, prop] of Object.entries(mergeFromProperties)) {
```
czlowiek488, line 168: I do not like that you are inconsistent. Sometimes you use `forEach` sometimes `for of` imo you should stick to `forEach`.
Also I think that you should always pass an array of properties into .addProperty method. It would be more DRY.

src/newsimplification/SimplifyProperties.ts
```diff
@@ -9,23 +9,14 @@ import { Simplifier } from './Simplifier';
...
-// eslint-disable-next-line sonarjs/cognitive-complexity
 export default function simplifyProperties(schema: Schema | boolean, model: CommonModel, simplifier : Simplifier) {
   if (typeof schema !== 'boolean' && schema.properties !== undefined) {
-    //Ensure object type is inferred
-    model.addToTypes('object');
+    model.addTypes('object');
 
     for (const [propertyName, propertySchema] of Object.entries(schema.properties)) {
       const propertyModels = simplifier.simplify(propertySchema);
       if (propertyModels.length > 0) {
```
czlowiek488, line 18: I guess that this if may be in `addProperty` it may be more DRY. I am not so sure in this case.

src/models/CommonModel.ts
```diff
@@ -164,7 +164,11 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
-          mergeTo.addProperty(propName, prop, originalSchema); 
+          if (mergeToProperties[`${propName}`] !== undefined) {
```
czlowiek488, line 169: I would consider using this if in `mergeCommonModels` method.

test/newsimplification/SimplifyProperties.spec.ts
```diff
@@ -44,30 +46,12 @@ describe('Simplification of properties', () => {
...
     expect(simplifier.simplify).toHaveBeenNthCalledWith(1, { type: 'string' });
-    expect(CommonModel.mergeCommonModels).toHaveBeenNthCalledWith(1, expect.anything(), expect.anything(), schema);
-    expect(model).toMatchObject(
-      {
-        properties: { 'property1': {} },
-      },
-    );
+    expect(model.addProperty).toHaveBeenNthCalledWith(1, "property1", mockedSimplifierModel, schema);
   });
 });
```
czlowiek488, line 57: I do not know why but there is some reason to add last line everywhere. You can add eslint rule for that.

src/models/CommonModel.ts
```diff
@@ -95,6 +96,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
   addProperty(propertyName: string, propertyModel: CommonModel, schema: Schema) {
     if (this.properties === undefined) this.properties = {};
     if (this.properties[`${propertyName}`] !== undefined) {
+      Logger.warn(`While trying to add property to model, duplicate properties found. Merging models together for property ${propertyName}`, propertyModel, schema, this);
```
czlowiek488, line 99: Pleas ensure that there is a way to disable all log levels. Also from my point of view it should not be a `warn` but a `info`.

src/models/CommonModel.ts
```diff
@@ -165,6 +167,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
       } else {
         for (const [propName, prop] of Object.entries(mergeFromProperties)) {
           if (mergeToProperties[`${propName}`] !== undefined) {
+            Logger.warn(`Found duplicate properties ${propName} for model. Model property from ${mergeFrom.$id || 'unknown'} merged into ${mergeTo.$id || 'unknown'}`, mergeTo, mergeFrom, originalSchema);
```
czlowiek488, line 170: It also should be an info.

src/models/CommonModel.ts
```diff
@@ -186,6 +189,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
     const mergeFromAdditionalProperties = mergeFrom.additionalProperties;
     if (mergeFromAdditionalProperties !== undefined) {
       if (mergeToAdditionalProperties !== undefined) {
+        Logger.warn(`Found duplicate additionalProperties for model. additionalProperties from ${mergeFrom.$id || 'unknown'} merged into ${mergeTo.$id || 'unknown'}`, mergeTo, mergeFrom, originalSchema);
```
czlowiek488, line 192: info

src/models/CommonModel.ts
```diff
@@ -234,20 +234,17 @@ export class CommonModel extends CommonSchema<CommonModel> {
   private static mergeItems(mergeTo: CommonModel, mergeFrom: CommonModel, originalSchema: Schema, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
     const merge = (models: CommonModel | CommonModel[] | undefined): CommonModel | undefined => {
       if (Array.isArray(models)) {
```
czlowiek488, line 236: You can easly unnest this if.

src/models/CommonModel.ts
```diff
@@ -234,20 +234,17 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
-            Logger.warn(`Found duplicate items at index ${index} for model. Model item for ${mergeFrom.$id || 'unknown'} merged into ${mergeTo.$id || 'unknown'}`, mergeTo, mergeFrom, originalSchema);
-            mergedItemsModel = CommonModel.mergeCommonModels(mergedItemsModel, model, originalSchema, alreadyIteratedModels); 
-          });
-          return mergedItemsModel;
-        } 
-        return undefined;
+        let mergedItemsModel: CommonModel | undefined = undefined;
+        models.forEach((model, index) => { 
+          Logger.warn(`Found duplicate items at index ${index} for model. Model item for ${mergeFrom.$id || 'unknown'} merged into ${mergeTo.$id || 'unknown'}`, mergeTo, mergeFrom, originalSchema);
+          mergedItemsModel = CommonModel.mergeCommonModels(mergedItemsModel, model, originalSchema, alreadyIteratedModels);
```
czlowiek488, line 240: Why are you iterating over whole array but using only the last result.? It is unclear for me. Please make it more clear.

src/models/CommonModel.ts
czlowiek488: Others should be reconsidered too.

jonaslagoni:
@czlowiek488 damn what a review, I will go through each of them and address them, thank you 🙏

src/models/CommonModel.ts
jonaslagoni: I don't see how this violates SRP. 

The function's single responsibility is to add type(s) to the model, we just allow one or more types to be added.

Why do you say it violates the SRP? 🤔

src/models/CommonModel.ts
jonaslagoni: We use it to provide a generated API file for the library - https://github.com/asyncapi/generator-model-sdk/blob/master/API.md and it does not imply the type from TS, unfortunately.

src/models/CommonModel.ts
jonaslagoni: I agree with `Object.values` is better 👍 

Not sure about map since we have to filter it 🤔 Guess it comes down to preference. Your suggested approach can be:

```ts
    if (this.properties !== undefined && Object.keys(this.properties).length) {
      const referencedProperties = Object.values(this.properties)
        .filter((propertyModel) => {return propertyModel.$ref !== undefined;})
        .map((propertyModel) => {return `${propertyModel.$ref}`;});
      dependsOn.push(...referencedProperties);
    }
```

Do you think this is better? 🤔

czlowiek488:
```
if (this.properties !== undefined && Object.keys(this.properties).length) { 
const referencedProperties = Object.values(this.properties)
.filter(propertyModel =>  propertyModel.$ref !== undefined) // you do not have to open braces in arrow function if you just return
.map(propertyModel => propertyModel.$ref); // is not $ref already a string
dependsOn.push(...referencedProperties); 
}
```

Yes I do, because you do not mutate an object inside a scope. Now if you move this part to a separete function
```
Object.values(this.properties)
.filter(propertyModel =>  propertyModel.$ref !== undefined) // you do not have to open braces in arrow function if you just return
.map(propertyModel => propertyModel.$ref);
```
unit test is easier to make because you do not modify internal state of class inside of this function.

jonaslagoni: True, need to remember JS have multiple ways to simplify the syntax 😄

src/models/CommonModel.ts
czlowiek488: Because you have multiple ways to execute this function and both of them give you same result. Also you need to add unit test to cover both cases. Few weeks ago colleague of mine ask me about exactly same case.

Now you need to write 2 tests and add a logic to handle both cases. But what I prefer to do is to pass even single element in the array.
So instead of calling it this way
```
addToTypes(SOME_TYPE);
```
you may want to do
```
addToTypes([SOME_TYPE]);
```

czlowiek488: SRP means that a function should be responsible only for smallest possible bit of whole process. This function above is responsible for 2 of them 
 - adding single element
 - adding array of elements
 
You may see in some libraries that this is something allowed. Yeah it is but those libraries are written and C++ (or violate SRP) where you can do arguments overload based on types of arguments.

src/models/CommonModel.ts
czlowiek488: I have seen generators for something like that but I never actually had to use them. Nevertheless you may find it usefull.

src/models/CommonModel.ts
czlowiek488: Personally I would write something like this.
```
const getReferences = (properties) => {
if(properties === undefined) {
return [];
}
const referenceNames = Object.values(properties);
if(referenceNames.length === 0) {
return [];
}
return referenceNames 
.filter(propertyModel =>  propertyModel.$ref !== undefined)
.map(propertyModel => propertyModel.$ref)
}
```
```
dependsOn.push(...getReferences(this.properties));
```
Now you can make unit test which cover more cases. Also it will reduce size of function where getReferences is called.

src/models/CommonModel.ts
jonaslagoni: Yea I see your point, also in terms of https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632272973. 

To me, it comes down to whether or not you want cleaner syntax i.e. `addToTypes(SOME_TYPE);` over `addToTypes([SOME_TYPE]);` by adding a single test to support it (since we use recursion). Cause In terms of coverage that is all it takes. 

Especially since TS does not allow overloading functions, if this was Java I would have split it into two functions to allow both parameters.

So this comes down to preference IMO, lets see what others think.

cc. @magicmatatjahu

src/models/CommonModel.ts
jonaslagoni: I tried and I agree, also I think it is time for the merge functions to get their own file. However, that is too much of a refactor in this PR (trying to keep it as small as possible). Would you mind adding an issue for this or should I?

test/models/CommonModel.spec.ts
jonaslagoni: Discussion in https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632269074

src/models/CommonModel.ts
jonaslagoni: Agree, IMO the parameters already explain it well enough, gonna remove it.

test/newsimplification/Simplifier.spec.ts
jonaslagoni: Good point, me and @magicmatatjahu have also discussed it, the mocking framework of jest is also not the greatest IMO..

jonaslagoni: But it works for now, so if you have a strong opinion feel free to create an issue 😄

src/models/CommonModel.ts
jonaslagoni: It is because of `Generic Object Injection Sink` warning in eslint. If the library is used in a JS setting any value can technically be passed to the function, meaning that the transpiled JS version of the function is 
```
addProperty(propertyName, propertyModel, schema) {
```
which does not check that the input is a string format, and any input from the user is accepted.

test/newsimplification/SimplifyProperties.spec.ts
jonaslagoni: True, it should be consistent as well, wanna create an issue, or should I? That we should apply https://eslint.org/docs/rules/eol-last

src/models/CommonModel.ts
jonaslagoni: The reason for choosing `warn` is that we are merging something that potentially can cause you to not get the output you expect. 

For example, say you have the following schema:
```
{
  "anyOf": [
    {
      "properties": {
        "test": {
          "$id": "test",
          ...
        }
      }
    },
    {
      "properties": {
        "test": {
          "$id": "test2",
          ...
        }
      }
    }
  ]
}
```

With the current merge functionality it results in the property `test` be a model with id `"test2"` and you might expect `"test"`. This is used consistently when something happens that you as a user might not expect.

Also in terms of disabling the logging, by default nothing is logged to the console, you have to set your own logging behavior. That reminds me that we need a doc for the logging behavior. Gonna add that.

src/models/CommonModel.ts
jonaslagoni: see https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632957284

src/models/CommonModel.ts
jonaslagoni: see https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632957284

src/models/CommonModel.ts
czlowiek488: @jonaslagoni 
Well it is not only preferences, you will produce more codebase which you will have maintain. 
Let me try to make it more clear with more code examples.
```ts
const ensureArray = (arrayOrOne: any[] | any) => {
   if(arrayOfStringsOrSingleString instanceof Array) {
      return arrayOfStringsOrSingleString;
   }
   return [arrayOfStringsOrSingleString];
};
```
and in tests
```ts
it("test for one type", () => {
   const payload = {};
   
   const result = ensureArray(payload);
   
   expect(result).toBeEqual([payload])
});
it("test for multiple types", () => {
   const payload = [{}, {}];
   
   const result = ensureArray(payload);
   
   expect(result).toBeEqual(payload);
});
```


```ts
addToTypes(ensureArray("payload1"));
// or
addToTypes(esnureArray(["payload#1","payload#2]));
```
But you can easly get rid of ensureArray function and tests if you just do.
```ts
addToTypes(["payload1"]);
// or
addToTypes(["payload#1","payload#2]);
```
Which is way cleaner and more maintainable way because you do not produce code you need to test and making clear that `addToTypes` function has only one way of working.
Reducing those `if`s inside `addToTypes` function will also reduce complexity in bot statistic (I am pretty sure I saw some bot with complexity reports here).

src/models/CommonModel.ts
czlowiek488: Well I do not belive it is a good way to make this.
It would be better to make a validation for user input with Joi or something similiar and throw an error to tell a user that something is wrong. This is a standard to make validation for user input for every API. 
Now if you use propertyName somewhere else you MUST duplicate parsing to string.
Btw. I would prefer to see something like `String(propertyName)`.

test/newsimplification/SimplifyProperties.spec.ts
czlowiek488: Do you by yourself, I have like 3 feature requsts to write for you 😄

jonaslagoni: Awesome 😄 

I will create it.

src/models/CommonModel.ts
czlowiek488: In my case I use allOf for merging those properties and it is something what I actually want.
But if you are saying that `warn` is better, let it be.

jonaslagoni: I would say it is better to provide the user with a clear way (warn) of seeing what might do that you see an output you do not agree with.

This can of course be changed later if we figure out it is just in the way 🙂

src/models/CommonModel.ts
jonaslagoni: One thing we need to be aware of is that this causes the array to be iterated twice when there is no need for it. In our case it does not affect it that much, however, it all depends on how large your dataset is... [Might be time for an eslint rule for loops](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md#chaining)?

src/models/CommonModel.ts
jonaslagoni: For now, let's keep it, but gonna keep it in mind.

test/newsimplification/SimplifyProperties.spec.ts
jonaslagoni: https://github.com/asyncapi/generator-model-sdk/issues/190

src/models/CommonModel.ts
jonaslagoni: Using Joi might be a bit overkill for this 🤔 We use `${}` pretty consistent, so let's stay that way with this. Not that I have any strong opinion about it so it is all about preference, but let's stay consistent.

But we should only parse it once, gonna adapt it.

jonaslagoni: Never mind, it is not that simple apparently, need to think about how to handle security/detect-object-injection another way. 

related: https://stackoverflow.com/questions/44882542/why-is-it-bad-pratice-calling-an-array-index-with-a-variable/44882765#44882765

src/models/CommonModel.ts
czlowiek488: @jonaslagoni 
ok, reduce would be better
```ts
return propertyModels
     .reduce((acc, propertyModel) => 
        propertyModel.$ref !== undefined
            ? [...acc, propertyModel.$ref]
            : acc
     ,[]);
```

czlowiek488: Also be aware that a lot of eslint rules was written for earlier versions of node.js.

As example in node 10 they changed how promise trace is logged, so using `return await Promise.resolve()` is better than `return Promise.resolve()`. Previously trace was the same but returning without await is a microoptimalization, now trace is showed up to the point where you use `await` not farther.

jonaslagoni: I also like this, but if I have to look at it with criticism and a new pair of eyes.

As soon as we start to introduce chaining, we simply reduce the clarity of the function. Newer developers might not directly know `reduce` increasing the complexity.

And since it is open-source we should as much as possible, have as low complexity as possible. I have a hard time finding the balance my self also why we inconsistently use different for loops.

```ts
Object.values(this.properties).forEach((propertyModel) => { 
  if (propertyModel.$ref !== undefined) {
      dependsOn.push(propertyModel.$ref);
  }
});
```
right of the bat everyone can probably guess what `Object.values` does, and so with `forEach`.

With filter and map chaining it becomes somewhat clearer.
```ts
const referencedProperties = Object.values(this.properties)
  .filter(propertyModel => propertyModel.$ref !== undefined)
  .map(propertyModel => `${propertyModel.$ref}`);
dependsOn.push(...referencedProperties);
```

With reduce it becomes somewhat more complex to understand what is happening: 
```ts
propertyModels
  .reduce((acc, propertyModel) => 
    propertyModel.$ref !== undefined
      ? [...acc, propertyModel.$ref]
      : acc
  ,[]);
```

I suggest we should stick with `filter` and `map` for now and simply accept the speed penalty. What do you think @czlowiek488 ? 

Gonna create an issue so we can agree on a consistent way of iterating arrays in the library, also in terms of `forEach`, `for of` etc.

jonaslagoni: https://github.com/asyncapi/generator-model-sdk/issues/192

src/models/CommonModel.ts
jonaslagoni: I think it is time yea 😄 Gonna add an issue for it.

jonaslagoni: https://github.com/asyncapi/generator-model-sdk/issues/193

test/models/CommonModel.spec.ts
jonaslagoni: Assertion added.

src/newsimplification/Simplifier.ts
jonaslagoni: Hehe 🙂 

I think this should be changed depending on the outcome of https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632269074

Since we are using `| undefined` consistently I think we should keep it.

src/newsimplification/Simplifier.ts
jonaslagoni: I agree that the `simplifier` function is a bit much... I was thinking it might be time to add a pre-processor where it iterates the simplified CommonModels and then split them appropriately. 

What if we restructure the simplifier to simply be an input processor for JSON Schema draft 7 inputs (since this is really what it is. This way adding a pre-processor also makes sense, and we can reduce the complexity behind the `simplify` function to only return a single `CommonModel` instead of multiple.

Thoughts @czlowiek488 ?

src/models/CommonModel.ts
jonaslagoni: added an issue for this: https://github.com/asyncapi/generator-model-sdk/issues/192

src/models/CommonModel.ts
czlowiek488: > As soon as we start to introduce chaining, we simply reduce the clarity of the function.
I do not think that chaining reduce clarity, I can agree that it reduce maintability.
> Newer developers might not directly know reduce increasing the complexity.
I totally disagree, if someone do not know the language it is his problem. Also this is in every JS documentation I know about so there is no reason for him to read docs and get this knowledge. So if developer do not know what reduce do, it was his decision for not reading specs. We should use all JS newest features to make possibly fastest and cleanest code.
> And since it is open-source we should as much as possible, have as low complexity as possible. I have a hard time finding the balance my self also why we inconsistently use different for loops.
When I would see filter and map (or something else) instead of reduce I have a question in my head "For what reasons did they not use reduce?" first thoughs is "Because reduce does not allow to do it" but... reduce can do literally everything with array. So I would investigate why it was not used and then if I would found that it could be used... it would give me a headache and make me swear the developer who wrote this.

czlowiek488: Another reason to use reduce is that you need to make more units to test if you would like.
So as example.
```ts
// 2 units for testing
export const filterProperties = propertyModel => propertyModel.$ref !== undefined;
export const mapProperties = propertyModel =>propertyModel.$ref;
Object.values(this.properties)
  .filter(filterProperties)
  .map(mapProperties)
```
vs
```ts
// 1 unit for testing
export const reducePropertes = (acc, propertyModel) => 
    propertyModel.$ref !== undefined
      ? [...acc, propertyModel.$ref]
      : acc;
Object.values(this.properties)
  .reduce(reduceProperties,[]);
```
Sorry for variable names, I do not want to think about it in example.

src/newsimplification/Simplifier.ts
czlowiek488: Please do not use `| undefined` consistently. There is a slight difference between `?` and `| undefined`.
In objects `?` mean that key may not exists but `| undefined` mean that key must exists but its value can be undefined.

czlowiek488: Also you reduce codebase by 10 chars each time you replace `?` with `| undefined`.

src/newsimplification/Simplifier.ts
czlowiek488: @jonaslagoni Could you show an example? I am not sure that I get what you want to do.

src/models/CommonModel.ts
czlowiek488: @jonaslagoni 
> We use ${} pretty consistent

This is only for injecting string into other like ```
`Creating ${entites.length} entities`
``` instead of `'Creating ' + entites.length + ' entites'`. It was not intendent to do weak parsing. 
Also if you may have something else than string why did you tell compiler that there will always be a string?
I think that you should ensure that this string is actually a string or fix wrong type.

> Using Joi might be a bit overkill for this
Why?

File: test/newsimplification/Simplifier.spec.ts (hunk 20 above)
Comment (jonaslagoni) [reply test/newsimplification/Simplifier.spec.ts:45]: Utility function to realize the name of the schema (`src/newsimplification/Utils`).

File: test/newsimplification/Utils.spec.ts (hunk 21 above)
Comment (jonaslagoni) [reply test/newsimplification/Utils.spec.ts:33]: Actually, I agree, gonna refactor it.

File: src/newsimplification/Simplifier.ts (hunk 16 above)
Comment (jonaslagoni) [reply src/newsimplification/Simplifier.ts:23]: I do not like that the simplifier returns `0 ... n` models it should just return a simplified CommonModel for the schema and nothing more (this would reduce a lot of the complexity behind the simplifier as well). By removing the `ensureModelsAreSplit` function and `iteratedModels` property to a pre-processor we can do it after the schema has been simplified.

It of course means that we have to iterate the model twice, but I think it would be worth it in terms of maintainability.

Gonna add an issue clarifying this.

Comment (jonaslagoni) [reply src/newsimplification/Simplifier.ts:23]: https://github.com/asyncapi/generator-model-sdk/issues/195

File: src/newsimplification/Simplifier.ts (hunk 14 above)
Comment (jonaslagoni) [reply src/newsimplification/Simplifier.ts:19]: You are right, but we will tackle it then if that happens, thanks for the headsup!

File: src/newsimplification/Simplifier.ts (hunk 15 above)
Comment (jonaslagoni) [reply src/newsimplification/Simplifier.ts:94]: `?` would also force us to refactor the argument of the function and the way these functions are always called with potential `undefined` values, where the key will always exist. But I agree that in other cases it makes sense to use `?` as we are doing for properties.

Feel free to create an issue if you feel strongly about it and suggest we should change it 🙂

File: src/models/CommonModel.ts (hunk 1 above)
Comment (jonaslagoni) [reply src/models/CommonModel.ts]: Refactored it 😄

File: src/models/CommonModel.ts (hunk 5 above)
Comment (jonaslagoni) [reply src/models/CommonModel.ts:157]: https://github.com/asyncapi/generator-model-sdk/issues/191

File: src/newsimplification/Simplifier.ts (hunk 14 above)
Comment (magicmatatjahu) [reply src/newsimplification/Simplifier.ts:19]: In future we can use deep-merge or write simple function for it, but as you wrote, currently it's enough. If someone add some nested option, then should also update this line.

File: src/models/CommonModel.ts (hunk 9 above)
Comment (jonaslagoni) [reply src/models/CommonModel.ts:298]: @czlowiek488 can you clarify what you mean here? 🤔

File: src/newsimplification/Utils.ts [hunk 33]
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function isModelObject(model: CommonModel) : boolean {
+  // This check should be done instead, needs a refactor to allow it though:
+  // this.extend !== undefined || this.properties !== undefined
+  if (model.type !== undefined) {
+    // If all possible JSON types are defined, don't split it even if it does contain object.
+    if (Array.isArray(model.type) && model.type.length === 7) {
+      return false;
+    }
+    return model.type.includes('object');
```
magicmatatjahu, line 14: I think that here we should also check if `model.type` is array. `includes` function is for Array.

src/newsimplification/Utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  return typeOfEnum;
+}
+
+/**
+ * Find the name for simplified version of schema
+ * 
+ * @param schema to find the name
+ */
+export function simplifyName(schema: any | boolean): string | undefined {
+  if (typeof schema === 'object') {
```
magicmatatjahu, line 44: The `null` has also `object` typeof. 

```suggestion
  if (schema && typeof schema === 'object') {
```

src/newsimplification/SimplifyProperties.ts
magicmatatjahu: Invert condition in `if` :)

src/newsimplification/Simplifier.ts
magicmatatjahu: @czlowiek488 But schema is in first argument, you cannot use optional statement here. Also schema should be used as first argument to function so `| undefined` is ok.

src/newsimplification/SimplifyProperties.ts
jonaslagoni: Done 👍

magicmatatjahu (commented):
Good work! I left some comments :)

src/newsimplification/SimplifyProperties.ts
jonaslagoni: I think it simplifies it using `addProperty`.

src/newsimplification/Utils.ts
jonaslagoni: Good catch, added test.

src/models/CommonModel.ts
jonaslagoni: Made them consistent, thanks 👍

src/models/CommonModel.ts
czlowiek488:
```
if (mergeTo === undefined) {
    return mergeFrom;
}
```

jonaslagoni: Ahh, gonna keep it as is for now, added an issue to add eslint rule about this - https://github.com/asyncapi/generator-model-sdk/issues/190

src/models/CommonModel.ts
jonaslagoni: I don't see the issue as it is a recursive call? 🤔 We have the unit tests in place to ensure that mutation beyond what a function should do does not change.

src/models/CommonModel.ts
jonaslagoni: I am not using the last result, I ensure that all tuple schemas in `items` are merged into one. At least until we start to support tuples: https://github.com/asyncapi/generator-model-sdk/issues/67

src/models/CommonModel.ts
jonaslagoni: > Why?

Because the only reason we have this problem is the `Generic Object Injection Sink`. We don't require complex formats that would require such a library, at least I don't see the benefits, at least not now. If the functions start to be regularly used by external users then it could be a good idea to add it, but not with the current state of the library.

Feel free to create a feature request with the suggested changes, that outlines the benefits of having such a library 🙂

> It was not intendent to do weak parsing.

> In parser we use `String(propertyName)` notation in every such a place :)

Fair enough, adding an issue to refactor it across all files as we use it multiple places.

jonaslagoni: https://github.com/asyncapi/generator-model-sdk/issues/196

docs/simplification of JSON Schema draft 7.md
```diff
@@ -0,0 +1,29 @@
+# Simplifying JSON Schema draft 7 to CommonModel
```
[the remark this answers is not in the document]
jonaslagoni: Done!

docs/simplification of JSON Schema draft 7.md
```diff
@@ -0,0 +1,29 @@
...
+
+The order of simplification:
+- [type](#determining-the-type-for-the-model)
+- `required` are determined as is.
+- `properties` are determined as is, where duplicate properties for the model are merged.
+- [oneOf/anyOf/then/else](#Processing-sub-schemas)
+
+## Determining the type for the model
+To determine the types for the model we use the following interpretation (and in that order):
+- `true` schema infers all model types (`object`, `string`, `number`, `array`, `boolean`, `null`, `integer`).
```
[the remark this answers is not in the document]
jonaslagoni: yea you are right, wanna create an issue for this?

src/newsimplification/Utils.ts
jonaslagoni: It is also for [string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), I think we are fine here.

src/models/CommonModel.ts
magicmatatjahu: In TS you can write overloading functions but at the end you must write only one function like currently.

```ts
addToTypes(types: string);
addToTypes(types: string[]);
addToTypes(types: string[] | string) {
 // ...logic
}
``` 

I can agree with @czlowiek488 in some points, but this function with this shape is fine for me. We can reduce complexity moving normal string to array of strings

```ts
addToTypes(types: string[] | string) {
  types = Array.isArray(types) ? types : [types];
  types = types.filter(Boolean);
  let result = types;
  if (types.length === 0) result = undefined;
  if (types.length === 1) result = types[0];
  return result;
}
```

src/models/CommonModel.ts
jonaslagoni: > I totally disagree, if someone do not know the language it is his problem. Also this is in every JS documentation I know about so there is no reason for him to read docs and get this knowledge. So if developer do not know what reduce do, it was his decision for not reading specs. We should use all JS newest features to make possibly fastest and cleanest code.

Hehe, yea I can easily agree that in most cases you would want to use the newest features, but keep in mind that we want to make time to contribution as low as possible. If this means not using a newer feature for it to be more maintainable so be it, but it all depends. It can vary when it makes sense to prioritize what.

> Another reason to use reduce is that you need to make more units to test if you would not.
I disagree here, yes in your scenario you have two units you are testing, but for this function, you have `getImmediateDependencies`. Nothing more nothing less, and the tests you are creating should not directly care whether you use `filter+map` or `reduce`, what you are testing is the same, especially in terms of coverage.

Also, we cannot use `reduce` in TS it restricts the use case and the return type to `CommonModel` not an array as you use it here. While it is doable in pure js, in TS it is not 🙂 

So lets stay with `filter+map`.

src/models/CommonModel.ts
jonaslagoni: Related to #190

docs/simplification of JSON Schema draft 7.md
```diff
@@ -0,0 +1,29 @@
+# Simplifying JSON Schema draft 7 to CommonModel
+
+The library transforms JSON Schema from data validation rules to data definitions (`CommonModel`(s)). 
+
+The algorithm tries to get to a model whose data can be validated against the JSON schema document. 
+
+We only provide the underlying structure of the schema file for the model formats such as `maxItems`, `uniqueItems`, `multipleOf`, etc, are not simplified.
+
+## Simplifier
```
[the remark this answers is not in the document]
jonaslagoni: @magicmatatjahu are you suggesting we name the process `Interpretation` instead? 🙂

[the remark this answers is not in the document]
jonaslagoni: @magicmatatjahu are you suggesting we name the process `Interpretation` instead? 🙂

src/models/CommonModel.ts
jonaslagoni: Done 👍

src/models/CommonModel.ts
jonaslagoni: I do not understand what you mean here, can you clarify it a bit more @czlowiek488?

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
jonaslagoni: @magicmatatjahu are you suggesting we rename `simplification` to `Interpretation`? 🙂

jonaslagoni:
@all-contributors please add @czlowiek488 for review, ideas

src/models/CommonModel.ts
czlowiek488: @jonaslagoni  You can type reduce easly.

src/models/CommonModel.ts
czlowiek488: I think that this should be changed. this mergeCommonModels should not hide mutation of mergedItemsModel.

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
czlowiek488: Just make sure that this is compatible with asyncapi, because if its not people may not be able to use this library.

src/models/CommonModel.ts
jonaslagoni: > @jonaslagoni You can type reduce easly.

We might be able to do so, but let's not push the use-case of the function, filter+map will do the job nicely in my opinion.

src/models/CommonModel.ts
czlowiek488: Also I will never agree to let it be like here. From my point of view you violate SRP.

src/models/CommonModel.ts
jonaslagoni: Agreed, should be changed as of that issue. But since this is a refactoring of the simplifier, this functionality is already in master, so let's not modify it here.

src/models/CommonModel.ts
czlowiek488: I guess it violate even YAGNI rule because you have multiple ways to achieve the same effect.

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
jonaslagoni: Of course. We are referring to the simplified `CommonModel` which should allow having the type `any` instead of inferring all types.

As we move further and further away from JSON Schema, and `CommonModel` have less in common with it, I think we should consider separating them completely, think it makes more sence. But that is something for later.

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
magicmatatjahu: exactly :)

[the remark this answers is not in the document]
jonaslagoni: @czlowiek488 what do you think about this? Does it make sense to rename the entire process to `interpretation`?

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
magicmatatjahu: Created :) https://github.com/asyncapi/generator-model-sdk/issues/198

src/models/CommonModel.ts
jonaslagoni: @czlowiek488 no one disputes that in other languages you would have split these functions up and overloaded them to provide a more maintainable code. 

However, with the nature of what the function does, how it is used, the underlying meaning of the `type` property, together with the limitations of TS this implementation looks fine for me as well. Especially since switching to only allowing array would make it more complex to understand IMO, and it would a discussion about potato/potato (given the nature of TS).

Always appreciate the feedback and feel free to prove me wrong with a PR/issue which clearly shows that maintainability increases and reduces complexity 🙂

docs/simplification of JSON Schema draft 7.md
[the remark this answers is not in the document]
czlowiek488: @jonaslagoni Yes it is good new name, simplifier doesnt mean anything but interpretator does.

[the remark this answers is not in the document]
jonaslagoni: Alright, gonna start renaming it 🙂

.github/workflows/coverall.yml
```diff
@@ -20,6 +20,6 @@ jobs:
         npm install
         npm run test
     - name: Coveralls
-      uses: coverallsapp/github-action@v1.1.2
+      uses: coverallsapp/github-action@master
```
magicmatatjahu, line 23: If you use `v1` then you will have latest version from `v1`.

src/interpreter/Interpreter.ts
```diff
@@ -0,0 +1,143 @@
...
+      //All schemas of type object MUST have ids
+      if (model.type !== undefined && model.type.includes('object')) {
+        model.$id = interpretName(schema) || `anonymSchema${this.anonymCounter++}`;
+      } else if (schema.$id !== undefined) {
+        model.$id = interpretName(schema);
+      }
+
+      if (schema.required !== undefined) {
+        model.required = schema.required;
+      }
```
magicmatatjahu, line 80:
```suggestion
        model.required = schema.required || model.required;
```

src/models/CommonModel.ts
```diff
@@ -107,11 +123,12 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
-      Object.entries(this.properties).forEach(([, propertyModel]) => {
-        if (propertyModel.$ref !== undefined) {
-          dependsOn.push(propertyModel.$ref);
-        }
-      });
+      const refProperties = (propertyModel: CommonModel) => propertyModel.$ref !== undefined;
+      const refProperty = (propertyModel: CommonModel) => `${propertyModel.$ref}`;
+      const referencedProperties = Object.values(this.properties)
+        .filter(refProperties)
+        .map(refProperty);
```
magicmatatjahu, line 130: Really, please add these function inside `fliter` and `map` 😄

magicmatatjahu (commented):
Only 3 additional comments

.github/workflows/coverall.yml
jonaslagoni: I know, but github runners are using a cached version of `@v1.1.2` which contains an underlying dependency cache that has not been updated (they broke a version). So if I use `1.1.2` at times the run will fail.

related issues https://github.com/nickmerwin/node-coveralls/issues/280 and https://github.com/coverallsapp/github-action/issues/76

jonaslagoni: Which was why https://github.com/asyncapi/generator-model-sdk/runs/2618803562 failed

magicmatatjahu: Ok

magicmatatjahu (approved):

jonaslagoni:
Thanks for the reviews and thoughts @magicmatatjahu  @czlowiek488 🎉

src/models/CommonModel.ts
czlowiek488: @jonaslagoni  @magicmatatjahu  It is not bad at all. It is actually good practice but you need to move them at the top of file and export them. In this way you could unit test those functions, it is also tiny step into functional programming.

jonaslagoni:
@czlowiek488 do you want me to request more reviews from you? Otherwise, it goes without saying you can always jump into a review of PRs whenever you have time :)

czlowiek488:
Just ping me if you wish a review, sometimes I can miss other notifications.

Issue #148, opened 2021-04-22
Add a validator infront of the input processor


jonaslagoni:
#### Describe the bug
We are expecting the input JSON Schema draft 7 to be a document that can validate correctly, but we do nothing when this is not the case.

Example: If a schema uses combination schemas that each contain the same property name, it will try to merge the two common models. However, if it is from an `allOf` combination schema, it should "fail" somehow if conflicts occur. Since the input JSON Schema could never validate any data we should consider it an invalid input.

#### How to Reproduce
Given the JSON Schema:
```json
{
  "allOf": [
    {
      "type": "object",
      "properties": {
        "somePropName": {
          "type": "number"
        }
      }
    },
    {
      "type": "object",
      "properties": {
        "somePropName": {
          "enum": [
            "residential",
            "business"
          ]
        }
      },
      "required": [
        "type"
      ]
    }
  ]
}
```
It would create the simplified model:
```json
{
  "type": "object",
  "properties": {
    "somePropName": {
      "type": [
        "number",
        "string"
      ]
    }
  }
}
```
Which is incorrect. However, it is the input that is incorrect, since the JSON Schema can never validate the input.

#### Expected behavior
I would expect the model generator to fail the generation, or handle invalid inputs somehow.

jonaslagoni:
While this would be great to have, there exist no tools for this.

In the simplest form we should be able to detect the following schemas:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "not": {
    "type": "object"
  }
}
```

A possible approach suggested by @jdesrosiers on the JSON Schema slack:

> I could imagine a tool that automatically refactors a schema to clean up common problems such as remove duplicated/unnecessary keywords or merge allOfs where possible. Such a tool might reduce a schema to false. It wouldn't be perfect, but it could at least detect some cases.

Closing this issue for now, as we have no way to enable this.

PR #200, opened 2021-05-18
refactor: const interpretation

jonaslagoni:
**Description**
This PR adds the refactoring of the interpretation of `const`.

**It adds**
- InterpretConst, to interpret the `const` schema keyword.

- partly https://github.com/asyncapi/generator-model-sdk/blob/master/src/simplification/SimplifyEnums.ts
- And all related tests have been refactored to have next to no dependencies by using the jest mocking framework.

**Todo**

- [x] Refactor docs
- [x] Solve code smells
- [x] Inspect tests
- [x] Rename to transformer

**Related issue(s)**
Blocked by #184


src/interpreter/InterpretConst.ts
```diff
@@ -0,0 +1,25 @@
...
+  const schemaConst = schema.const;
+  model.enum = [schemaConst];
+
+  //If schema does not contain type interpret the schema
+  if (schema.type === undefined) {
+    const inferredType = inferTypeFromValue(schemaConst);
+    if (inferredType !== undefined) {
+      model.setType(inferredType);
+    }
+  }
```
magicmatatjahu, line 24: Hmm... If someone pass `const` then it should (probably) override types. Am I wrong?

test/interpreter/InterpretConst.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const schema: any = { const: 'test'};
+    const model = new CommonModel();
+    interpretConst(schema, model);
+    expect(model.enum).toEqual([schema.const]);
+    expect(inferTypeFromValue).toHaveBeenNthCalledWith(1, 'test');
+    expect(model.type).toEqual("string");
+  });
+  test('should not infer unknown type', function() {
+    (inferTypeFromValue as jest.Mock).mockReturnValue(undefined);
+    const schema: any = { const: 'test'};
```
magicmatatjahu, line 45: You can use object `const: {}`

src/interpreter/InterpretConst.ts
jonaslagoni: With this approach, we favor already sat `type` keyword intstead of interpreting `const` as type. Which is why` if (schema.type === undefined) {` is there. This is just to keep it as we have it now with the old simplifier.

jonaslagoni: We can change this behavior based on the interpretation options later, if we want to.

magicmatatjahu: Yep, but if you have `{ type: 'boolean', const: "foo" }` then you have invalid schema... however I agree to go with what is in the scheme.

test/interpreter/InterpretConst.spec.ts
magicmatatjahu: Ok sorry 😅 Then you will have `object` type... So checking undefined is necessary?

src/interpreter/InterpretConst.ts
jonaslagoni: yep it would, but neither `type: 'boolean'` or `type: 'string'` is correct in this case 😄 

This is also one of the reasons why I suggested #148.

magicmatatjahu: Probably `const` doesn't combine with other types and only enforces one type. Maybe I don't understand JSON Schema correctly.

test/interpreter/InterpretConst.spec.ts
jonaslagoni: I can change it to `{}` if you think that makes it more clear, because of `(inferTypeFromValue as jest.Mock).mockReturnValue(undefined);` it does not really matter what the const is 🙂

magicmatatjahu: Yep, but even if you change to `object` you will infer `object` type. Probably `undefined` will never happen.

src/interpreter/InterpretConst.ts
jonaslagoni: If you try and validate `{ type: 'boolean', const: "foo" }` with some input data, you won't be able to because the `type` and `const` are contradicting. 

i.e. if your input is `"foo"` it is accepted by `const: "foo"` but is rejected by `type: 'boolean'`, and vice versa if the input is a boolean 😄 

So we cant really win if you create invalid schemas 😄

test/interpreter/InterpretConst.spec.ts
jonaslagoni: In practice, not really no. In theory, it could

```js
const jsonSchema = {
    type: "string",
    const: () => {}
}
const generator = new TypeScriptGenerator({ modelType: 'interface' });
const interfaceModel = await generator.generate(jsonSchema);
```

jonaslagoni: The test is purely there to ensure that the branch is covered 🙂

magicmatatjahu (approved):

jonaslagoni:
Thanks 👍

PR #203, opened 2021-05-18
refactor: not interpretation

jonaslagoni:
**Description**
This PR adds the refactoring of the interpretation for `not`.

**It adds**
- interpretNot, to interpret the `not` schema keyword.

This does not directly replace a single file, however, it fixes a few bugs with the current simplifier.

**Todo**

- [x] Refactor docs
- [x] Solve code smells
- [x] Inspect tests
- [x] Rename to interpreter

**Related issue(s)**
Blocked by #184


src/models/CommonModel.ts
```diff
@@ -36,15 +36,39 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+      for (const type of typesToRemove) {
+        this.removeType(type);
+      }
+    } else if (this.type !== undefined && this.type.includes(typesToRemove)) {
+      if (Array.isArray(this.type)) {
+        this.setType(this.type.filter((el) => {
+          return el !== typesToRemove;
+        }));
+      } else {
+        this.setType(undefined);
```
magicmatatjahu, line 70: You can here remove reassign types directly like `this.type = this.type.filter(el => el !== typesToRemove)` and `this.type = undefined` only for easier reading the code.

src/interpreter/InterpretNot.ts
```diff
@@ -0,0 +1,32 @@
...
+    const newInterpreterOptions: InterpreterOptions = {
+      ...interpreterOptions, 
+      splitModels: false,
+      allowInheritance: false
+    };
+    const interpretedModels = interpreter.interpret(notSchema, newInterpreterOptions);
+    if (interpretedModels.length > 0) {
+      const interpretedModel = interpretedModels[0];
+      if (interpretedModel.type !== undefined) model.removeType(interpretedModel.type);
+      model.removeEnum(interpretedModel.enum);
```
magicmatatjahu, line 27:
```suggestion
      if (interpretedModel.enum !== undefined) model.removeEnum(interpretedModel.enum);
```

For easier reading the code :) I know that you check statement inside the function, but let's be consistent here (with `interpretedModel.type`) :)

src/interpreter/InterpretNot.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+  } else if (schema.not === true) {
+    Logger.warn(`Encountered true not schema. Which rejects everything for ${model.$id}. This schema are not applied!`, schema);
```
magicmatatjahu, line 30: This should be also documented in docs :)

magicmatatjahu (commented):
Great job! Only few comments :) My first thought about passing directly to each function `interpreterOptions` was `wtf` 😄 but then then I realized for what and, of course, it's redundant but it's needed :)

src/interpreter/InterpretNot.ts
jonaslagoni: True, changed it 🙂

src/interpreter/InterpretNot.ts
jonaslagoni: Good call, changed 😄

src/models/CommonModel.ts
jonaslagoni: `type` is just interpreted a little differently because it can be an array of a single value which is why I use the result of the filter and call `setType` so it handles that logic 😄

src/interpreter/InterpretNot.ts
jonaslagoni: Agree 👍

magicmatatjahu (approved):
LGTM!

PR #204, opened 2021-05-18
refactor: additionalProperties interpretation

jonaslagoni:
**Description**
This PR adds the refactoring of the interpretation for `additionalProperties`.

**It adds**
- InterpretAdditionalProperties, to interpret the `additionalProperties` schema keyword.

- https://github.com/asyncapi/generator-model-sdk/blob/master/src/simplification/SimplifyAdditionalProperties.ts
- And all related tests have been refactored to have next to no dependencies by using the jest mocking framework.

**Todo**

- [x] Refactor docs
- [x] Solve code smells
- [x] Inspect tests
- [x] Rename to transformer

**Related issue(s)**
Blocked by #184


src/interpreter/InterpretAdditionalProperties.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Interpreter function for JSON Schema draft 7 additionalProperties keyword.
+ * 
+ * @param schema
+ * @param model
+ * @param interpreter
+ */
+export default function interpretAdditionalProperties(schema: Schema | boolean, model: CommonModel, interpreter : Interpreter) {
+  if (typeof schema === 'boolean' || !isModelObject(model)) return;
+  const newModels = interpreter.interpret(schema.additionalProperties || true);
```
magicmatatjahu, line 15: If you have `true` then you never go to `15` line so

```ts
const newModels = interpreter.interpret(schema.additionalProperties);
```

is enough.

jonaslagoni: There is a difference between the schema being boolean and the additionalProperty schema being a boolean.

Not sure what you are suggesting here? 🤔

jonaslagoni: waiiit, `schema` should not have type `boolean | schema`... 

Gonna change it

magicmatatjahu (approved):

jonaslagoni:
Thanks!

PR #211, opened 2021-05-22
refactor: proper string parsing

### LINKED ISSUE #196: Use proper string parsing
#### Reason/Context
We have a security eslint rule for `Generic Object Injection Sink`, that forces us to cast a variable to string when used as a key. At the moment we in many places use string injections such as `${variable}` where we should use `String(variable)`.

Based on the following discussion: https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632278908

#### Description
Example location that should change: https://github.com/asyncapi/generator-model-sdk/blob/10250e8338ddbcf8222f68e0bed0c6512c863a99/src/models/CommonModel.ts#L194

arjungarg07:
**Description**

Replaced string injections such as ${variable} to String(variable).

**Related issue(s)**

Resolves #196


src/models/CommonModel.ts
```diff
@@ -29,7 +29,7 @@ export class CommonModel extends CommonSchema<CommonModel> {
   getFromSchema<K extends keyof Schema>(key: K) {
     const schema = this.originalSchema || {};
     if (typeof schema === 'boolean') return undefined;
-    return schema[`${key}`];
+    return schema[String(key)];
```
fmvilas, line 32: Why not simply `schema[key]`?

arjungarg07: It is because of `Generic Object Injection` Sink warning in eslint.

Based on the following discussion: [#184](https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632278908)

fmvilas: Got it. Thanks!

fmvilas (approved):
LGTM

src/models/CommonModel.ts
```diff
@@ -281,9 +281,9 @@ export class CommonModel extends CommonSchema<CommonModel> {
         for (const [pattern, patternModel] of Object.entries(mergeFromPatternProperties)) {
           if (mergeToPatternProperties[`${pattern}`] !== undefined) {
```
jonaslagoni, line 282:
```suggestion
          if (mergeToPatternProperties[String(pattern)] !== undefined) {
```

Could potentially introduce a variable here since it is used 2 places, but up to you 🙂

jonaslagoni (changes requested):
Only 1 comment, thanks for the quick PR! 🙇

src/models/CommonModel.ts
arjungarg07: aah! missed it😅

jonaslagoni (approved):
Awesome 👍

derberg:
@jonaslagoni fyi, Arjun cannot merge 😉 unless you were just afraid to merge on weekend 😆

jonaslagoni:
> @jonaslagoni fyi, Arjun cannot merge 😉 unless you were just afraid to merge on weekend 😆

@derberg Let's just say I prefer not to do it on the weekends 😆

jonaslagoni:
@all-contributors please add @arjungarg07 for code

jonaslagoni:
@all-contributors please add @fmvilas for review

PR #214, opened 2021-05-24
fix: array of items not supported by getNearestDependencies

jonaslagoni:
**Description**
This PR fixes an issue with an array of items not supported and then renames the function `getImmediateDependencies` to `getNearestDependencies`.

**Related issue(s)**
fixes https://github.com/asyncapi/generator-model-sdk/issues/122
fixes https://github.com/asyncapi/generator-model-sdk/issues/186

### DIFF
--- a/src/models/CommonModel.ts
+++ b/src/models/CommonModel.ts
@@ -123,11 +123,11 @@ export class CommonModel extends CommonSchema<CommonModel> {
    */
   addProperty(propertyName: string, propertyModel: CommonModel, schema: Schema) {
     if (this.properties === undefined) this.properties = {};
-    if (this.properties[`${propertyName}`] !== undefined) {
+    if (this.properties[String(propertyName)] !== undefined) {
       Logger.warn(`While trying to add property to model, duplicate properties found. Merging models together for property ${propertyName}`, propertyModel, schema, this);
-      this.properties[`${propertyName}`] = CommonModel.mergeCommonModels(this.properties[`${propertyName}`], propertyModel, schema);
+      this.properties[String(propertyName)] = CommonModel.mergeCommonModels(this.properties[String(propertyName)], propertyModel, schema);
     } else {
-      this.properties[`${propertyName}`] = propertyModel;
+      this.properties[String(propertyName)] = propertyModel;
     }
   }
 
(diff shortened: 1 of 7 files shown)


magicmatatjahu (commented):
If we will support `patternProperties` etc we should also support them in `getNearestDependencies` function.

jonaslagoni:
> If we will support `patternProperties` etc we should also support them in `getNearestDependencies` function, so this PR should be fixed/merged after refactor.

Good point. However, I think we can adapt it already since it does not affect the refactor.

src/models/CommonModel.ts
```diff
@@ -188,31 +188,38 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
     if (this.properties !== undefined && Object.keys(this.properties).length) {
       const referencedProperties = Object.values(this.properties)
         .filter((propertyModel: CommonModel) => propertyModel.$ref !== undefined)
         .map((propertyModel: CommonModel) => `${propertyModel.$ref}`);
       dependsOn.push(...referencedProperties);
     }
+    if (this.patternProperties !== undefined && Object.keys(this.patternProperties).length) {
+      const referencedPatternProperties = Object.values(this.patternProperties)
+        .filter((patternPropertyModel: CommonModel) => patternPropertyModel.$ref !== undefined)
+        .map((patternPropertyModel: CommonModel) => `${patternPropertyModel.$ref}`);
```
magicmatatjahu, line 220: String(patternPropertyModel.$ref)

src/models/CommonModel.ts
```diff
@@ -188,31 +188,38 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+        const itemRef = item.$ref;
+        if (itemRef !== undefined) {
+          dependsOn.push(itemRef);
+        }
+      });
     }
     if (this.properties !== undefined && Object.keys(this.properties).length) {
       const referencedProperties = Object.values(this.properties)
         .filter((propertyModel: CommonModel) => propertyModel.$ref !== undefined)
         .map((propertyModel: CommonModel) => `${propertyModel.$ref}`);
```
magicmatatjahu, line 214: String(propertyModel.$ref)

src/models/CommonModel.ts
jonaslagoni: Right 🤦

src/models/CommonModel.ts
jonaslagoni: Right 🤦

Issue #190, opened 2021-05-15
Apply more strict eslint rules


jonaslagoni:
#### Reason/Context
We should apply more strict eslint rules to keep it consistant. 

- [x] [New line ending](https://eslint.org/docs/rules/eol-last), sprung from https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632280767
- [x] [Curly](https://eslint.org/docs/rules/curly), sprung from https://github.com/asyncapi/generator-model-sdk/pull/184#discussion_r632271883
- [ ] Iteration of array - https://github.com/asyncapi/generator-model-sdk/issues/192

PR #215, opened 2021-05-24
refactor: add more strict eslint rules

jonaslagoni:
**Description**
This PR adds a more strict eslint configuration to ensure we have consistency for developers to know exactly the syntax which is required. This is extended to test files as well.

**Related issue(s)**
Partly solves #190


magicmatatjahu:
@jonaslagoni I can accept the first two, but last `Force no curlies with for one lines curly` isn't good for me. I'm not a fan with that strict writing. I can accept code like `if (...) something()`, but

```ts
if (...)
  something()
```

isn't acceptable for me.

jonaslagoni:
> @jonaslagoni I can accept the first two, but last `Force no curlies with for one lines curly` isn't good for me. I'm not a fan with that strict writing. I can accept code like `if (...) something()`, but
> 
> ```ts
> if (...)
>   something()
> ```
> 
> isn't acceptable for me.

👍 Lets change it then...

jonaslagoni:
Soo, apperant the style we are looking for is not really [possible](https://stackoverflow.com/questions/65697539/prettier-forces-curlyless-single-line-if-statement-to-be-on-the-same-line-inste), I wonder if we want to force always brackets to enable consistancy 🤔?

magicmatatjahu:
> Soo, apperant the style we are looking for is not really possible, I wonder if we want to force always brackets to enable consistancy 🤔?

We can give free will to contributors in this topic :)

jonaslagoni:
> We can give free will to contributors in this topic :)

When another contributor comes along and uses 

```
if (...)
  something()
```

Would you then accept it? Cause if nothing is explicitly telling the developer that this syntax is not preferred it is IMO up to them to decide.

magicmatatjahu:
@jonaslagoni Yes, I forgot about that 😅 So maybe we should force the brackets, even for inline ifs

jonaslagoni:
> @jonaslagoni Yes, I forgot about that 😅 So maybe we should force the brackets for inline ifs

Yea I agree, lets do it.

magicmatatjahu (approved):
These inline returns look a bit weird 😅 I give accept to not block you anymore, but maybe we shouldn't bother with such rules? If someone will write inline ifs, then we should only aware about that, we'll just pay attention.

jonaslagoni:
> These inline returns look a bit weird 😅 I give accept to not block you anymore, but maybe we shouldn't bother with such rules? If someone will write inline ifs, then we should only aware about that, we'll just pay attention.

I would rather have us use one specific rule and force that, regardless of what it is, to be consistent. However, that is just my opinion, I can see it might be cumbersome when we are used to applying whatever style we see fit. 

In the long run, I think applying this rule is better so we don't mix it everywhere especially as each developer has their own opinion.

Don't mind this being blocked, as long as we agree on a decision in the end 🙂 

cc @czlowiek488

czlowiek488:
It does not matter what is preffered by who but what the world standards are.  And you should not fight with those standards because they are standards for some reason.

You should not give a choice for developer, it should have strict autoformatting to make code consistent. Consistent code makes you read it faster because you know what you can expect in every part of codebase.

I would also recommend you to use prettier together with eslint. It may be a bit tricky to set it up so you may want to take a look at [this project](https://github.com/TheSoftwareHouse/express-boilerplate) and copy what you need. It may be also a source of good practices e.g. usage of dependency injection engine.

.eslintrc
```diff
@@ -25,7 +26,8 @@
     "no-mixed-requires": 0,
     "no-process-exit": 0,
     "no-warning-comments": 0,
-    "curly": 0,
+    "no-loop-func": "error",
+    "curly": ["error"],
     "no-multi-spaces": 0,
```
czlowiek488, line 31: You may want to add those rules.
    
    "require-await": "error",
    "no-only-tests/no-only-tests": "error",
    "no-console": "error",

jonaslagoni: Added all but `no-only-tests/no-only-tests`, since IMO test should apply the same rules but a few.

jonaslagoni:
> It does not matter what is preffered by who but what the world standards are. And you should not fight with those standards because they are standards for some reason.
> 
> You should not give a choice for developer, it should have strict autoformatting to make code consistent. Consistent code makes you read it faster because you know what you can expect in every part of codebase.

I agree let's be strict in the syntax that can be used so the developer cannot be in doubt whether one or the other syntax should be used, and of course, so we are consistent.

I changed eslint to include tests now, and removed any unnecessary rules or changed them to "errors". The generic preset types @magicmatatjahu are going wild and I had to disable certain eslint rules for that file since the ripple effect was insane.

> I would also recommend you to use prettier together with eslint. It may be a bit tricky to set it up so you may want to take a look at [this project](https://github.com/TheSoftwareHouse/express-boilerplate) and copy what you need. It may be also a source of good practices e.g. usage of dependency injection engine.

@czlowiek488  long term that might be a great idea 👍 Feel free to create a feature request for this if you feel like it could be a good idea 🙂

src/generators/javascript/JavaScriptGenerator.ts
```diff
@@ -27,17 +27,17 @@ export class JavaScriptGenerator extends AbstractGenerator<JavaScriptOptions> {
     super('JavaScript', JavaScriptGenerator.defaultOptions, options);
   }
 
-  async render(model: CommonModel, inputModel: CommonInputModel): Promise<string> {
+  render(model: CommonModel, inputModel: CommonInputModel): Promise<string> {
```
czlowiek488, line 30: Shouldnt be here `: string` instead of `: Promise<string>`?

czlowiek488: I see that this is repeated in many more places.

src/generators/java/renderers/EnumRenderer.ts
```diff
@@ -34,7 +34,7 @@ ${this.indent(this.renderBlock(content, 2))}
     return `${content};`;
   }
 
-  normalizeKey(value: any): string {
+  normalizeKey(value: unknown): string {
```
czlowiek488, line 37: I do not think that unknown is a good here. I guess any was better. I think that unknown should be used only if you have API entrypoint and the you have unknown user input which is not used by your code at all.
I guess that you should do `any | number | boolean` and maybe bigint if it is possible.

test/utils/LoggingInterface.spec.ts
```diff
@@ -94,10 +94,10 @@ describe('LoggingInterface', function() {
...
       errorMock.mockClear();
     });
-    test('debug', function() {
-      Logger.debug("test", {}, "test2");
+    test('debug', () => {
+      Logger.debug('test', {}, 'test2');
       expect(debugMock).toHaveBeenCalledTimes(1);
-      expect(debugMock).toHaveBeenCalledWith("test", {}, "test2");
+      expect(debugMock).toHaveBeenCalledWith('test', {}, 'test2');
       expect(infoMock).toHaveBeenCalledTimes(0);
```
czlowiek488, line 101: This is overkill :D

jonaslagoni: It is a simple `--fix` and I gotta admit I like that its constant 😄

src/generators/javascript/JavaScriptGenerator.ts
jonaslagoni: It is because we wanted to allow the async `presets` that we used `Promise` throughout 🙂

src/generators/java/renderers/EnumRenderer.ts
jonaslagoni: as soon as you have `any` what is the reason for adding more types? 🤔

jonaslagoni:
Because of this issue: https://github.com/asyncapi/.github/issues/55 I have to disable eslint rule for newlines.

magicmatatjahu (approved):
LGTM :)

jonaslagoni:
Thanks for the reviews @magicmatatjahu and @czlowiek488 👍

PR #216, opened 2021-05-24
ci: fail linting for forEach loops

### LINKED ISSUE #190: Apply more strict eslint rules

jonaslagoni:
**Description**
This PR adds an eslint error when forEach is used instead of `for .. of`.

Reasons for this change: https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md

blocked by #218 

**Related issue(s)**
fixes #190


jonaslagoni:
Waiting for the refactor of the interpreter to finish before this can be merged.

src/models/CommonModel.ts
```diff
@@ -366,14 +366,15 @@ export class CommonModel extends CommonSchema<CommonModel> {
    * @param originalSchema 
    * @param alreadyIteratedModels
    */
+  // eslint-disable-next-line sonarjs/cognitive-complexity
```
magicmatatjahu, line 369: You probably should use also `// NOSONAR`, because I still see the problem on sonarcloud

jonaslagoni: I am not sure it is possible for a function like this 🤔

magicmatatjahu: Even if you add this comment after function like `functioName(...) { // NOSONAR` ? :)

jonaslagoni: That worked 🍻 Thanks!

magicmatatjahu (approved):
LGTM!

PR #218, opened 2021-05-26
refactor: switch to interpreter

jonaslagoni:
**Description**
This PR is the final one that switches the old simplifier function out with the interpreter.

**Related issue(s)**
fixes https://github.com/asyncapi/generator-model-sdk/issues/64


jonaslagoni:
@czlowiek488 would love to get your feedback on the documentation of the interpreter if you have time 🙂 Can't request a review from you for some reason, so this will have to do 😄

docs/interpretation_of_JSON_Schema_draft_7.md
```diff
@@ -34,21 +35,28 @@ Currently the following `not` model properties are interpreted:
...
-The following JSON Schema keywords are merged with the already interpreted model:
+The following JSON Schema keywords are [merged](#Merging-models) with the already interpreted model:
 - `oneOf`
 - `anyOf`
 - `then`
 - `else`
+
+
+## Merging models
+Because of the recursive nature of the interpreter (and the nested behavior of JSON Schema) it happens that two models needs to be merged together.
```
magicmatatjahu, line 51:
```suggestion
Because of the recursive nature of the interpreter (and the nested nature of JSON Schema) it happens that two models needs to be merged together.
```

docs/interpretation_of_JSON_Schema_draft_7.md
```diff
@@ -34,21 +35,28 @@ Currently the following `not` model properties are interpreted:
...
+
+If only one side has a property defined, it is used as is, if both have it defined they are merged based on the following logic:
+- `additionalProperties` if both models contain it the two are recursively merged together.
```
magicmatatjahu, line 54:
```suggestion
- `additionalProperties` if both models contain items they are recursively merged together.
```

docs/interpretation_of_JSON_Schema_draft_7.md
```diff
@@ -4,28 +4,29 @@ The library transforms JSON Schema from data validation rules to data definition
 
 The algorithm tries to get to a model whose data can be validated against the JSON schema document. 
 
-We only provide the underlying structure of the schema file for the model formats such as `maxItems`, `uniqueItems`, `multipleOf`, etc, are not transformed.
+We only provide the underlying structure of the schema file for the model, where formats such as `maxItems`, `uniqueItems`, `multipleOf`, etc, are not transformed.
```
magicmatatjahu, line 7:
```suggestion
We only provide the underlying structure of the schema file for the model, where constraints/annotations such as `maxItems`, `uniqueItems`, `multipleOf`, etc, are not transformed.
```

docs/interpretation_of_JSON_Schema_draft_7.md
```diff
@@ -34,21 +35,28 @@ Currently the following `not` model properties are interpreted:
...
+If only one side has a property defined, it is used as is, if both have it defined they are merged based on the following logic:
+- `additionalProperties` if both models contain it the two are recursively merged together. 
+- `patternProperties` if both models contain a pattern the corresponding models are recursively merged together. 
+- `properties` if both models contain the same property the corresponding models are recursively merged together. 
+- `items` if both models contain items they are recursively merged together.
+- `types` if both models contain types they are merged together, duplicate types are removed.
+- `enum` if both models contain enums they are merged together, duplicate enums are removed.
+- `required` if both models contain required properties they are merged together, duplicate required properties are removed.
+- `$id`, `$ref`, `extend` uses left model value if present otherwise right model value if present.
+- `originalSchema` are overwritten.
```
magicmatatjahu, line 62: We have information about `originalSchema` only in `API.md` file, we should described what is `originalSchema` here in another place :)

magicmatatjahu (commented):
Some comments :)

docs/interpretation_of_JSON_Schema_draft_7.md
jonaslagoni: Where does `items` fit in? 🤔

magicmatatjahu: Sorry 😅 My fault. I did a review right after waking up. It can stay as is, it makes more sense :)

docs/interpretation_of_JSON_Schema_draft_7.md
jonaslagoni: Hmm, yea true. What do you think is the best solution? 🤔

magicmatatjahu: It can be described here, or somewhere at the beginning of this doc :)

jonaslagoni: Added a link to the input processor docs which explains the CommonModel, otherwise, we have to keep in mind that we have a description of a property in this document as well.

magicmatatjahu (approved):

src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -212,7 +212,8 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
    * @param schema to simplify to common model
    */
   static convertSchemaToCommonModel(schema: Schema | boolean): Record<string, CommonModel> {
-    const commonModels = simplify(schema);
+    const interpreter = new Interpreter();
```
czlowiek488, line 215: I guess interpreter should be constructed before this method execution. Please remember that creating instance by using keyword `new` is not well optimized in javascript and should be avoided at execution time.

test/processors/JsonSchemaInputProcessor.spec.ts
```diff
@@ -2,12 +2,27 @@ import * as fs from 'fs';
...
 import { JsonSchemaInputProcessor } from '../../src/processors/JsonSchemaInputProcessor';
 import { CommonModel, Schema } from '../../src/models';
-import { simplify } from '../../src/simplification/Simplifier';
-jest.mock('../../src/simplification/Simplifier');
+import { Logger } from '../../src/utils';
+jest.mock('../../src/interpreter/Interpreter');
+jest.mock('../../src/utils/LoggingInterface');
 jest.spyOn(JsonSchemaInputProcessor, 'convertSchemaToCommonModel');
+
+let mockedReturnModels = [new CommonModel()];
```
czlowiek488, line 10: I should blame you every single time you use `let` or mutate something in your code ;)

src/interpreter/Interpreter.ts
```diff
@@ -104,6 +105,7 @@ export class Interpreter {
    * 
    * @param schema to go through
    * @param currentModel the current output
+   * @param options to control the interpret process
    */
   combineSchemas(schema: (Schema | boolean) | (Schema | boolean)[] | undefined, currentModel: CommonModel, rootSchema: Schema, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
```
czlowiek488, line 110: For me you violate YAGNI, KISS and SRP. And I will point it up whenever I see someting like that. If you would be my collegue at work I would block your PR.

czlowiek488 (changes requested):
Looks fine but I would change few things. You may be fine with this codebase. However I am not 😞

src/interpreter/Interpreter.ts
jonaslagoni: Hehe, I mean you kinda are, let me deep dive a bit and try and break it up to better understand it

jonaslagoni:
> Looks fine but I would change few things. You may be fine with this codebase. However I am not 😞

Thanks for the review @czlowiek488, gonna go through your comments now 💪

test/processors/JsonSchemaInputProcessor.spec.ts
jonaslagoni: What do you mean 😄 ? 

This is the only way I found that could alter the mocked return, damn Jest..

src/processors/JsonSchemaInputProcessor.ts
jonaslagoni: We could just have one interpreter instance where we could reset the iterated models, but it's an optimization we can make later if it ever becomes a problem.

src/interpreter/Interpreter.ts
jonaslagoni: I agree with you here, it should be split in this case. Changed 👍

jonaslagoni:
So caught 2 problems with the current interpreter.
1. The blackbox testing actually caught an error with how the separate models got saved, had to add an integration test to ensure that models are iterated and returned correctly. Unit testing with mocks would be too complex, so switched to that.
2. I forgot to implement split functionality for `patternProperties`, `additionalProperties`, and `items`, so I added tests and implementation.

src/interpreter/Interpreter.ts
```diff
@@ -90,43 +91,54 @@ export class Interpreter {
...
 
   /**
-  * This function splits up a model if needed and add the new model to the list of models.
+  * This function splits up a model if it is determined it should.
   * 
   * @param model check if it should be split up
-  * @param models which have already been split up
   */
-  private splitModels(model: CommonModel): CommonModel {
+  private tryAndSplitModels(model: CommonModel): CommonModel {
```
magicmatatjahu, line 141:
```suggestion
  private trySplitModels(model: CommonModel): CommonModel {
```

src/interpreter/Interpreter.ts
```diff
@@ -138,18 +150,36 @@ export class Interpreter {
...
 
   /**
-   * Split up all models which should and use ref instead.
+   * Split up all model which should be and use $ref instead.
    * 
-   * @param model to ensure are split
-   * @param models which are already split
+   * @param model
    */
   ensureModelsAreSplit(model: CommonModel): void {
```
magicmatatjahu, line 157:
```suggestion
  ensureModelsAreSplitted(model: CommonModel): void {
```

magicmatatjahu (commented):
Only two changes :)

src/interpreter/Interpreter.ts
jonaslagoni: This sounds weird? "splitted" are not a word 🙂

src/interpreter/Interpreter.ts
jonaslagoni: I don't know why, but this sounds a bit weird as well 🤔? Does this make more sense for you?

src/interpreter/Interpreter.ts
magicmatatjahu: 😅 I was wondering too but I found the use of this word in some articles ... my mistake.

src/interpreter/Interpreter.ts
magicmatatjahu: My problem is `tryAnd...`, but `try` what? 😄

jonaslagoni: "try split models" vs "try and split models". When I read the name of the function the latter sounds more natural vs the first one 🤔

magicmatatjahu: As you want :) I don't think that we should use `And` and `Or` in function and variable names.

jonaslagoni: Yea true.

jonaslagoni: Sure lets change it 😄

jonaslagoni: Changed

magicmatatjahu: Ok, If you want, let's wait for last review from czlowiek488 :)

jonaslagoni:
@czlowiek488 do you want to do a final review as well?

magicmatatjahu (approved):
LGTM! :)

test/processors/JsonSchemaInputProcessor.spec.ts
czlowiek488: @jonaslagoni I mean that you should write inmutable code.

jonaslagoni:
@czlowiek488  with no specific reply whether you wanted to review it, I am merging this PR now 🙂

czlowiek488 (commented):
It is fine for me

PR #219, opened 2021-05-31
ci: added sonar cloud config to ignore duplicate code

jonaslagoni:
**Description**
This PR adds a sonarcloud configuration file to ignore duplicates lines for renderers.


sonar-project.properties
```diff
@@ -0,0 +1 @@
+sonar.cpd.exclusions=src/generators/javascript/renderers/ClassRenderer.ts, src/generators/java/renderers/ClassRenderer.ts, src/generators/typescript/renderers/ClassRenderer.ts, src/generators/javascript/JavaScriptRenderer.ts, src/generators/typescript/renderers/EnumRenderer.ts, src/generators/java/renderers/EnumRenderer.ts
```
magicmatatjahu, line 1: Probably you can use wildcards here :)

magicmatatjahu (commented):
You should rename PR title to `ci:` not `docs:`

sonar-project.properties
jonaslagoni: I disagree actually. 

I think we have to specifically say which files should have removed this check. Otherwise, we won't catch certain checks on files we didnt approve. If that makes sense 😄

magicmatatjahu: probably renderers will have always duplication for `renderProperties` etc :)

jonaslagoni: Yea true, but then we must explicitly say it is 🙂 

I just wonder whether this config file should be applied now or not until it is merged 🤔 ?

magicmatatjahu: I don't know, but sonar is "old" solution, so probably after merging. Do you want to accept it?

magicmatatjahu (approved):

sonar-project.properties
magicmatatjahu: Please also add 
image
files

jonaslagoni: They are fixed by the new interpreter, so it solves it self.

magicmatatjahu (approved):
LGTM

jonaslagoni:
Lets 🤞

PR #220, opened 2021-06-01
feat: support tuples as array types

jonaslagoni:
**Description**
This PR introduces tuples for arrays.

Blocked by #218 

**Related issue(s)**
fixes https://github.com/asyncapi/generator-model-sdk/issues/67


src/models/CommonModel.ts
```diff
@@ -359,37 +385,47 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+      }
+    }
+
+    //mergeFrom is a tuple && mergeTo is not
+    if (Array.isArray(mergeFrom.items) && !Array.isArray(mergeToItems)) {
+      mergeTo.items = [];
+      for (const [index, mergeFromTupleModel] of mergeFrom.items.entries()) {
+        mergeTo.items[Number(index)] = CommonModel.mergeCommonModels(mergeToItems, mergeFromTupleModel, originalSchema, alreadyIteratedModels); 
+      }
+    }
```
magicmatatjahu, line 423: You can merge these ifs to single one `if (Array.isArray(mergeFrom.items)) { if(Array.isArray(mergeToItems)) {} else ... }`

src/models/CommonModel.ts
```diff
@@ -123,6 +124,31 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+   * If a item already exist it will be merged.
+   * 
+   * @param tupleModel 
+   * @param schema 
+   * @param index 
+   */
+  addItemTuple(tupleModel: CommonModel, schema: Schema, index: number): void {
+    let modelItems = this.items;
+    if (!Array.isArray(modelItems)) {
+      Logger.warn('Trying to add item tuple to a non-tuple item, will drop existing item model', tupleModel, schema, index);
```
magicmatatjahu, line 139: Maybe there are cases, when we should throw an error, when items were single schema and then we change to tuple 🤔

docs/interpretation_of_JSON_Schema_draft_7.md
```diff
@@ -54,7 +54,10 @@ If only one side has a property defined, it is used as is, if both have it defin
 - `additionalProperties` if both models contain it the two are recursively merged together. 
 - `patternProperties` if both models contain a pattern the corresponding models are recursively merged together. 
 - `properties` if both models contain the same property the corresponding models are recursively merged together. 
-- `items` if both models contain items they are recursively merged together.
+- `items` are merged together based on a couple of rules:
+    - If both models are simple arrays those item models are merged together as is.
+    - If both models are tuple arrays each tuple model (at specific index) is merged together.
+    - If either one side is different from the other, the simple array is merged together with each tuple model.
```
magicmatatjahu, line 60: Do you think it's a correct behaviour? If we found tuple and we have defined items as standalone schema, why do you want merge models? I think we should override only when two models are tuple or single schema, otherwise, tuple should override items.

magicmatatjahu (commented):
Great job! I have some problem with merging items. Please describe your point of view for this problem 😅

src/models/CommonModel.ts
jonaslagoni: I tried to do that, but I feel like it the maintainability declines and it is harder to grasp what is happening 🤷 

Do you think it is better to merge them?

magicmatatjahu: The only problem for me is constantly checking one condition, but i can see that it is now easier to understand, with multiple ifs

src/models/CommonModel.ts
jonaslagoni: Because of the restrictiveness of tuples, that should always be preferred over simple array type. So I feel like this is the correct approach.

In which cases would you rather have the generation fail? 🤔

magicmatatjahu: This logic for tuple/items will be only when user will define combined schema like `oneOf/anyOf` etc, then we should consider it as union type, not override it, but at the moment, as I remember, we don't support union types explicit :) Am I right? 😅

docs/interpretation_of_JSON_Schema_draft_7.md
jonaslagoni: You are right... Also, tuples should ONLY be considered when `additionalItems` is false (Default is `true`), otherwise, you are allowed to have multiple values in your array so it can no longer be considered a tuple.

jonaslagoni: Unless `maxItems` is the same length as the `items` defined, then it can be considered a tuple.

jonaslagoni: Ohh the posibilities 😅

magicmatatjahu: > You are right... Also, tuples should ONLY be considered when additionalItems is false (Default is true), otherwise, you are allowed to have multiple values in your array so it can no longer be considered a tuple.

I think that if you have `additionalItems` set to true, then `tuple` are handled as prefix items :)

jonaslagoni: @magicmatatjahu can you clarify what you mean by prefix items?

magicmatatjahu: when you have (eg) `items: [{ type: string }, { type: number }]` and `additionalItems : true` then you can define array as ["foo", 1, "something"], so `items` are treated as "prefix" items :)

jonaslagoni: Ahh yes, added another issue for further discussion https://github.com/asyncapi/generator-model-sdk/issues/224

src/models/CommonModel.ts
jonaslagoni: You are, changed it.

src/models/CommonModel.ts
```diff
@@ -359,39 +385,40 @@ export class CommonModel extends CommonSchema<CommonModel> {
...
+    //mergeFrom and mergeTo is tuple
+    if (Array.isArray(mergeFrom.items) && Array.isArray(mergeToItems)) {
+      for (const [index, mergeFromTupleModel] of mergeFrom.items.entries()) {
+        (mergeTo.items as CommonModel[])[Number(index)] = CommonModel.mergeCommonModels(mergeToItems[Number(index)], mergeFromTupleModel, originalSchema, alreadyIteratedModels); 
       }
-    } else if (mergeTo.items !== undefined) {
-      mergeTo.items = merge(mergeTo.items);
     }
+
+    //mergeFrom is a tuple && mergeTo is not, use as is
```
magicmatatjahu, line 417:
```suggestion
    //mergeFrom is a tuple && mergeTo is not, use mergeFrom items (the tuple is prioritized)
```

magicmatatjahu (commented):
LGTM, but only one suggestion to comment.

jonaslagoni:
> LGTM, but only one suggestion to comment.

Good call 👍 Thanks!

magicmatatjahu (approved):
LGTM! :)

PR #221, opened 2021-06-03
refactor: add post interpreter

jonaslagoni:
**Description**
This PR introduces a post-interpreter that takes the CommonModel representation of the schema and splits up any models into separate ones. This makes the interpreter process cleaner.

**Related issue(s)**
fixes https://github.com/asyncapi/generator-model-sdk/issues/195


src/interpreter/InterpretNot.ts
```diff
@@ -17,15 +17,13 @@ export default function interpretNot(schema: Schema, model: CommonModel, interpr
   if (typeof schema.not === 'object') {
     const notSchema = schema.not;
     const newInterpreterOptions: InterpreterOptions = {
-      ...interpreterOptions, 
-      splitModels: false,
```
magicmatatjahu, line 21: I don't see this option in other "new" places. We will split models by default from now, is it a good?

src/interpreter/InterpretNot.ts
```diff
@@ -17,15 +17,13 @@ export default function interpretNot(schema: Schema, model: CommonModel, interpr
...
     };
-    const interpretedModels = interpreter.interpret(notSchema, newInterpreterOptions);
-    if (interpretedModels.length > 0) {
-      const interpretedModel = interpretedModels[0];
-      if (interpretedModel.type !== undefined) {model.removeType(interpretedModel.type);}
-      if (interpretedModel.enum !== undefined) {model.removeEnum(interpretedModel.enum);}
+    const notModel = interpreter.interpret(notSchema, newInterpreterOptions);
+    if (notModel !== undefined) {
+      if (notModel.type !== undefined) {model.removeType(notModel.type);}
+      if (notModel.enum !== undefined) {model.removeEnum(notModel.enum);}
```
magicmatatjahu, line 26: I know that `notModel.type !== undefined` is more readable, but maybe we should go with

```js
notModel.type && model.removeType(notModel.type);
```

magicmatatjahu (commented):
More and more I don't understand "new" code 😅 I added some suggestions.

src/interpreter/InterpretNot.ts
jonaslagoni: The reason why it is no longer located there is because we no longer try to keep a list of models that should be returned. The interpreter method ONLY tries to interpret the JSON Schema to a single CommonModel and that's it. 

The reason it was added was that the odd level `not` schema should at that time not be saved to a separate model. 

Does that make sense? 🤔

src/interpreter/PostInterpreter.ts
```diff
@@ -0,0 +1,70 @@
+import { CommonModel } from '../models';
```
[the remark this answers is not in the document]
jonaslagoni: Hmm, do you think that makes sense? especially since it processes it as is using "simple" functions? 🤔

src/interpreter/InterpretNot.ts
jonaslagoni: hmmm, good suggestion! I just don't know if I like that format 😄

jonaslagoni:
> More and more I don't understand "new" code 😅 I added some suggestions.

Hmm, it should not go in that direction!

What do you think is confusing about it?

magicmatatjahu:
@jonaslagoni Rather situation that I haven't been in the code for 2-3 months and I don't know exactly how it works underneath, so it's only a confusing for me.

src/interpreter/InterpretNot.ts
magicmatatjahu: You can do this in the separate PR or create (another) issue 😅

src/interpreter/InterpretNot.ts
magicmatatjahu: Ok, so from this PR it will be split by default. :)

src/interpreter/PostInterpreter.ts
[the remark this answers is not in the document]
magicmatatjahu: Except that "interpretation" helper functions we have everything written in classes, so it's only a suggestion to have it also in class.

src/interpreter/InterpretNot.ts
jonaslagoni: It will yes, because at the time the post-interpreter is run, only the core CommonModel is present 🙂

src/interpreter/InterpretNot.ts
jonaslagoni: This syntax actually violates the eslint rule eslintno-unused-expressions. So it cannot be such a format 🙂

magicmatatjahu: So eslint sucks in this place 😄

jonaslagoni:
> @jonaslagoni Rather situation that I haven't been in the code for 2-3 months and I don't know exactly how it works underneath, so it's only a confusing for me.

Fair enough, that's understandable 🙂

src/interpreter/InterpretNot.ts
jonaslagoni: I mean, to me this syntax is honestly better but it is a matter of opinion I guess 😄

magicmatatjahu: It was a joke :)

src/interpreter/PostInterpreter.ts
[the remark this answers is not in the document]
magicmatatjahu: but ok, we can stay with these functions.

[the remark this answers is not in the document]
jonaslagoni: Fair point.

The reason I didn't wrap it in a class was that I didn't find the need. The Interpreter class was needed because it had to have access to a counter for anonym models as well as what schemas it already iterated. Passing these from function to function would have been annoying IMO.

src/interpreter/InterpretNot.ts
jonaslagoni: 😆 Sure! I know you like your sweet short syntax though! So it might as well could have been real xD

magicmatatjahu (approved):
LGTM! :)

PR #231, opened 2021-06-09
feat: split enums and use as values

### LINKED ISSUE #188: Generate enum and use it for values.
#### Context
I have a function `saveProject` which save a asyncApiMessage as a document to a database.
This document is message wrapped in Envelope with additional properties.
A message looks like this:
```ts
export interface MessageProjectCreated { //this is what is generated from generator-model-sdk everything else is my code
   eventName: "tms.event.project.created";
};
export const message: MessageProjectCreated = {
   eventName: "tms.event.project.created",
};
```
My document look like this:
```ts
export enum EventName {
  ProjectCreated = "tms.event.project.created",
  ...
};
export interface Envelope<T> {
   name: EventName;
   payload: T;
};

export const document: Envelope<MessageProjectCreated> = {
   name: message.eventName,
   payload: event,
};
```
#### Reason#1
My problem with this is that typescript treat generated "tms.event.project.created" as a string not as part of EventName.
To make this work I had to replace Envelope with this.
```ts
export interface Envelope<T> {
   name: EventName | string; // a hacky way to make use of generated types
   payload: T;
};
```

#### Reason#2
I have to declare EventName enum by my own which can lead to unconsistency between event names from specification and enum written by me.

#### Description
I need a way to create enum in asyncapi and a way to specify field type as enum element.
So something like this: (I modified json below from [this answer](https://stackoverflow.com/a/63620260/9731304) in SO)
```json
{
  "$id":"MessageProjectCreated",
  "properties": {
    "eventName": {
      "$ref": "http://example.com/some_schema.json#/definitions/EventName/tms.event.project.created"
    }
  }
}
```
```json
{
  "$id": "http://example.com/some_schema.json",
  "definitions": {
    "EventName": {
      "enum": [ "tms.event.project.created"]
    }
  }
}
```
should generate something like this.
```ts
export enum EventName {
  ProjectCreated = "tms.event.project.created",
};
export interface MessageProjectCreated {
   eventName: EventName.ProjectCreated;
};

```

jonaslagoni:
**Description**
This PR introduces a couple of things:
1. Splits up enums into separate models so they can be rendered as so.
2. Ensure that enum models have ids
3. Switch from `clearAllMocks` to `resetAllMocks` as it is causing problems if a mock return value is not overwritten.
4. Anonym ids are now given backwards, because we require that a model is fully determined to know the full state. 
 

**Related issue(s)**
fixes #188


jonaslagoni:
Blocked by #238

magicmatatjahu (approved):

src/interpreter/Interpreter.ts
```diff
@@ -84,6 +76,13 @@ export class Interpreter {
       this.interpretAndCombineSchema(schema.else, model, schema, interpreterOptions);
 
       interpretNot(schema, model, this, interpreterOptions);
+
+      //All schemas of type model object or enum MUST have ids
+      if (isModelObject(model) === true || isEnum(model) === true) {
```
magicmatatjahu, line 81: So we end with helper functions to check what type model has 😅 Creating an EnumModel class or the ObjectModel class would be easier in this way as I suggested in long time ago :P

magicmatatjahu: And we must remember, `enum` is a generic keyword, it can be also combined with `object` type.

test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -14,16 +14,16 @@ describe('TypeScriptGenerator', () => {
       }
     };
     const expected = `export class Address {
-  private _state?: "Texas" | "Alabama" | "California" | "other";
+  private _state?: State;
```
magicmatatjahu, line 17: You should also test here the shape of `State` enum :) As I remember with

```ts
    const model = inputModel.models['State'];
```

function.

magicmatatjahu (commented):
Only one suggestion to test.

src/interpreter/Interpreter.ts
jonaslagoni: It can yea, which also makes the separate classes difficult to work with.

jonaslagoni: But we definitely have to keep an eye out for whether we can reduce the complexity of the model in some way in the future.

magicmatatjahu: we can have `modelType` in `CommonModel` and then don't need check it in every place.

jonaslagoni: Hmm, we could yea, not sure it is needed when we have the support functions? 🤔 

But feel free to create an issue 🙂

test/generators/typescript/TypeScriptGenerator.spec.ts
jonaslagoni: Technically, this test should be removed as we no longer have `union property type` with enums 🤔

jonaslagoni: I have removed it.

magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks 👍

PR #233, opened 2021-06-09
fix: nested object models are not split

jonaslagoni:
**Description**
This PR fixes a bug where nested models (i.e. first encountered model is not an object model but the nested one is) are not split.


magicmatatjahu (approved):
LGTM

PR #239, opened 2021-06-10
feat: make it possible to add dependencies to rendering of models

### LINKED ISSUE #238: Generated model requires unknown imports
#### Describe the bug
At the moment the generators have no way of telling the user which imports they require.

For example, for the Java generator, it generates enums with the expectation of `com.fasterxml.jackson` library to be present, however, the user has no way of knowing it is.

#### Expected behavior
I expected to be able to generate the model as is, without having to require me to dig into the generator to figure out which imports/required libraries are needed.

jonaslagoni:
**Description**
This PR adds a `dependencies` preset which can be used to add dependencies to the model rendering.

**Related issue(s)**
fixes #238

### DIFF
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -96,6 +96,18 @@ const generator = new TypeScriptGenerator({ presets: [
 ] });
 ```
 
+## Adding new dependencies
+Each preset hook has the possibility of adding its own dependencies that needs to be rendered. It can be done through the `renderer` argument.
+
+```ts
+...
+self({ renderer, content }) {
+  renderer.addDependency('import java.util.*;');
+  return content;
+}
+...
+```
+
 ## Overriding the default preset
 
 Each implemented generator for appropriate language must have defined the default preset. However, we can override it by passing it as the `defaultPreset` parameter in the generator options. Check the example for TypeScript generator:
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -115,7 +115,6 @@ const generator = new TypeScriptGenerator({ defaultPreset: DEFAULT_PRESET });
 For each model type, you can implement two basic methods:
 
 - `self` - the method for extending the model shape.
-- `dependencies` - the method for adding dependencies for the model.
 - `additionalContent` - the method which adds additional content to the model.
 
 Each customization method receives the following arguments:
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -115,6 +115,7 @@ const generator = new TypeScriptGenerator({ defaultPreset: DEFAULT_PRESET });
 For each model type, you can implement two basic methods:
 
 - `self` - the method for extending the model shape.
+- `dependencies` - the method for adding dependencies for the model.
 - `additionalContent` - the method which adds additional content to the model.
 
 Each customization method receives the following arguments:
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -97,7 +97,8 @@ const generator = new TypeScriptGenerator({ presets: [
 ```
 
 ## Adding new dependencies
-Each preset hook has the possibility of adding its own dependencies that needs to be rendered. It can be done through the `renderer` argument.
+
+Each preset hook has the possibility of adding its own dependencies that needs to be rendered for the given model. It can be done through the `addDependency` function from `renderer` property.
 
 ```ts
 ...
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -20,7 +20,6 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
    * @param dependency complete dependency string so it can be rendered as is.
    */
   addDependency(dependency: string): void {
-    this.dependencies = this.dependencies || [];
     if (!this.dependencies.includes(dependency)) {
       this.dependencies.push(dependency);
     }
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -19,7 +19,7 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
    * 
    * @param dependency complete dependency string so it can be rendered as is.
    */
-  addUniqueDependency(dependency: string): void {
+  addDependency(dependency: string): void {
     this.dependencies = this.dependencies || [];
     if (!this.dependencies.includes(dependency)) {
       this.dependencies.push(dependency);
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -11,6 +11,7 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
     protected readonly presets: Array<[Preset, unknown]>,
     protected readonly model: CommonModel, 
     protected readonly inputModel: CommonInputModel,
+    public dependencies: string[] = []
   ) {}
 
   renderLine(line: string): string {
@@ -36,15 +37,18 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
     return this.runPreset('self');
   }
 
+  runDependenciesPreset(): Promise<string[]> {
+    return this.runPreset<string[]>('dependencies');
+  }
+
   runAdditionalContentPreset(): Promise<string> {
     return this.runPreset('additionalContent');
   }
-
-  async runPreset(
+  async runPreset<RT = string>(
     functionName: string,
     params: Record<string, unknown> = {},
-  ): Promise<string> {
-    let content = '';
+  ): Promise<RT> {
+    let content;
     for (const [preset, options] of this.presets) {
       if (typeof preset[String(functionName)] === 'function') {
         content = await preset[String(functionName)]({ 
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -14,6 +14,18 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
     public dependencies: string[] = []
   ) {}
 
+  /**
+   * Adds a dependency while ensuring that only one dependency is preset at a time.
+   * 
+   * @param dependency complete dependency string so it can be rendered as is.
+   */
+  addUniqueDependency(dependency: string): void {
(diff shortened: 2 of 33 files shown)


magicmatatjahu (commented):
Before review, I have some questions:

- I wonder of using preset for dependencies. First I thought about writing to `Renderer` `dependencies` property, because each model has separate renderer (thanks God that I didn't change it 😅) and use the `selfPreset` to render dependencies at the top of the model. We can add then to `Renderer` some helper functions to handle dependencies like `addDependency`, `removeDependency` etc.
- some languages (like js/ts) support aliases etc for imports so I think that `string[]` isn't a good type for dependencies. I think that we should create some object for each import with `path` and `aliases` fields - we should think also about another possibilites

jonaslagoni:
>     * I wonder of using preset for dependencies. First I thought about writing to `Renderer` `dependencies` property, because each model has separate renderer (thanks God that I didn't change it 😅) and use the `selfPreset` to render dependencies at the top of the model. We can add then to `Renderer` some helper functions to handle dependencies like `addDependency`, `removeDependency` etc.

My thought behind adding it to a preset hook is because users, when customizing the output potentially want to add custom dependencies. And if it was not done through a preset hook it would look weird to add dependencies through helper functions of the `Renderer`. Especially if you change the property output that runs for each property in the model, then you would need to call the same function multiple times? That does not make much sense. And handling this in any of the other functions also makes little sense from the user perspective (IMO).

>     * some languages (like js/ts) support aliases etc for imports so I think that `string[]` isn't a good type for dependencies. I think that we should create some object for each import with `path` and `aliases` fields - we should think also about another possibilites

Makes sense yea, but what if we just simplified it to providing the full import string? 

So instead of saying 
```
  dependencies() {
        return ['com.fasterxml.jackson.annotations.*'];
  }
```
we do
```
  dependencies() {
        return ['import com.fasterxml.jackson.annotations.*;'];
  }
```
Meaning we provide the full statement as is? Is there any benefit to splitting it up into an object? 🤔

magicmatatjahu:
> Meaning we provide the full statement as is? Is there any benefit to splitting it up into an object? 🤔

Only benefit if someone wants render deps in custom way, but returning full import statement is better :)

> My thought behind adding it to a preset hook is because users, when customizing the output potentially want to add custom dependencies. And if it was not done through a preset hook it would look weird to add dependencies through helper functions of the Renderer. Especially if you change the property output that runs for each property in the model, then you would need to call the same function multiple times? That does not make much sense. And handling this in any of the other functions also makes little sense from the user perspective (IMO).

I agree that it's not a good DX, but you don't need to save dependency in `property` preset but in `self` preset, because it's executed once for model, of course for single preset :) My main problem with separate preset is that we shouldn't create preset for every possible case, but reuse existing one (in this case `self` and describe how to add dependencies in docs), I don't want to end with 20+ presets :P

jonaslagoni:
@magicmatatjahu that's understandable, also why I didn't add a `additionalProperties´ hook in #240.

I tried to change it to using `self`, but the problem is self is used to prepend code as well. So in that case, how does the customization know that the super preset requires the dependency? - https://github.com/asyncapi/modelina/blob/be178009cfd1dd8a1bd11d7485940a71f1590849/test/generators/java/JavaGenerator.spec.ts#L303

magicmatatjahu:
@jonaslagoni `dependencies` hook is one way to achieve solution. My main problem with `dependencies` hook and with `additionalProperties` hook (as I remember I had also problem with that in the latest PR for it) is that you create another hooks and I don't want to end with several hooks - also due to this we don't have `prepend`, `postConstructor` etc hooks, because they can be achieved with existing one.

> So in that case, how does the customization know that the super preset requires the dependency? Located here:

The super doesn't need to know. Dependency should be added by other presets given during rendering in `self` hook. The one problem what I see is that then we will have to make this:

```ts
self({ renderer, content }) {
  renderer.addUniqueDependency('import com.fasterxml.jackson.annotations.*;');
  return content;
}
```

but we can change the behaviour of returning `undefined` from preset - then we should pass the content from earlier preset to the next one and `self` can works similar to your `dependencies`.

Also regarding to your changes. Please know that you have `dependencies` property in each render - you wrote it in the `AbstractRenderer` - so you can also change deps inside the hooks :) 

Really, please consider using `addUniqueDependency` function in `self` preset.

Some example when adding deps in hooks is better than `dependencies` hook: when you have complex logic with adding some annotations to properties/class/method in Java class you can add corresponding deps inside corresponding `if/else/switch cases`, otherwise you must duplicate code for `dependencies` hooks. For complex I mean that you can end with adding annotations when given class/property has required metadata.

src/generators/AbstractRenderer.ts
```diff
@@ -14,6 +14,18 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
...
+   * Adds a dependency while ensuring that only one dependency is preset at a time.
+   * 
+   * @param dependency complete dependency string so it can be rendered as is.
+   */
+  addUniqueDependency(dependency: string): void {
+    this.dependencies = this.dependencies || [];
+    if (!this.dependencies.includes(dependency)) {
+      this.dependencies.push(dependency);
+    }
+  }
```
magicmatatjahu, line 27: I think that we should rename it to shortest name. `addDependency` is enough :)

```suggestion
  addDependency(dependency: string): void {
    this.dependencies = this.dependencies || [];
    if (!this.dependencies.includes(dependency)) {
      this.dependencies.push(dependency);
    }
  }
```

magicmatatjahu (changes requested):
Thanks for considering my idea :) Please add small docs for new feature. Maybe you should create new doc or add info about new function in `customization` doc. Also, please remember to document that output model have `dependencies` field.

src/generators/AbstractRenderer.ts
magicmatatjahu: Also consider to write `removeDependency` function.

jonaslagoni: I dont see a reason to add it at the moment? 🤔We can always add it later if we ever get a use-case for it.

Regarding the renaming, hmmm... You really think `addDependency` is better? If it only added to the array then it would make sense, but since it has a check before doing it it makes less sense to me 🤔

magicmatatjahu: > Regarding the renaming, hmmm... You really think addDependency is better? If it only added to the array then it would make sense, but since it has a check before doing it it makes less sense to me 🤔

image

> I dont see a reason to add it at the moment? 🤔We can always add it later if we ever get a use-case for it.

I agree :)

src/generators/java/presets/ConstraintsPreset.ts
```diff
@@ -7,9 +7,12 @@ import { CommonModel } from '../../../models';
  * 
  * @implements {JavaPreset}
  */
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 export const JAVA_CONSTRAINTS_PRESET: JavaPreset = {
   class: {
+    self({renderer, content}) {
+      renderer.addUniqueDependency('import javax.validation.constraints.*;');
+      return content;
```
magicmatatjahu, line 14: I think that we can change the line https://github.com/asyncapi/modelina/blob/master/src/generators/AbstractRenderer.ts#L50 and return (regarding to my last comment in PR) the latest not falsey value of content, so we can change the mentioned line to:

```ts
        content = await preset[String(functionName)]({ 
            ...params,           
            renderer: this,           
            content,           
            options,           
            model: this.model,           
            inputModel: this.inputModel        
        }) || content;
```

and then you don't need to write `return content;` in every preset if you add only deps :)

magicmatatjahu (commented):
I wrote this suggestion in the last review but I don't see it 🤦🏼

src/generators/java/presets/ConstraintsPreset.ts
jonaslagoni: While I would do that, it requires further refactoring and testing, which I do not want to introduce in this PR 🙂 

Good suggestion though, maybe we should add a new issue?

src/generators/AbstractRenderer.ts
jonaslagoni: Fine 😂 added.

src/generators/java/presets/ConstraintsPreset.ts
magicmatatjahu: Ok, make it as followup :)

jonaslagoni: @magicmatatjahu are you gonna add the issue or should I?

magicmatatjahu: Created https://github.com/asyncapi/modelina/issues/247

jonaslagoni:
@magicmatatjahu ready for review again 🙂

docs/presets.md
```diff
@@ -96,6 +96,18 @@ const generator = new TypeScriptGenerator({ presets: [
 ] });
 ```
 
+## Adding new dependencies
+Each preset hook has the possibility of adding its own dependencies that needs to be rendered. It can be done through the `renderer` argument.
```
magicmatatjahu, line 100:
```suggestion

Each preset hook has the possibility of adding its own dependencies that needs to be rendered for the given model. It can be done through the `addDependency` function from `renderer` property.
```

src/generators/AbstractRenderer.ts
```diff
@@ -11,8 +11,21 @@ export abstract class AbstractRenderer<O extends CommonGeneratorOptions = Common
...
+    public dependencies: string[] = []
   ) {}
 
+  /**
+   * Adds a dependency while ensuring that only one dependency is preset at a time.
+   * 
+   * @param dependency complete dependency string so it can be rendered as is.
+   */
+  addDependency(dependency: string): void {
+    this.dependencies = this.dependencies || [];
```
magicmatatjahu, line 23:
```suggestion
```

If you add as default value in the constructor for `dependencies` the empty array, you shouldn't reassign it here :) I know why you did this, but it's unnecessary.

magicmatatjahu (commented):
Two suggestions.

magicmatatjahu (approved):
LGTM! :)

jonaslagoni:
Thanks 👍

Issue #198, opened 2021-05-17
Start using Any type for true schemas


magicmatatjahu:
Currently we are using `["boolean", "string"...]` type for `true` JSON Schema. We should move to more verbose `any` type for such a schema. I know that it's not a compatible with JSON Schema but we should go with easier code as well.

jonaslagoni:
#365 solves this issue.

PR #240, opened 2021-06-10
feat: render additionalProperties

### LINKED ISSUE #68: should render additionalProperties

jonaslagoni:
**Description**
This PR ensures that additionalProperties are rendered. It introduces:
- Disables `no-unused-vars` and enable `@typescript-eslint/no-unused-vars` as a replacement as the normal one is faulty in typescript.
- Introduces a `type` argument to rendering properties as we have different types of properties that should be rendered differently. 
- Adds the rendering of `additionalProperties` alongside accessor functions.
- Ensures that the correct property name is used since we must handle if a model already contains the property `additionalProperties`.

**Related issue(s)**
fixes #68


test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -154,6 +163,7 @@ ${content}`;
   members?: string | number | boolean;
   tupleType?: [string, number];
   arrayType: Array<string>;
+  additionalProperties?: Map<String, object | string | number | Array<unknown> | boolean | null | number>;
```
magicmatatjahu, line 166: LOL, we should introduce the `any` type, really 😂

test/generators/java/JavaGenerator.spec.ts
```diff
@@ -75,27 +80,37 @@ describe('JavaGenerator', () => {
...
             const annotation = renderer.renderAnnotation('JsonProperty', `"${propertyName}"`);
-            return `${annotation}\n${content}`;
+            return `${annotation}
+${content}`;
           },
           setter({ renderer, propertyName, content }) {
             const annotation = renderer.renderAnnotation('JsonProperty', `"${propertyName}"`);
-            return `${annotation}\n${content}`;
+            return `${annotation}
+${content}`;
```
magicmatatjahu, line 113: I don't like this approach, because then you must write the next line from beginning of line, we should stay with `\n`.

src/helpers/NameHelpers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Recursively find the proper property name for additionalProperties
+ * 
+ * This function ensures that the property name for additionalProperties is unique 
+ * 
+ * @param rootModel 
+ * @param additionalPropertyName 
+ */
+export function findPropertyNameForAdditionalProperties(rootModel: CommonModel, additionalPropertyName = 'additionalProperties'): string {
```
magicmatatjahu, line 11: `findPropertyNameForAdditionalProperties` Java in JS 😂 Maybe another name like 

image

```suggestion
export function getUniquePropertyName(rootModel: CommonModel, propertyName = 'additionalProperties'): string {
```

test/generators/typescript/TypeScriptGenerator.spec.ts
magicmatatjahu: For TS we can add here https://github.com/asyncapi/modelina/blob/master/src/generators/typescript/TypeScriptRenderer.ts#L22 logic to check this 7 types and then render `any`

magicmatatjahu (commented):
Overall good, only suggestion for changing the name of one function.

test/generators/typescript/TypeScriptGenerator.spec.ts
jonaslagoni: agreed we should 👍

test/generators/java/JavaGenerator.spec.ts
jonaslagoni: Sorry! This was an unintended side effect of a search/replace 😄 Gonna revert this

src/helpers/NameHelpers.ts
jonaslagoni: hahahaha, true 😂

src/generators/typescript/TypeScriptRenderer.ts
```diff
@@ -96,16 +100,29 @@ ${lines.map(line => ` * ${line}`).join('\n')}
...
-    return `${name}${signature};`;
+    
+    if (type === PropertyType.property) {
+      const signature = this.renderTypeSignature(property, { isRequired: this.model.isRequired(propertyName) });
+      return `${name}${signature};`;
+    } else if (type === PropertyType.additionalProperty) {
+      const additionalPropertyType = this.renderType(property);
+      return `${name}?: Map<String, ${additionalPropertyType}>;`;
+    }
+    return '';
```
magicmatatjahu, line 122: Please use the `switch` statement.

jonaslagoni: 👍 Checked the other locations, don't feel like a switch statement there makes sense.

src/generators/typescript/TypeScriptRenderer.ts
```diff
@@ -96,16 +100,32 @@ ${lines.map(line => ` * ${line}`).join('\n')}
...
+    case PropertyType.property:
+      signature = this.renderTypeSignature(property, { isRequired: this.model.isRequired(propertyName) });
+      return `${name}${signature};`;
+      break;
+    case PropertyType.additionalProperty:
+      signature = this.renderType(property);
+      return `${name}?: Map<String, ${signature}>;`;
+    default:
+      return '';
+    }
```
magicmatatjahu:
```suggestion
    switch (type) {
    case PropertyType.property:
      const signature = this.renderTypeSignature(property, { isRequired: this.model.isRequired(propertyName) });
      return `${name}${signature};`;
    case PropertyType.additionalProperty:
      const signature = this.renderType(property);
      return `${name}?: Map<String, ${signature}>;`;
    default:
      return '';
    }
```

magicmatatjahu (commented):
Last suggestion. It looks weird :D

src/generators/typescript/TypeScriptRenderer.ts
jonaslagoni: then we have to disable `no-case-declarations` which causes `Unexpected lexical declaration in case block`

test/generators/typescript/TypeScriptGenerator.spec.ts
jonaslagoni: But we have an issue for it #198

src/generators/typescript/TypeScriptRenderer.ts
magicmatatjahu: So we need this `let signature: string;` outside the switch?

magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks 🙇

PR #245, opened 2021-06-11
refactor: replace `switch` statements with `if else`.

### LINKED ISSUE #236: Ensure consistancy between switch statements
#### Reason/Context
Currently, we have very small switch statements which are considered a [code smell by sonarcloud](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&open=AXjKlPuxVLGRvzF4eQNt&resolved=false&rules=typescript%3AS1301). We should fix all of these issues so we get rid of them.

Make sure to read the readme file [how to contribute](https://github.com/asyncapi/modelina#contributing), if you have any questions feel free to reach out here, or in our [Slack](https://www.asyncapi.com/slack-invite).

Blakelist7:
This commit replaces `switch` statements with `if else` since there are very small `switch` statements at some places which have only 1 case.

Signed-off-by: Kanwal Singh <kanwal.88singh@gmail.com>

**Related issue(s)**

Fixes #236

### DIFF
--- a/src/generators/java/JavaGenerator.ts
+++ b/src/generators/java/JavaGenerator.ts
@@ -27,12 +27,10 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
 
   render(model: CommonModel, inputModel: CommonInputModel): Promise<string> {
     const kind = TypeHelpers.extractKind(model);
-    switch (kind) {
-    case ModelKind.ENUM: {
+    if (kind === ModelKind.ENUM) {
       return this.renderEnum(model, inputModel);
     }
-    default: return this.renderClass(model, inputModel);
-    }
+    return this.renderClass(model, inputModel);
   }
 
   renderClass(model: CommonModel, inputModel: CommonInputModel): Promise<string> {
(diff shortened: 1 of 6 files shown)


jonaslagoni:
@Blakelist7 do you want me to point you in the right direction or tell you straight up where the error is? 🤔 

Depends from person to person how they want to be helped, which is why I am asking 🙂

Blakelist7:
Let's go forward in the right direction first, then if it does not work. I still have the other option, right ? 😄

jonaslagoni:
> Let's go forward in the right direction first, then if it does not work. I still have the other option, right ? 😄

Of course 😄 

Take a look at the switch statements and what values are evaluated vs what if statements are created in its place 🙂

Blakelist7:
In switch statements we are returning something after selecting a particular case but in if else statements I am not selecting, am just comparing it with a const value, right?

jonaslagoni:
It is not a matter of return statements, but the values that are evaluated in the switch statement vs the expression you added in the if statement. 

As the tests point out it is a change that happened in the Java enum generator (as it is the enum values that are wrongfully generated) 🙂

Blakelist7:
I think there was just a minor issue after fixing that everything worked smooth :). BTW Thanks.

jonaslagoni (approved):
Awesome, well done 🎉 thanks!

jonaslagoni:
@all-contributors please add @Blakelist7 for code

PR #257, opened 2021-06-18
feat: interpret additionalItems

### LINKED ISSUE #224: Support additionalItems in CommonModel
#### Reason/Context
The CommonModel currently has no knowledge of `additionalItems` that the array/tuple must also support. This should change in the same way as `additionalProperties` are implemented.

jonaslagoni:
**Description**
This PR adds the interpretation of `additionalItems` to the interpreter.

**Related issue(s)**
fixes #224

### DIFF
--- a/src/interpreter/InterpretAdditionalItems.ts
+++ b/src/interpreter/InterpretAdditionalItems.ts
@@ -12,7 +12,7 @@ import { Interpreter, InterpreterOptions } from './Interpreter';
  */
 export default function interpretAdditionalItems(schema: Schema, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
   if (model.type?.includes('array') === false) {return;}
-  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
+  const additionalItemsModel = interpreter.interpret(schema.additionalItems === undefined ? true : schema.additionalItems, interpreterOptions);
   if (additionalItemsModel !== undefined) {
     model.addAdditionalItems(additionalItemsModel, schema);
   }
--- a/src/interpreter/InterpretAdditionalItems.ts
+++ b/src/interpreter/InterpretAdditionalItems.ts
@@ -1,0 +1,19 @@
+import { CommonModel } from 'models';
+import { Schema } from 'models/Schema';
+import { Interpreter, InterpreterOptions } from './Interpreter';
+
+/**
+ * Interpreter function for JSON Schema draft 7 additionalProperties keyword.
+ * 
+ * @param schema
+ * @param model
+ * @param interpreter
+ * @param interpreterOptions to control the interpret process
+ */
+export default function interpretAdditionalItems(schema: Schema, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
+  if (model.type?.includes('array') === false) {return;}
+  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
+  if (additionalItemsModel !== undefined) {
+    model.addAdditionalItems(additionalItemsModel, schema);
+  }
+}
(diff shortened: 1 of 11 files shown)


src/interpreter/InterpretAdditionalItems.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Interpreter function for JSON Schema draft 7 additionalProperties keyword.
+ * 
+ * @param schema
+ * @param model
+ * @param interpreter
+ * @param interpreterOptions to control the interpret process
+ */
+export default function interpretAdditionalItems(schema: Schema, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
+  if (model.type?.includes('array') === false) {return;}
+  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
```
magicmatatjahu, line 15: `additionalItems` can be has `false` value, so this if isn't correct.

You should check first the `false` value:

```suggestion
  if (additionalItems === false) {return;}
  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
```

magicmatatjahu (changes requested):
I see that it's a copy from the `additionalProperties` but I add one comment, very important. I don't know if we should also check that statement and fix it in the `additionalProperties`.

src/interpreter/InterpretAdditionalItems.ts
jonaslagoni: Good catch!

Gonna refactor it for both 🙂

jonaslagoni:
Forgot docs 🤦

src/interpreter/InterpretAdditionalItems.ts
```diff
@@ -12,7 +12,7 @@ import { Interpreter, InterpreterOptions } from './Interpreter';
  */
 export default function interpretAdditionalItems(schema: Schema, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
   if (model.type?.includes('array') === false) {return;}
-  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
+  const additionalItemsModel = interpreter.interpret(schema.additionalItems === undefined ? true : schema.additionalItems, interpreterOptions);
```
magicmatatjahu, line 15: Not better `if (schema.additionalItems === false) {return;}` ?

src/interpreter/InterpretAdditionalProperties.ts
```diff
@@ -13,7 +13,7 @@ import { isModelObject } from './Utils';
  */
 export default function interpretAdditionalProperties(schema: Schema, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
   if (isModelObject(model) === false) {return;}
-  const additionalPropertiesModel = interpreter.interpret(schema.additionalProperties || true, interpreterOptions);
+  const additionalPropertiesModel = interpreter.interpret(schema.additionalProperties === undefined ? true : schema.additionalProperties, interpreterOptions);
```
magicmatatjahu, line 16: Not better `if (schema.additionalProperties === false) {return;}` ?

src/interpreter/InterpretAdditionalItems.ts
jonaslagoni: I find it a bit weird behavior tbh, to precheck the schema before trying to interpret it since if the interpreter encounter a false schema it returns undefined. Making the process more dynamic and more natural IMO 🙂

magicmatatjahu: Why weird? 😄 If someone is knows the JSON Schema when see the `schema.additionalItems === false` he/she knows then the next code will perform actions only on `undefined`, `true` or on object.

magicmatatjahu: IMHO

```ts
if (additionalItems === false) {return;}  const additionalItemsModel = interpreter.interpret(schema.additionalItems || true, interpreterOptions);
```

is more readable

jonaslagoni: It's weird because it should not be the sub-interpreters job to figure out how to interpret a schema | boolean, that's the job of the main interpret 🙂

magicmatatjahu: Let it be

magicmatatjahu (approved):
LGTM

jonaslagoni:
Thanks 💪

PR #263, opened 2021-06-18
feat: render tuple additionalItems for TS

### LINKED ISSUE #260: TypeScript generator should take additionalItems into consideration
#### Reason/Context
Even though we interpret additionalItems (#224) we do not take it into consideration when rendering the TS types for arrays. 

See https://github.com/asyncapi/modelina/issues/224#issuecomment-863200247 for further details on how it should be rendered.

jonaslagoni:
**Description**
Whenever an array is defined as a tuple, additionalItems control if it is allowed to specify addition items that may be added to the tuples so it acts more like an array.

**Related issue(s)**
fixes #260


magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks 🙇

PR #129, opened 2021-03-08
feat: add namingConvention option and pass generator to renderers

magicmatatjahu:
**Description**

- Add `namingConvention` option to options: naming for type and single property
- Pass `generator` to `renderers`
- Add `collectionType` option to Java generator to render `List<type>` or `type[]` signature for collections
- Add unit tests
- Update renderers and use `namingConvention` everywhere
- Update docs

**Related issue(s)**
Resolves https://github.com/asyncapi/generator-model-sdk/issues/124
Resolves https://github.com/asyncapi/generator-model-sdk/issues/126

Issue #176, opened 2021-05-05
language specific keywords are rendered as property names


jonaslagoni:
#### Describe the bug
There are certain keywords that are reserved. This is different between languages, we need to make sure we dont render such properties as-is.

#### How to Reproduce
For example, in Java, you cannot have a property named `enum`. 

#### Expected behavior
I expected the library to handle any property names.

jonaslagoni:
Go generator did not need this functionality as all properties and types are rendered with pascal case and all reserved keywords are camelcased.

PR #266, opened 2021-06-22
feat: add namingConvention option and pass generator to renderers

jonaslagoni:
**Description**

- Add `namingConvention` option to options: naming for type and single property
- Pass `generator` to `renderers`
- Add `collectionType` option to Java generator to render `List<type>` or `type[]` signature for collections
- Add unit tests
- Update renderers and use `namingConvention` everywhere
- Update docs

**Related issue(s)**
fixes https://github.com/asyncapi/generator-model-sdk/issues/124
fixes https://github.com/asyncapi/generator-model-sdk/issues/126

Hijacked from https://github.com/asyncapi/modelina/pull/129 as agreed with @magicmatatjahu


magicmatatjahu:
I don't know if it's a good idea to make review for my own (in 80-90%) code 😅 @smoya Could you also check PR as co-reviewer?

jonaslagoni:
> I don't know if it's a good idea to make review for my own (in 80-90%) code 😅 @smoya Could you also check PR as co-reviewer?

Gonna need to add some JSDoc to this, gonna request a review once it is ready

jonaslagoni:
Just occurred to me that, because each language might have different naming conventions, that will be called differently, this needs to be in the individual generators, and not in the AbstractGenerator... Example Go properties are called `fields` and with no way of defining that convention, this feature falls through..

src/generators/java/JavaRenderer.ts
```diff
@@ -1,24 +1,50 @@
...
+   * Renders the name of a property based on provided generator option naming convention property callback.
+   * 
+   * @param propertyName 
+   * @param property
+   */
+  nameProperty(propertyName: string | undefined, property?: CommonModel): string {
+    return this.options?.namingConvention?.property 
+      ? this.options.namingConvention.property(propertyName, { model: this.model, inputModel: this.inputModel, property })
+      : propertyName || '';
+  }
```
jonaslagoni, line 46: While at the moment it looks like duplicate code, solving #176 will result in different implementations between renderers.

jonaslagoni:
@magicmatatjahu @smoya okay it is ready to be reviewed now 🙇

src/generators/java/JavaRenderer.ts
```diff
@@ -29,10 +55,16 @@ export abstract class JavaRenderer extends AbstractRenderer<JavaOptions> {
       return 'Object'; // fallback
     }
     if (model.$ref !== undefined) {
-      return FormatHelpers.toPascalCase(model.$ref);
+      return this.nameType(model.$ref, model);
+    }
+    if (
+      TypeHelpers.extractKind(model) === ModelKind.PRIMITIVE ||
+      TypeHelpers.extractKind(model) === ModelKind.ARRAY
+    ) {
```
magicmatatjahu, line 63:
```suggestion
    const kind = TypeHelpers.extractKind(model);
    if (
      kind === ModelKind.PRIMITIVE ||
      kind === ModelKind.ARRAY
    ) {
```

magicmatatjahu (changes requested):
We should also add the `namingConvention` for the Golang generator :)

jonaslagoni:
> We should also add the `namingConvention` for the Golang generator :)

Done, completely forgot about it 😆 Updated docs and Go generator.

src/helpers/NameHelpers.ts
```diff
@@ -29,7 +29,7 @@ export function getUniquePropertyName(rootModel: CommonModel, propertyName: stri
  */
 export type CommonNamingConvention = {
   type?: (name: string | undefined, ctx: { model: CommonModel, inputModel: CommonInputModel }) => string;
-  property?: (test: string | undefined, ctx: { model: CommonModel, inputModel: CommonInputModel, property?: CommonModel }) => string;
+  property?: (name: string | undefined, ctx: { model: CommonModel, inputModel: CommonInputModel, property?: CommonModel }) => string;
```
magicmatatjahu, line 32: Sorry I didn't notice it 😅

magicmatatjahu (approved):
LGTM!

Issue #47, opened 2021-02-11
Implement GoLang generator and renderers


magicmatatjahu:
Implement GoLang generator and renderers like in TS or Java.

Generator should have possibility to define presets for `struct`, `enum` and normal `type` type. 

Additional: create presets for serialisation/converters from json.

PR #267, opened 2021-06-22
feat: add basic struct generation for Go

smoya:
**Description**

This PR adds the very first steps for generating models in *Go*.  It is only supported the generation of *struct* type with simple field types:
- `number` => `float64`
- `integer` => `int`
- `boolean` => `bool`
- `string` => `string`
- `object` => `interface{}`
- `array` => Either `[]<type>` or `[]interface{}`
  - `array` with just one type => `[]<type>`
  - `array` as tuple with only one type => `[]<type>`
  - `array` as tuple with multiple types => `[]interface{}`

Most of the code has been developed during a pairing session videocall that has been recorded:
Link: https://postman.zoom.us/rec/share/IqPkCzFU1LrVod8DtN5XqhRUnRjdlbrPgrMEfaqwyBop8Z6v3AThlhFSC8BAJyU1.D4Q1cSZUYcveTZbJ
Pass: `86?8XEFM`

**Next steps**
From here, after review happens, this can be merged. 
After, we should keep improving the support for generating go code. The most difficult part is generating union types such as `enum` or array of multiple types. We could use https://app.quicktype.io/ as guideline for this development.

**Related issue(s)**
fixes https://github.com/asyncapi/modelina/issues/47


jonaslagoni:
@smoya you have a few issues with some tests 🙂

smoya:
> @smoya you have a few issues with some tests 🙂

Yeah, also fixing linting issues.

smoya:
Sonarcloud fails because we are mainly introducing a new generator, which forced me to do some copy paste for the scaffolding of it. We could perhaps reunify or abstract some logic into shared one but this will also introduces complexity at this moment. 

I definitely believe we can improve code reusability and/or make less contact points when creating new generators. However there is work to do on modelina side.

At this point, shall we make an exception on the Sonarcloud code duplication detector?

cc @magicmatatjahu @jonaslagoni @derberg @fmvilas

jonaslagoni:
> At this point, shall we make an exception on the Sonarcloud code duplication detector?

Accept it for now, but create a new issue to correctly solve this as there are pro/cons for both decisions, and not an easy solution 🙂

src/generators/go/GoGenerator.ts
```diff
@@ -0,0 +1,42 @@
...
+    super('Go', GoGenerator.defaultOptions, options);
+  }
+
+  render(model: CommonModel, inputModel: CommonInputModel): Promise<RenderOutput> {
+    const kind = TypeHelpers.extractKind(model);
+    if (kind === ModelKind.OBJECT) {
+      return this.renderStruct(model, inputModel);
+    }
+    return Promise.resolve(RenderOutput.toRenderOutput({ result: '', dependencies: [] }));
+  }
```
magicmatatjahu, line 34: You can make the function asynchronous using the `async` annotation, and then you don't need call `Promise.resolve`:

```suggestion
  async render(model: CommonModel, inputModel: CommonInputModel): Promise<RenderOutput> {
    const kind = TypeHelpers.extractKind(model);
    if (kind === ModelKind.OBJECT) {
      return this.renderStruct(model, inputModel);
    }
    return RenderOutput.toRenderOutput({ result: '', dependencies: [] });
  }
```

src/generators/go/GoRenderer.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this.renderBlock(content);
+  }
+
+  runFieldPreset(fieldName: string, field: CommonModel): Promise<string> {
+    return this.runPreset('field', { fieldName, field });
+  }
+
+  renderType(model: CommonModel): string {
+    if (model.$ref !== undefined) {
+      return FormatHelpers.toPascalCase(model.$ref, { transform: pascalCaseTransformMerge });
```
magicmatatjahu, line 40: LOL. I don't know about `pascalCaseTransformMerge` option... 🤦🏼

src/generators/go/GoRenderer.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (type === undefined) {
+      return 'interface{}';
+    }
+
+    switch (type) {
+    case 'string':
+      return 'string';
+    case 'integer':
+    case 'number':
+      return 'int';
```
magicmatatjahu, line 60: The `number` should be rendered as `float64` or at least `float32`:

```suggestion
    case 'integer':
      return 'int';
    case 'number':
      return 'float64';
```

src/generators/go/GoRenderer.ts
```diff
@@ -0,0 +1,75 @@
...
+    case 'boolean':
+      return 'bool';
+    case 'object':
+      return 'struct{}';
```
magicmatatjahu, line 64: Object should be rendered as `interface{}`, because we don't know the shape of object.

src/generators/go/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './GoGenerator';
+export { GO_DEFAULT_PRESET } from './GoPreset';
+export type { GoPreset } from './GoPreset';
```
magicmatatjahu, line 3: This is important, as I see in the other parts of PR, some code is still in development, so maybe we should not export at the moment new language? :)

src/generators/go/renderers/StructRenderer.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export class StructRenderer extends GoRenderer {
+  public async defaultSelf(): Promise<string> {
+    const content = [
+      await this.renderFields(),
+      await this.runAdditionalContentPreset()
+    ];
+
+    const formattedName = this.model.$id && FormatHelpers.toPascalCase(this.model.$id, { transform: pascalCaseTransformMerge });
+    return `// ${formattedName} represents a ${formattedName} model.
```
magicmatatjahu, line 19: You should create the helper like `renderComments()` in [TS](https://github.com/asyncapi/modelina/blob/master/src/generators/typescript/TypeScriptRenderer.ts#L85) to render comments :)

src/generators/go/renderers/StructRenderer.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+  field({ fieldName, field, renderer }) {
+    return `${FormatHelpers.toPascalCase(fieldName, { transform: pascalCaseTransformMerge }) } ${ renderer.renderType(field)}`;
+  },
+  getter() {
+    return 'getterFunc';
+  },
+  setter() {
+    return 'setterFunc';
+  },
```
magicmatatjahu, line 41: I have in mind these parts when I wrote in another comments:

> This is important, as I see in the other parts of PR, some code is still in development, so maybe we should not export at the moment new language? :)

So we shouldn't export at the moment the Golang generator, or... please write the implementation of `setter`, `getter`, `ctor` 😄

src/generators/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './AbstractRenderer';
 export * from './java';
 export * from './javascript';
 export * from './typescript';
+export * from './go';
```
magicmatatjahu, line 6: As in above comments.

magicmatatjahu (changes requested):
Awesome work! How you feel when coding in the most beautiful programming language? 😄 I added some comments, only related to rendered Golang types and exports of new language.

src/generators/go/GoRenderer.ts
smoya: Good catch. Fixed!

src/generators/go/renderers/StructRenderer.ts
smoya: But how do I generate comments based on the field name? I'm confused right now.

src/generators/go/renderers/StructRenderer.ts
smoya: I understand that merging code that is not complete and making it available to the final user could lead to some misunderstanding, believing that we support Go in Modelina. However, progressive changes are always better. 
You suggest to skip the export part, but then tests won't work ever as they need to import the Generator, among others. Any smart way of fixing this?

src/generators/go/renderers/StructRenderer.ts
magicmatatjahu: In TS you put content as lines, so in GO you can do this in this same way

```ts
  renderComments(lines: string | string[]): string {
    lines = FormatHelpers.breakLines(lines);
    return lines.map(line => `// ${line}`).join('\n');
  }
```

And then you can write in your code:

```ts
return this.renderComments(`${formattedName} represents a ${formattedName} model.`);
```

src/generators/go/renderers/StructRenderer.ts
magicmatatjahu: You can import them from path where they are:

```ts
import ... from ".../src/generators/go/renderers/StructRenderer.ts"
```

etc.

smoya: Ok, I removed the `export * from './GoGenerator';` from the `index.ts` file so `GoGenerator` is un-exported now.

src/generators/index.ts
smoya: If I do this, the changes I did here are not needed, right? https://github.com/asyncapi/modelina/pull/267#discussion_r659729267

magicmatatjahu: Yes :)

smoya: Oki doki! fixed then :)

src/generators/go/renderers/StructRenderer.ts
smoya: Done!

test/generators/go/GoGenerator.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+}`;
+
+    const inputModel = await generator.process(doc);
+    const model = inputModel.models['_address'];
+
+    let classModel = await generator.renderStruct(model, inputModel);
+    expect(classModel.result).toEqual(expected);
+    expect(classModel.dependencies).toEqual([]);
+
+    classModel = await generator.render(model, inputModel);
```
magicmatatjahu, line 43:
```suggestion
    let structModel = await generator.renderStruct(model, inputModel);
    expect(classModel.result).toEqual(expected);
    expect(classModel.dependencies).toEqual([]);

    structModel = await generator.render(model, inputModel);
```

magicmatatjahu (changes requested):
Good! But you also should write simple tests for presets like here https://github.com/asyncapi/modelina/blob/master/test/generators/typescript/TypeScriptGenerator.spec.ts#L111 :)

smoya:
> Good! But you also should write simple tests for presets like here https://github.com/asyncapi/modelina/blob/master/test/generators/typescript/TypeScriptGenerator.spec.ts#L111 :)

It's done !

magicmatatjahu (approved):
LGTM! Please wait for Jonas :)

jonaslagoni (changes requested):
The actual implementation looks great! 

I only have a couple of stuff we need to add:
- Please update all relevant docs.
    - https://github.com/asyncapi/modelina/blob/master/docs/generators.md#supported-languages
    - Readme file.
- Please add a blackbox test along the rest of the languages here: https://github.com/asyncapi/modelina/blob/master/test/blackbox/Dummy.spec.ts. It is to ensure that everything runs smoothly and can be "compiled" or what ever the language does 😄 
    - https://github.com/asyncapi/modelina/blob/master/docs/development.md#blackbox-testing

smoya:
> Please update all relevant docs.
> * https://github.com/asyncapi/modelina/blob/master/docs/generators.md#supported-languages
> * Readme file.

We do not support Go language yet. It is not even exported so I would not include it in that list.

jonaslagoni:
> We do not support Go language yet. It is not even exported so I would not include it in that list.

Ahh did not see you had a discussion about not exporting the generator 🧐

smoya:
> * Please add a blackbox test along the rest of the languages here: https://github.com/asyncapi/modelina/blob/master/test/blackbox/Dummy.spec.ts. It is to ensure that everything runs smoothly and can be "compiled" or what ever the language does 😄

I'm also asking myself if makes sense to do that at this moment since the generator is not exported. I mean, I can totally do it but I'm not sure if makes sense. The generator has it's own unit tests. Blackbox test would make more sense once we export it since we will be testing the end-to-end scenario.
But open to add the test here if you consider it should be there.

src/generators/go/renderers/StructRenderer.ts
jonaslagoni: Why not just remove the hooks so the Go generator just generates a simple struct with fields, no accessors or constructor? 

I feel like it is wrong to merge something that cannot be used unless there is a specific reason/plan to enable the support 😄

jonaslagoni:
@magicmatatjahu @smoya added a comment to your discussion, feel like it is better to release it as a working generator and export it to be used (even though it is missing accessors/constructor)

src/generators/go/renderers/StructRenderer.ts
smoya: If we do that we would then need to add proper documentation mentioning the limitations and the current state of the generator (pretty early days, limited features).
I'm not against that, In fact I see it can add value to the final user. Some users will find this, generation of structs, useful for their projects.
If we move with that option, I vote for merging this PR, then working on enabling it in another PR so we do not pollute this one and have to review it again. 

Tbh, I vote to make the generator public so users can use it (as @jonaslagoni suggested).

jonaslagoni: > If we do that we would then need to add proper documentation mentioning the limitations and the current state of the generator (pretty early days, limited features).

To be fair to me as long as the repo is < v1 all generators are to be seen as WIP, but you right we should probably start defining the states of the generators 🤔 

> If we move with that option, I vote for merging this PR, then working on enabling it in another PR so we do not pollute this one and have to review it again.

That's fair, that works for me 👍 

@magicmatatjahu any thoughts?

magicmatatjahu: Ok, merge this and work on another :)

jonaslagoni (approved):
LGTM! 🎉

jonaslagoni:
@smoya do you mind creating follow-up issues with the changes to the go generator?

jonaslagoni:
@all-contributors please add @smoya for code, test

PR #268, opened 2021-06-23
feat: render more accurate array type for Java output

### LINKED ISSUE #261: Java generator should take additionalItems into consideration
#### Reason/Context
Even though we interpret additionalItems (#224) we do not take it into consideration when rendering the Java types for arrays. 

See https://github.com/asyncapi/modelina/issues/224#issuecomment-863200247 for further details on how it should be rendered.

jonaslagoni:
**Description**
This PR renders array type more closely to the underlying structure.

If tuples are encountered the tuple models are combined to ensure that the following CommonModel tuples:
```
items: [
  {
    type: 'string'
  },
  {
    type: 'string'
  }
]
```
are rendered as `String[]` and not `Object[]`.

Furthermore `additionalItems` are taken into account when rendering tuple type to ensure that the appropriate array type is rendered.

**Related issue(s)**
fixes #261

### DIFF
--- a/src/generators/java/JavaRenderer.ts
+++ b/src/generators/java/JavaRenderer.ts
@@ -69,7 +69,15 @@ export abstract class JavaRenderer extends AbstractRenderer<JavaOptions> {
     case 'binary':
       return 'byte[]';
     case 'array': {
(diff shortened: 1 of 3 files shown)


magicmatatjahu (approved):
LGTM

PR #269, opened 2021-06-23
fix: ensure reserved keywords can never be rendered as properties for Java class

### LINKED ISSUE #176: language specific keywords are rendered as property names

jonaslagoni:
**Description**
This PR ensures that reserved keywords can never be rendered for the Java class.

I actually struggled quite a bit with how this functionality should work in cohesion with the name convention option. I could not come up with any other way of solving this problem other than this way. In case you see another please let me know 🙏 

**Related issue(s)**
partly solves #176


jonaslagoni:
Blocked by https://github.com/asyncapi/modelina/pull/266

jonaslagoni:
Once we agree that this approach is as expected I will replicate it across all generators.

docs/generators.md
```diff
@@ -24,6 +24,7 @@ Default options contain:
 | Option | Type | Description | Default value |
 |---|---|---|---|
 | `indentation` | Object | Options for indentation. | - |
+| `reservedPropertyWord` | String | Word or character to append to property name when reserved keywords or duplicate property names are encountered. | 'reserved' |
```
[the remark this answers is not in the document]
jonaslagoni: Hmm, yea you are right 🤔 Good suggestion 👍

src/models/Preset.ts
```diff
@@ -24,7 +24,8 @@ export enum PropertyType {
 export interface PropertyArgs {
   propertyName: string;
   property: CommonModel;
-  type: PropertyType
+  type: PropertyType;
+  modelPropertyName: string;
```
magicmatatjahu, line 28: For what you need `modelPropertyName`? As I see in the implementation, it's the same as `propertyName` and also you don't use it anywhere.

src/helpers/NameHelpers.ts
```diff
@@ -19,29 +19,47 @@ export enum DefaultPropertyNames {
...
     return FormatHelpers.toPascalCase(name);
   },
-  property: (name: string | undefined) => {
+  property: (name, ctx) => {
     if (!name) {return '';}
+    if (ctx.isReservedKeyword) { 
+      // If name is considered reserved, make sure we rename it appropriately
+      // and make sure no clashes occur.
+      name = FormatHelpers.toCamelCase(`reserved_${name}`);
+      if (Object.keys(ctx.model.properties || {}).includes(name)) {
```
magicmatatjahu, line 58: Good catch! 💪🏼

magicmatatjahu (changes requested):
Good work! Please refer to my comments, and also please add missed TS tests :)

jonaslagoni:
> Good work! Please refer to my comments, and also please add missed TS tests :)

I just added the keyword list, next PR is gonna be to add the implementation of using the list the same way Java has been adapted. Just didnt want to remove it now that I had added it 😅

src/models/Preset.ts
jonaslagoni: Good catch, left over from old implementation, removing!

magicmatatjahu:
@jonaslagoni Ok, remember about promise 😄

magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks 🙇

Issue #154, opened 2021-04-26
Render pattern properties


jonaslagoni:
#### Reason/Context
It should be possible to render pattern properties.

Example: https://github.com/asyncapi/generator-model-sdk/issues/43#issuecomment-782104228

Blocked by https://github.com/asyncapi/generator-model-sdk/pull/153

Issue #262, opened 2021-06-18
Add helper functions to un/marshal classes 


jonaslagoni:
#### Reason/Context
We need some helper functions to be able to marshal the data model instances to JSON and unmarshal bytes to instances of the data model.

#### Description
The reason we need to write this our self is that we add properties to data models which are technically not properties directly, an example here is `additionalProperties`, as it is rendered as a property in the data model, but when you want to convert an instance of the data model to bytes or JSON for a payload, it should not be used as is but unwrapped.

magicmatatjahu:
@jonaslagoni I'm not on the topic, is this issue still relevant?

jonaslagoni:
Still relevant for Go, JS and Java (maybe).

PR #270, opened 2021-06-23
feat: render patternProperties for TS class and interface

jonaslagoni:
**Description**
This PR renders a pattern properties property for both class and interface presets.

**Related issue(s)**
Partly solves #154


test/blackbox/docs/dummy.json
```diff
@@ -47,6 +47,11 @@
       "$ref": "#/definitions/dummyObject"
     }
   },
+  "patternProperties": {
+    "S_pattern": {
+      "type": "string"
+    }
```
magicmatatjahu, line 53: Please try to render the pattern like `^S(.?*)pattern&`, because `S_pattern` is a simple "regex".

magicmatatjahu (changes requested):

test/blackbox/docs/dummy.json
jonaslagoni: Still renders it as `private _sPatternPatternProperty?: Map<String, string>;`

src/helpers/NameHelpers.ts
```diff
@@ -4,7 +4,8 @@ import { CommonModel } from 'models';
  * Default property names for different aspects of the common model
  */
 export enum DefaultPropertyNames {
-  additionalProperties = 'additionalProperties'
+  additionalProperties = 'additionalProperties',
+  patternProperty = 'PatternProperty'
```
magicmatatjahu, line 8:
```suggestion
  patternProperty = 'patternProperties'
```

Because we save pattern properties as collection, not single one :)

magicmatatjahu (changes requested):
One thing, but overall LGTM. We must remember about proper serialization for pattern properties and additional properties in https://github.com/asyncapi/modelina/issues/262 issue. At the moment we save mentioned keywords to collection but in JSON everything is flattened.

src/helpers/NameHelpers.ts
jonaslagoni: That's true, but each pattern is rendered as property here, meaning the default property name for pattern properties are `{pattern}{DefaultPropertyNames.patternProperty}`. 

To me this makes more sense then using plural 🤔?

jonaslagoni:
> At the moment we save mentioned keywords to collection but in JSON everything is flattened.

Yep, that's right. Gonna add it to the description once this is merged.

src/helpers/NameHelpers.ts
jonaslagoni: I guess you are right, changing 🧐

magicmatatjahu (approved):
LGTM! :)

PR #273, opened 2021-06-24
feat: support AsyncAPI 2.1.0

jonaslagoni:
**Description**
This PR ensures that we support specific versions of the AsyncAPI spec.


jonaslagoni:
Blocked until the release of the spec, but I think we can review it now so we are ready to merge.

src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -145,14 +145,30 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
...
+   * @param input 
+   */
+  tryGetVersionOfDocument(input: Record<string, any>) : string | undefined {
+    if (AsyncAPIInputProcessor.isFromParser(input)) {
+      return input.version();
+    }
+    if (input.asyncapi !== undefined) {
+      return input.asyncapi;
+    }
+    return undefined;
```
magicmatatjahu, line 169:
```suggestion
    if (AsyncAPIInputProcessor.isFromParser(input)) {
      return input.version();
    }
    return input && input.asyncapi;
```

src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -145,14 +145,30 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
...
-    //Check if we got a parsed document from out parser
-    //Check if we just got provided a pure object
-    if (AsyncAPIInputProcessor.isFromParser(input) || input.asyncapi !== undefined) {
+    const version = this.tryGetVersionOfDocument(input);
+    if (version === undefined) { return false; }
+    const supportedVersions = ['2.0.0', '2.1.0'];
+    if (supportedVersions.includes(version)) {
       return true;
     }
     return false;
```
magicmatatjahu, line 154:
```suggestion
    const version = this.tryGetVersionOfDocument(input);
    return supportedVersions.includes(version);
```

and 

```js
const supportedVersions = ['2.0.0', '2.1.0'];
```

should be treated as `private static` value for class or `const` outside the class :)

magicmatatjahu (changes requested):

src/processors/AsyncAPIInputProcessor.ts
jonaslagoni: Good call 👍

magicmatatjahu (approved):

PR #281, opened 2021-06-30
feat: add Dockerfile

smoya:
**Description**

This PR adds a `Dockerfile` to this project so developers can run all test suites without configuring any other dependency locally. Specially useful for blackbox test suite. 
In fact, you can run any other command you want. Once we have a CLI, we could publish the Docker image as well so people can start using it directly from a Docker registry.

Docs has been updated.


jonaslagoni (commented):
@smoya I think we should make sure the docker image actually works with the CI, maybe we can adapt the workflow from the generator to build and test it? 🤔 - https://github.com/asyncapi/generator/blob/master/.github/workflows/release-docker.yml

smoya:
> @smoya I think we should make sure the docker image actually works with the CI, maybe we can adapt the workflow from the generator to build and test it? 🤔 - https://github.com/asyncapi/generator/blob/master/.github/workflows/release-docker.yml

Good point. However, the main test suite runs on a shared workflow: https://github.com/asyncapi/modelina/blob/master/.github/workflows/if-nodejs-pr-testing.yml (comes from our .github repository).

What about making https://github.com/asyncapi/modelina/blob/master/.github/workflows/blackbox-testing.yml use the docker image instead?

Regarding the release of the Docker image, I would not include it yet. Once we have a reason to provide public Docker images  (such as having a CLI tool) we can add that step.

jonaslagoni:
> Regarding the release of the Docker image, I would not include it yet. Once we have a reason to provide public Docker images (such as having a CLI tool) we can add that step.

Yea that's my point, remove the release step from the workflow so it only builds the docker image and runs the tests afterward 👍 

> What about making https://github.com/asyncapi/modelina/blob/master/.github/workflows/blackbox-testing.yml use the docker image instead?

Good suggestion! Lets try that, so it builds the docker image and then runs the tests inside the docker image 👍

smoya:
> Good suggestion! Lets try that, so it builds the docker image and then runs the tests inside the docker image

I'm gonna add this in a follow up issue as it is not that trivial as it sounds. We would need to reuse docker images somehow in order to not penalize CI time, among others.

package.json
```diff
@@ -62,9 +62,12 @@
     "build:cjs": "tsc",
     "build:esm": "tsc --project tsconfig.json --module ESNext --outDir ./lib/esm",
     "build:types": "tsc --project tsconfig.json --declaration --emitDeclarationOnly --declarationMap --outDir ./lib/types",
+    "build:docker": "docker build -q -t asyncapi/modelina .",
     "test": "cross-env CI=true jest --coverage --testPathIgnorePatterns ./test/blackbox",
     "test:blackbox": "cross-env CI=true jest ./test/blackbox",
     "test:watch": "jest --watch",
+    "test:docker": "npm run build:docker && docker run asyncapi/modelina npm run test",
```
jonaslagoni, line 69: Would `docker:test` not make more sense? As we are testing using docker, we don't test docker 🤔

package.json
```diff
@@ -62,9 +62,12 @@
...
+    "test:docker:blackbox": "npm run build:docker && docker run asyncapi/modelina npm run test:blackbox",
```
jonaslagoni, line 70: Same here as above

jonaslagoni (changes requested):
Does the script naming make sense? 🤔

package.json
jonaslagoni: hmm... I guess it might be better to stay as is, in terms of how the rest of the scripts are 🤔

jonaslagoni: Thoughts?

smoya: Yeah, I'm happy with both naming patterns. At the end you could also read this as `test:*` is all related to testing. But Agree prefixing with `docker:` instead could make more sense. Changing!

smoya: Done!

package.json
smoya: Fixed

jonaslagoni (approved):
LGTM 👍

smoya:
Test coverage failed because of a random weird error. I’m gonna skip it as I did not add any test or change it (no work on code)

derberg:
yo, just a comment that this probably should not be a `feat` triggering release as release artifact did not change, and you are not publishing docker image to dockerhub

smoya:
> yo, just a comment that this probably should not be a `feat` triggering release as release artifact did not change, and you are not publishing docker image to dockerhub

Good catch. My fault, now it's a bit late unfortunately. Will take care next time.

jonaslagoni:
Yea agreed you are right @derberg, will keep it in mind 👍

derberg:
that is fine folks, I left my comment as a future remark to always keep in mind to look at the title, especially if one day we find time to automate merge process by the bot

PR #283, opened 2021-06-30
feat: enable go generator

smoya:
This enables the go generator, limited in terms of feature by now.
It also includes its blackbox test but also a small refactor replacing
current blackbox `AsyncAPI.spec.ts` tests, that are identical between languages,
with a table-driven one.


test/blackbox/AsyncAPI.spec.ts
```diff
@@ -1,29 +1,19 @@
...
-    const generator = new JavaScriptGenerator();
-    const generatedContent = await generator.generate(inputSchema);
-    expect(generatedContent).not.toBeUndefined();
-    expect(generatedContent.length).toBeGreaterThan(1);
+  describe.each([
+    ['TypeScript', new TypeScriptGenerator()],
+    ['Java', new JavaGenerator()],
+    ['Javascript', new JavaScriptGenerator()],
+    ['Go', new GoGenerator()],
+  ])('code generated in %s', (_, generator) => {
```
jonaslagoni, line 10: Wow, now thats a cool feature!

jonaslagoni (changes requested):
We need to remove the unused hooks here - https://github.com/asyncapi/modelina/blob/018ba36f44fc7484affd372c70f06bbd60d2c702/src/generators/go/renderers/StructRenderer.ts#L37

jonaslagoni:
@smoya do you mind adding `src/generators/javascript/JavaScriptRenderer.ts,src/generators/java/renderers/ClassRenderer.ts,src/generators/typescript/TypeScriptRenderer.ts` to the following file: https://github.com/asyncapi/modelina/blob/master/sonar-project.properties 

This makes sure that Sonar starts ignoring these lines.

jonaslagoni (commented):
You also need to update https://github.com/asyncapi/modelina/blob/master/docs/generators.md#supported-languages 🙂 

Generally, I would ask you to include the tests in the Dummy blackbox test and provide an update to the Dockerfile so it can be tested, do you have time for this or should it be added in a follow-up issue?

smoya:
> You also need to update https://github.com/asyncapi/modelina/blob/master/docs/generators.md#supported-languages 🙂
> 
> Generally, I would ask you to include the tests in the Dummy blackbox test and provide an update to the Dockerfile so it can be tested, do you have time for this or should it be added in a follow-up issue?

I am confused. Isn't that already in my PR?

jonaslagoni:
I can see why that is confusing 🤣

So we have two black box tests, we have the asyncapi test which is more of a guideline of what we want to be able to generate eventually. But the JSON Schema is so complex that it's hard to truly test the output. Which is why we created the dummy https://github.com/asyncapi/modelina/blob/master/test/blackbox/Dummy.spec.ts blackbox test. Which is easier to adapt and test output with 😄

smoya:
> `src/generators/javascript/JavaScriptRenderer.ts,src/generators/java/renderers/ClassRenderer.ts,src/generators/typescript/TypeScriptRenderer.ts` to the following file: https://github.com/asyncapi/modelina/blob/master/sonar-project.properties

I guess we should also add the go one.

jonaslagoni (dismissed):
LGTM, awesome @smoya  🎉

src/generators/go/GoGenerator.ts
```diff
@@ -27,9 +29,17 @@ export class GoGenerator extends AbstractGenerator<GoOptions> {
...
-    if (kind === ModelKind.OBJECT) {
+    switch (kind) {
+    case ModelKind.OBJECT: {
       return this.renderStruct(model, inputModel);
     }
+    case ModelKind.ENUM: {
+      const typeName = model.$id && FormatHelpers.toPascalCase(model.$id, { transform: pascalCaseTransformMerge });
+      const result = `// ${typeName} represents an enum\ntype ${typeName} interface{}`;
+      return Promise.resolve(RenderOutput.toRenderOutput({ result, dependencies: [] }));
+    }
```
magicmatatjahu, line 40: It should be written as separate function. As I see you wanna make then the "true" enum in golang, not using the `interface{}`, yes?

magicmatatjahu: Also for this 

```ts
const result = `// ${typeName} represents an enum\ntype ${typeName} interface{}`;
```

Please use the `renderComments` and `renderBlock` functions.

smoya: No I don't wanna make the enums to look right. This is not the final version of an enum as enums in go does not exist. There will need more work to make that happen.

For simplicity, I did not create a EnumRenderer since it is not rendering real enums.

magicmatatjahu: I know that there isn't real enum but you can create the "fake" enums like for string:

```go
type LeaveType string

const(
    AnnualLeave LeaveType = "AnnualLeave"
    Sick = "Sick"
    BankHoliday = "BankHoliday"
    Other = "Other"
)
```

I think that we go in this way.

smoya: The point is that this PR is not meant for that. Its purpose is to enable the generator, and for that, I had to print some type on enums when generating the structs.

smoya: There will be more issues to come for enabling all features in the Go Generator.

magicmatatjahu: OK, but really, please move the logic to separate function, not Renderer, and use the `renderComments` and `renderBlock` functions.

smoya: I ended up creating the `EnumRenderer` and some tests around.

jonaslagoni:
@smoya you have a sonarcloud code smell 🙂

src/generators/go/renderers/EnumRenderer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  renderCommentForEnumType(name: string, type: string): string {
+    const globalType = type === 'interface{}' ? 'mixed types' : type;
+    return this.renderComments(`${name} represents an enum of ${globalType}.`);
+  }
+
+  renderComments(lines: string | string[]): string {
+    lines = FormatHelpers.breakLines(lines);
+    return lines.map(line => `// ${line}`).join('\n');
+  }
```
magicmatatjahu, line 37: This function should be written in `GoRenderer` class.

magicmatatjahu (commented):
One thing :)

src/generators/go/renderers/EnumRenderer.ts
smoya: done

magicmatatjahu (approved):
LGTM!

smoya:
> @smoya you have a sonarcloud code smell 🙂

@jonaslagoni fixed!

jonaslagoni (approved):
🎉

PR #290, opened 2021-07-07
fix: publish esm modules and TS types

magicmatatjahu:
**Description**

As in title: publish `esm` modules and TS types - update `prepublishOnly` script.


smoya (approved):
LGTM! 🚀

PR #292, opened 2021-07-07
fix: wrong import paths

smoya:
*Description**

Running the example I found that compiling TS gave the following error:

```
node_modules/@asyncapi/modelina/lib/types/generators/AbstractGenerator.d.ts:3:30 - error TS2307: Cannot find module 'models/RenderOutput' or its corresponding type declarations.
3 import { RenderOutput } from 'models/RenderOutput';
```

I noticed the import path was wrong in this case. Fixed that one and a couple more.


magicmatatjahu (approved):
LGTM

jonaslagoni:
Damn, I am super surprised that the tests did not catch this 🤔 

@magicmatatjahu do you have any idea why?

PR #300, opened 2021-07-07
fix: remove pascal-case dependency in favour of change-case

derberg:
- we should not use dependencies of our dependencies as this is a bad practice, they remove the package and we have issues
- after adding dependency there were no updates to package-lock 🤷🏼 like it was already there but manually removed from package.json 🤷🏼


derberg:
@jonaslagoni I think you have to look again into duplications setup on sonarcube, I see here 50% in my PR and I think an hour ago it was 7% on Sergio's PR which is strange

jonaslagoni (changes requested):
Instead of adding another dependency, we can use the dependency directly through  `change-case`. 

I.e. changing https://github.com/asyncapi/modelina/blob/65ead0e83933414ced451ba758b63834df0faf03/src/generators/go/GoGenerator.ts#L11 to `import { pascalCaseTransformMerge } from 'change-case';`

jonaslagoni:
> @jonaslagoni I think you have to look again into duplications setup on sonarcube, I see here 50% in my PR and I think an hour ago it was 7% on Sergio's PR which is strange

I am starting to hate sonarcloud with their duplicate code setup. I am gonna see if I cant pull myself together and fix it someday...

derberg:
@jonaslagoni did you try with their official support? I tried on twitter but no luck so far

jonaslagoni:
> @jonaslagoni did you try with their official support? I tried on twitter but no luck so far

I have not, no, but I will reach out to them whenever I take this issue up again. That or simply fixing the duplications so I dont have to deal with it 🤔 hmm

derberg:
this change-case is strange, like lodash, you take all or just an individual package.

anyway removed pascal, and sonar started working, magic 😄

sonar-project.properties
```diff
@@ -1,2 +1,2 @@
 # Disable specific duplicate code since it would introduce more complexity to reduce it.
-sonar.cpd.exclusions=src/generators/javascript/renderers/ClassRenderer.ts,src/generators/java/renderers/ClassRenderer.ts,src/generators/typescript/renderers/ClassRenderer.ts,src/generators/javascript/JavaScriptRenderer.ts,src/generators/java/renderers/ClassRenderer.ts,src/generators/typescript/TypeScriptRenderer.ts,src/generators/typescript/renderers/EnumRenderer.ts,src/generators/java/renderers/EnumRenderer.ts
```
jonaslagoni, line 2: Why did you remove the path? 🤔

jonaslagoni:
> this change-case is strange, like lodash, you take all or just an individual package.
> 
> anyway removed pascal, and sonar started working, magic 😄

I don't like magic 😆

sonar-project.properties
derberg: I explained in description that there was a duplicate entry

derberg: actually I don't think I did 😄

jonaslagoni: Ahh!

jonaslagoni (approved):
👍

jonaslagoni:
@all-contributors please add @derberg for bug, code

Issue #299, opened 2021-07-07
Add json tags to Go generated struct


smoya:
#### Reason/Context
Current generated Go structs does not have json tags. Those are useful for encoding/decoding JSON.

#### Description
**Currently**,  for a given doc like:

```js
const doc = {
  $id: "Address",
  type: "object",
  properties: {
    street_name:    { type: "string" },
    city:           { type: "string", description: "City description" },
    house_number:   { type: "number" },
    marriage:       { type: "boolean", description: "Status if marriage live in given house" },
    pet_names:      { type: "array", items: { type: "string" } },
    state:          { type: "string", enum: ["Texas", "Alabama", "California", "other"] },
  },
  required: ["street_name", "city", "state", "house_number", "state"],
};
```

The generator generates the following struct:

```go
// Address represents a Address model.
type Address struct {
  StreetName string
  City string
  HouseNumber float64
  Marriage bool
  PetNames []string
  State *State
}
// State represents an enum of string.
type State string
```

After adding Json tags would look like:

```go
// Address represents a Address model.
type Address struct {
  StreetName string `json:"street_name"`
  City string `json:"city"`
  HouseNumber float64 `json:"house_number"`
  Marriage bool `json:"marriage"`
  PetNames []string `json:"pet_names"`
  State *State `json:"state"`
}
// State represents an enum of string.
type State string
```

#### Related issue

https://github.com/asyncapi/modelina/issues/262

magicmatatjahu:
I think that it's still important? @smoya :)

H3rby7:
Any progress on this?
Without this, it renders the code generation for different languages kind of useless as for example the current GOLANG structs are not in sync with their TS counterparts

For ref: Tackled again by https://github.com/asyncapi/modelina/pull/867, but stale?

jonaslagoni:
Its still definitely needed... I just dont have the bandwidth to finish the PR, feel free to take it over, happy to help facilitate this 👍

H3rby7:
It just looks like a lot to me, judging from the amount of code (changes) and I am not sure exactly where to start with it. As far as a (my) simple use case goes it would be a sufficient start to have the json tags simply be the lowerCamelCase variant of the field name in the struct to give us compability to other generated sources

jonaslagoni:
@H3rby7 you can always change the naming rules: https://github.com/asyncapi/modelina/blob/master/docs/advanced.md#changing-the-constrain-rules

```
const generator = new GoGenerator({
[...]

PR #301, opened 2021-07-07
feat: render pattern properties and additionalProperties as fields for Go struct

jonaslagoni:
**Description**
This PR enables the rendering of `patternProperties` as well as `additionalProperties` for Go struct.

**Related issue(s)**
Related to #154


magicmatatjahu (commented):
Overall good, but please also add one tests for patternProperties, like in other languages :)

jonaslagoni:
> Overall good, but please also add one tests for patternProperties, like in other languages :)

Completely forgot about it, added 👍

magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks 🥂

magicmatatjahu:
@smoya I don't remember some features of Golang tag, but probably If we put the additionalProperties and patternProperties as map we can have problem with this issue, am I right? https://github.com/asyncapi/modelina/issues/299

smoya:
Before understanding the expectation around `additionalProperties` and `patternProperties` here, I would like to better understand what's the mission of Modelina. 

We've said that Modelina generates models from JSON Schema (among others). What is not clear enough for me is the main use cases. 
I see one use case is to use it inside our generators. But I also saw Modelina as a tool for generating domain model that users can use in their applications, for example, with the purpose of encoding/decoding a message to a model, etc. Correct me if I'm wrong so most of the following speech here should then be skipped.

In the second case, and talking, in particular, about the Go model generation, this is not gonna be possible that easy. As stated in https://github.com/asyncapi/modelina/issues/299, Go structs need some special tags (as @magicmatatjahu mentioned in the comment above) to be able to encode/decode (marshal/unmarshal) JSON, otherwise manual encoding/decoding should happen. It is also the way to go with any other kind of encoder/decoder such as Yaml, XML, etc.

In particular, Go [JSON marshaler/unmarshaler](https://golang.org/pkg/encoding/json/) only maps fields that are explicitly set in those tags. Any other fields are **discarded** otherwise. There are some other alternatives that can be used but that won't be then as standard, like using https://github.com/mitchellh/mapstructure which keeps all fields that were not mapped in a `map[string]interface{}` field so you can access later (This is how we will be doing mostly for the parsing in the event-gateway).

Having said that, I'm answering the following question: 

> @smoya I don't remember some features of Golang tag, but probably If we put the additionalProperties and patternProperties as map we can have problem with this issue, am I right? #299

We can generate whatever model we want, but if the mission is to use those for encoding/decoding, at least JSON, that won't work unless we either force users to go with something like https://github.com/mitchellh/mapstructure or rather provide our own encoder/decoder implementation.

jonaslagoni:
Thanks for the feedback @smoya 👍 

> We've said that Modelina generates models from JSON Schema (among others). What is not clear enough for me is the main use cases.

To me there is not a main use-case, it can be used for both `domain model´ and also just for the pure class definition. 

For the `domain model` use-case, it is expected that we are adding a custom [helper functions to un/marshal data models](https://github.com/asyncapi/modelina/issues/262), it has just not been added yet 😄 

I suspect that https://github.com/asyncapi/modelina/issues/299 will simply not be sufficient, because of the nature of the data models where properties do not map 1 to 1 with the JSON data. I.e. `additionalProperties` should be unwrapped and not be used as-is. But it is something we will see once we start those tasks 🙂 

But yea, this issue simply allows the user to specify the "correct" data, i.e. can set additional properties, as the nature of JSON Schema allows.

Hope that clarified it @smoya 🙂

smoya:
> I suspect that #299 will simply not be sufficient, because of the nature of the data models where properties do not map 1 to 1 with the JSON data. I.e. `additionalProperties` should be unwrapped and not be used as-is. But it is something we will see once we start those tasks 🙂

That's right. And as you mentioned it will be covered by https://github.com/asyncapi/modelina/issues/262 so as far as I see, this PR looks good to me then 👍 

Thanks for clarifying @jonaslagoni 🚀

smoya (approved):
https://github.com/asyncapi/modelina/pull/301#issuecomment-877132867

PR #311, opened 2021-07-28
fix: ensure reserved keywords can never be rendered for TS output

jonaslagoni:
**Description**
This PR ensures that reserved keywords cannot be rendered for specific parts of class and interface output. It might seem a bit weird, but apparently, TS is not as strict as Java. Only parts of the syntax need to be protected.

**Related issue(s)**
related to #176

### DIFF
--- a/src/generators/typescript/TypeScriptRenderer.ts
+++ b/src/generators/typescript/TypeScriptRenderer.ts
@@ -108,9 +108,9 @@ export abstract class TypeScriptRenderer extends AbstractRenderer<TypeScriptOpti
    * @param propertyName 
    * @param property
    */
-  nameProperty(propertyName: string | undefined, property?: CommonModel, checkReserved = true): string {
+  nameProperty(propertyName: string | undefined, property?: CommonModel): string {
     return this.options?.namingConvention?.property 
-      ? this.options.namingConvention.property(propertyName, { model: this.model, inputModel: this.inputModel, property, isReservedKeyword: checkReserved ? TypeScriptRenderer.isReservedTypeScriptKeyword(`${propertyName}`) : false })
+      ? this.options.namingConvention.property(propertyName, { model: this.model, inputModel: this.inputModel, property, isReservedKeyword: TypeScriptRenderer.isReservedTypeScriptKeyword(`${propertyName}`) })
       : propertyName || '';
   }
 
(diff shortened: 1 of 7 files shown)


test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -5,6 +5,67 @@ describe('TypeScriptGenerator', () => {
...
+  constructor(input: {
+    reservedReservedEnum?: string,
+    reservedEnum?: string,
+  }) {
+    this._enum = input.reservedReservedEnum;
+    this._reservedEnum = input.reservedEnum;
+  }
+
+  get enum(): string | undefined { return this._enum; }
+  set enum(reservedReservedEnum: string | undefined) { this._enum = reservedReservedEnum; }
```
jonaslagoni, line 53: I wonder if we should use this instead? 🤔 
```
 set enum(_enum: string | undefined) { this._enum = _enum; }
```

src/generators/typescript/TypeScriptRenderer.ts
```diff
@@ -209,8 +212,8 @@ ${lines.map(line => ` * ${line}`).join('\n')}
     return this.renderBlock(content);
   }
 
-  renderProperty(propertyName: string, property: CommonModel, type: PropertyType = PropertyType.property): string {
-    const formattedPropertyName = this.nameProperty(propertyName, property);
+  renderProperty(propertyName: string, property: CommonModel, type: PropertyType = PropertyType.property, checkReserved = true): string {
```
magicmatatjahu, line 215: I think that we should go with strict checking of the reserved keyword, even if TS allows it.

test/generators/typescript/TypeScriptGenerator.spec.ts
magicmatatjahu: Check my comment below :)

magicmatatjahu (changes requested):

src/generators/typescript/TypeScriptRenderer.ts
jonaslagoni: That is also a possibility, it will make the implementation easier to understand as well, I am just not sure whether the end-user will find it weird 🤔

magicmatatjahu: I mean, in the JS you don't have such keywords like `enum`, `interface` etc, in TS you have - in some scope - so I opt to make it strict :) We can also add options to the TS generator if generator should be strict in this area, or not.

magicmatatjahu: such an option can be added in another PR.

jonaslagoni: If the use-case ever comes up, it can of course be added then. Gonna change to a strict no keyword usage across both presets.

magicmatatjahu (approved):
LGTM

PR #313, opened 2021-07-28
fix: ensure reserved keywords can never be rendered for JS output

### LINKED ISSUE #176: language specific keywords are rendered as property names

jonaslagoni:
**Description**
This PR ensures that reserved keywords cannot be rendered for JS class as well as restricting TS outputs for the same keywords.

**Related issue(s)**
fixes #176


src/generators/javascript/JavaScriptRenderer.ts
```diff
@@ -3,7 +3,201 @@ import { JavaScriptGenerator, JavaScriptOptions } from './JavaScriptGenerator';
...
+  'setTimeout',
+  'status', 
+  'submit', 
+  'taint', 
+  'text',
+  'textarea', 
+  'top', 
+  'unescape', 
+  'untaint',
+  'window'
```
magicmatatjahu, line 199: These same we should use also in TS. Also please move the all arrays (also for other langs) to the separate file like `constants.ts`:
- generators/typescript/constant.ts
- generators/java/constant.ts

etc.

test/generators/javascript/JavaScriptGenerator.spec.ts
```diff
@@ -6,6 +6,41 @@ describe('JavaScriptGenerator', () => {
...
+        reservedEnum: { type: 'string' }
+      },
+      additionalProperties: false
+    };
+    const expected = `class Address {
+  reservedReservedEnum;
+  reservedEnum;
+
+  constructor(input) {
+    this.reservedReservedEnum = input.reservedReservedEnum;
```
magicmatatjahu, line 24: `enum` isn't keyword in JS, also not `interface` or `type` :)

magicmatatjahu (changes requested):

test/generators/javascript/JavaScriptGenerator.spec.ts
jonaslagoni: Based on https://www.w3schools.com/js/js_reserved.asp they are?

src/generators/javascript/JavaScriptRenderer.ts
jonaslagoni: > These same we should use also in TS.
We should? Thought the transpile process would ensure compatibility there?

test/generators/javascript/JavaScriptGenerator.spec.ts
magicmatatjahu: LOL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#future_reserved_keywords I don't know about it. Ok, leave it :)

src/generators/javascript/JavaScriptRenderer.ts
magicmatatjahu: Nope, for example TS `window` will be `window` in transpiled JS, so we should also treat the below keywords as reserved for TS.

jonaslagoni: Huh, fair enough 😄 Changed

magicmatatjahu (approved):
LGTM

PR #315, opened 2021-08-09
feat: un/marshal functions for TypeScript class

jonaslagoni:
**Description**
This PR introduces a new common preset for TS, which renders un/marshal functions to transform JSON -> class -> JSON. This is important as the class cannot be automatically converted as we have properties that need to be unwrapped.

**Related issue(s)**
Related to https://github.com/asyncapi/modelina/issues/262


src/generators/typescript/presets/CommonPreset.ts
```diff
@@ -0,0 +1,145 @@
...
+ * @implements {TypeScriptPreset}
+ */
+export const TS_COMMON_PRESET: TypeScriptPreset = {
+  class: {
+    additionalContent({ renderer, model, content, options }) {
+      options = options || {};
+      const blocks: string[] = [];
+      
+      if (options.marshal === undefined || options.marshal === true) {blocks.push(renderMarshal({ renderer, model }));}
+      if (options.unmarshal === undefined || options.unmarshal === true) {blocks.push(renderUnmarshal({ renderer, model }));}
```
jonaslagoni, line 140: Should the two functions be disabled by default?

src/generators/typescript/presets/CommonPreset.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Render `marshal` function based on model
+ */
+function renderMarshal({ renderer, model }: {
+  renderer: TypeScriptRenderer,
+  model: CommonModel,
+}): string {
+  const properties = model.properties || {};
+  const propertyKeys = [...Object.entries(properties)];
+  const marshalProperties = propertyKeys.map(([prop, propModel]) => {
+    const camelCasedProp = FormatHelpers.toCamelCase(prop);
```
magicmatatjahu, line 22: You shouldn't use the `FormatHelpers.toCamelCase` function, because someone can pass another format casing in the `renderer.nameProperty` function - use this last.

src/generators/typescript/presets/CommonPreset.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Render `unmarshal` function based on model
+ */
+function renderUnmarshal({ renderer, model }: {
+  renderer: TypeScriptRenderer,
+  model: CommonModel,
+}): string {
+  const properties = model.properties || {};
+  const propertyKeys = [...Object.entries(properties)];
+  const unmarshalProperties = propertyKeys.map(([prop, propModel]) => {
+    const camelCasedProp = FormatHelpers.toCamelCase(prop);
```
magicmatatjahu, line 81: You shouldn't use the `FormatHelpers.toCamelCase` function, because someone can pass another format casing in the `renderer.nameProperty` function - use this last.

src/generators/typescript/presets/CommonPreset.ts
magicmatatjahu: This function should be disabled by default and I think that `marshal` and `unmarshal` should be applied together, not with different options, so `options.marshalling = true` should render these two functions.

test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+        if (obj.stringProp !== undefined) {
+          instance.stringProp = obj.stringProp;
+        }
+    
+        //Not part of core properties
+        for (const [key, value] of Object.entries(obj).filter((([key,]) => {return !["stringProp"].includes(key);}))) {
+          //Check all pattern properties
+        
+    
+          if (instance.additionalProperties === undefined) {instance.additionalProperties = new Map();}
```
magicmatatjahu, line 92: This should be done outside the for, the best before `for`

magicmatatjahu: Also for `petternProperties`

src/generators/typescript/presets/CommonPreset.ts
magicmatatjahu: https://github.com/asyncapi/modelina/blob/master/src/generators/typescript/TypeScriptRenderer.ts#L150

test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    
+        if(this.additionalProperties !== undefined) { 
+          for (const [key, value] of this.additionalProperties.entries()) {
+            //Only render additionalProperties which are not already a property
+            if(Object.keys(this).includes(String(key))) continue;
+            json += `"${key}": ${JSON.stringify(value)},`;    
+          }
+        }
+        return `${json.charAt(json.length-1) === ',' ? json.slice(0, json.length-1) : json}}`;
+      }
```
magicmatatjahu, line 77: Did you think about solution when you create based on given instance the normal JS object with keys and at the end make `JSON.stringify`, so you operate on object (which should be easier to handle) not on string?

test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      get sTestPatternProperties(): Map<String, string> | undefined { return this._sTestPatternProperties; }
+      set sTestPatternProperties(sTestPatternProperties: Map<String, string> | undefined) { this._sTestPatternProperties = sTestPatternProperties; }
+    
+      public marshal() : string {
+        let json = '{'
+        if(this.stringProp !== undefined) {
+          json += `"stringProp": ${JSON.stringify(this.stringProp)},`; 
+        }
+        if(this.numberProp !== undefined) {
+          json += `"numberProp": ${JSON.stringify(this.numberProp)},`;
```
magicmatatjahu, line 136: `${JSON.stringify(7)}` makes this `"7"` not `7`, so you should pass numbers (and booleans) as values not as string.

test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+            json += `"${key}": ${JSON.stringify(value)},`;    
+          }
+        }
+        //Remove potential last comma 
+        return `${json.charAt(json.length-1) === ',' ? json.slice(0, json.length-1) : json}}`;
+      }
+    
+      public static unmarshal(json: string | object): Test {
+        const obj = typeof json === "object" ? json : JSON.parse(json);
+        const instance = new Test({});
```
magicmatatjahu, line 163: What about the case when some properties is required? User will have TS error that we we try to pass the empty object without required properties. To fix this, please change it to:

```suggestion
        const instance = new Test({} as unknown as any);
```

Please also add at least the one required property in the json schema :)

test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+          //Check all pattern properties
+          if (key.match(new RegExp('^S(.?)test'))) {
+            if (instance.sTestPatternProperties === undefined) {instance.sTestPatternProperties = new Map();}
+            instance.sTestPatternProperties.set(key, value as any);
+            continue;
+          }
+    
+          if (instance.additionalProperties === undefined) {instance.additionalProperties = new Map();}
+          instance.additionalProperties.set(key, value as any);
+        }
```
magicmatatjahu, line 186: You should split this for to the two for, one for additionalProperties and one for patternProperties

magicmatatjahu (changes requested):
Some things should be improved :) Overall good.

src/generators/typescript/presets/CommonPreset.ts
jonaslagoni: Right, good catch! Also, line 23+24 needs to use the property name as is (without formatting) otherwise validation fails. Changed.

src/generators/typescript/presets/CommonPreset.ts
jonaslagoni: I think that makes sense yea 👍 Changing

test/generators/typescript/preset/MarshallingPreset.spec.ts
jonaslagoni: That's actually a good point, I wonder if there are any performance improvements to be had there too 🤔? Or if string manipulation is faster.

jonaslagoni: I wouldn't mind changing it tbh, as it would improve readability I think. But not if we half the performance 😆

magicmatatjahu: Maybe performance will be worse, but then you don't handle manually cases such as https://github.com/asyncapi/modelina/pull/315#discussion_r685813405 so object would improve readability :)

jonaslagoni: Hmm... I think we keep it as is for now, in the future, we can always address the readability, as un/marshalling will be one of the main executed functions for many libraries, I would opt for performance > readability. What do you think @magicmatatjahu ?

magicmatatjahu: ok :) Could you create issue for that and point to our discussion? So then, remember about this https://github.com/asyncapi/modelina/pull/315#discussion_r685813405 case.

jonaslagoni: I can yes, and yea I will remember that comment, I will go through all before pushing changes, so it will be addressed 😄

jonaslagoni: https://github.com/asyncapi/modelina/issues/317

test/generators/typescript/preset/MarshallingPreset.spec.ts
jonaslagoni: Damn, good catch once again 👍 `as any` is enough though.

test/generators/typescript/preset/MarshallingPreset.spec.ts
jonaslagoni: I don't think I can do that as I default to adding the property to `additionalProperties` property if not part of pattern properties.

magicmatatjahu: Ahh yes, you're right. It can stay.

magicmatatjahu (approved):
LGTM

PR #316, opened 2021-08-09
feat: added C# class and enum generator

jonaslagoni:
**Description**
This PR adds C# class and enum generator alongside tests.


src/generators/csharp/renderers/ClassRenderer.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    return this.renderBlock(content);
+  }
+
+  async renderAccessors(): Promise<string> {
+    const properties = this.model.properties || {};
+    const content: string[] = [];
+    const accessorFactory = async (property: CommonModel, propertyName: string, type: PropertyType) => {
+      const formattedPropertyName = pascalCase(this.nameProperty(propertyName, property));
```
magicmatatjahu, line 60: Why do you wanna wrap the nameProperty with the `pascalCase`? If it's needed, It should be done inside `nameProperty` function.

magicmatatjahu: Also `accessorFactory` function could be written outside this function (as private function of Renderer), because as I see you don't use the closures.

magicmatatjahu (changes requested):
I'm not familiar with C# language so I only add the comments related to the code.

src/generators/csharp/renderers/ClassRenderer.ts
jonaslagoni: The reason for wrapping it, is because the accessor name (should always be upper-case first) and the actual property should be lowercase for the accessors to take effect. 
I.e. the following needs to be generated, notice camelcase property name and uppercase camelcase property name for the accessor.
```csharp
public class Address {
  private string streetName;
  public string StreetName 
  {
    get { return streetName; }
    set { streetName = value; }
  }
}
```

test/generators/csharp/CSharpGenerator.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+          type: 'string'
+        }
+      },
+    };
+
+    const inputModel = await generator.process(doc);
+    const model = inputModel.models['_address'];
+
+    let classModel = await generator.renderClass(model, inputModel);
+    expect(classModel.result).toMatchSnapshot();
```
magicmatatjahu, line 58: Please aware that using snapshot isn't a good idea, because then you must "believe" of content saved in snapshot. Why did you start using the snapshots?

src/generators/csharp/renderers/ClassRenderer.ts
magicmatatjahu: @jonaslagoni It's a design pattern for C# or what?

test/generators/csharp/CSharpGenerator.spec.ts
jonaslagoni: Because it is simply too much work to continue to manually update the expected results when there is this much content. To me, this is not a problem, as you inspect the changes of the snapshots in PR's and manually when they are changed.

src/generators/csharp/renderers/ClassRenderer.ts
jonaslagoni: Not a design pattern per say, or is accessors a design pattern? 🧐 It is simply just the syntax

magicmatatjahu: I mean, should it be documented that we will generate accessors?

test/generators/csharp/CSharpGenerator.spec.ts
magicmatatjahu: Personally I don't like snapshot, but if you want use them, feel free.

src/generators/csharp/renderers/ClassRenderer.ts
jonaslagoni: We have this: https://github.com/jonaslagoni/generator-model-sdk/blob/feature/csharp/docs/presets.md#class-3

But it comes back to https://github.com/asyncapi/modelina/issues/318

test/generators/csharp/CSharpGenerator.spec.ts
jonaslagoni: If it was as simple as copy/paste the content, I would mind. But the content it produces need to be converted, because of different escaped characters, which is making it a pain 😆

src/generators/csharp/renderers/ClassRenderer.ts
jonaslagoni: I mean, it is based on their official documentation - https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties#the-set-accessor 

It is just how they do it in C# 🤷

magicmatatjahu (approved):

jonaslagoni:
Thanks 🎉

Issue #128, opened 2021-03-05
JSON example preset


jonaslagoni:
#### Reason/Context
I have had a few use-cases now where I need to generate example JSON based on JSON Schema, and if we can generate the underlying model we should be able to generate an example JSON for that model as well. 

#### Description
It does not have to be fancy just a quick and dirty example will suffice.

jonaslagoni:
Gonna close this one down as we already have it implemented in one generator. Specific issues can then be created for each generator, when the use-cases occurs.

PR #327, opened 2021-08-13
feat: example function for TS class renderer

jonaslagoni:
**Description**
This PR introduces a new TS example renderer, which renders an example object creation with the associated class renderer. This is especially useful when you want to quickly generate an example instance, for example in auto-generated tests.

**Related issue(s)**
Partly solves https://github.com/asyncapi/modelina/issues/128


jonaslagoni:
My honest opinion here is that we are stretching the use-case of the library a bit, and it should probably not be done... Thoughts @magicmatatjahu ?

test/generators/typescript/preset/ExamplePreset.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { TypeScriptGenerator, TS_COMMON_PRESET } from '../../../../src/generators';
+const doc = {
+  $id: 'Test',
+  type: 'object',
+  additionalProperties: true,
+  required: ['string prop'],
```
magicmatatjahu, line 6: What about required properties? Should be rendered only these props, or another also, non-required?

src/generators/typescript/presets/utils/ExampleFunction.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Render `example` function based on model properties
+ */
+export default function renderExampleFunction({ renderer, model }: {
+  renderer: TypeScriptRenderer,
+  model: CommonModel,
+}): string {
+  const properties = model.properties || {};
+  const setProperties = [];
+  for (const [propertyName, property] of Object.entries(properties)) {
```
magicmatatjahu, line 63: What about `patternProperties` and `additionalProperties`? Will be added in the followup?

src/generators/typescript/presets/utils/ExampleFunction.ts
```diff
@@ -0,0 +1,78 @@
+import { TypeScriptRenderer } from '../../TypeScriptRenderer';
+import { CommonModel } from '../../../../models';
+
+/**
+ * LAZY implementation, find first acceptable value
```
magicmatatjahu, line 5: LAZY? 😄  It's related to draft PR, or you really think about Lazy?

test/generators/typescript/preset/__snapshots__/ExamplePreset.spec.ts.snap
```diff
@@ -0,0 +1,69 @@
...
+  get sTestPatternProperties(): Map<String, string> | undefined { return this._sTestPatternProperties; }
+  set sTestPatternProperties(sTestPatternProperties: Map<String, string> | undefined) { this._sTestPatternProperties = sTestPatternProperties; }
+
+  public static example(): Test {
+    const instance = new Test({} as any);
+    instance.stringProp = \\"string\\";
+    instance.numberProp = 0;
+    instance.objectProp = NestedTest.example();
+    return instance;
+  }
```
magicmatatjahu, line 42: Did you think about using the `openapi-sampler` or something like this? I understand that you wanna add examples directly to the props, but maybe `openapi-sampler` + Object.assign(...) will make the implementation easier, what do you think? I know that problem will be with nested objects, like in your example for `NestedTest`.

test/generators/typescript/preset/ExamplePreset.spec.ts
jonaslagoni: All properties are rendered if it can find an example value. It does not care whether it's required or optional as it is just a "dummy" generator 😄

src/generators/typescript/presets/utils/ExampleFunction.ts
jonaslagoni: I have no use-case for this, which is also why it is a lazy implementation, find the first example representation you can and use that 🙂 That being said, if others find the need they can of course adapt it 😄

src/generators/typescript/presets/utils/ExampleFunction.ts
jonaslagoni: Lazy as in find first acceptable example value and use that, you can even change the generation of values.

test/generators/typescript/preset/__snapshots__/ExamplePreset.spec.ts.snap
jonaslagoni: We could, but... remember each generator can define their naming behavior, meaning a property name, is not the same as the JSON property name.

One could do, which I currently am in the ts-nats-template use `openapi-sampler` and the un/marshal functionality, but that creates a direct dependency and unnecessary complexity in my opinion.

magicmatatjahu: ok :) Someone can also write his own preset to achieve similar output based on `openapi-sampler`.

src/generators/typescript/presets/utils/ExampleFunction.ts
magicmatatjahu: I always think about `Lazy` in programming as of `lazy` initialisation of something, so probably `inferring first acceptable value` will be better to describe the logic of function rather than `Lazy`

src/generators/typescript/presets/utils/ExampleFunction.ts
magicmatatjahu: Yeah, we can add support for this (as option) if someone will raise that issue.

test/generators/typescript/preset/ExamplePreset.spec.ts
magicmatatjahu: We can add support for this (as option) if someone will raise that issue :)

jonaslagoni: Yea exactly 👍

src/generators/typescript/presets/utils/ExampleFunction.ts
jonaslagoni: Yea exactly 👍

src/generators/typescript/presets/utils/ExampleFunction.ts
jonaslagoni: True, lets be precise 👍

test/generators/typescript/preset/__snapshots__/ExamplePreset.spec.ts.snap
jonaslagoni: Yes, exactly 👍

src/generators/typescript/presets/utils/ExampleFunction.ts
```diff
@@ -0,0 +1,78 @@
+import { TypeScriptRenderer } from '../../TypeScriptRenderer';
+import { CommonModel } from '../../../../models';
+
+/**
+ * Inferring first acceptable value from the model.
+ * 
+ * @param model
```
magicmatatjahu, line 7: If you add params, please also add param for `renderer` :) My last comment, everything is good.

magicmatatjahu (approved):

PR #328, opened 2021-08-16
feat: pass options to processors

jonaslagoni:
**Description**
This PR introduces ProcessorOptions, which can be used by the processors to change certain behavior of the internals. This is especially useful for example when the library is used in browsers the AsyncAPI processor needs to alter the `path` for derefences.

### DIFF
--- a/src/generators/AbstractGenerator.ts
+++ b/src/generators/AbstractGenerator.ts
@@ -9,7 +9,7 @@ export interface CommonGeneratorOptions<P extends Preset = Preset> {
   };
   defaultPreset?: P;
   presets?: Presets<P>;
-  processorOption?: ProcessorOptions
+  processorOptions?: ProcessorOptions
 }
 
 export const defaultGeneratorOptions: CommonGeneratorOptions = {
@@ -36,7 +36,7 @@ export abstract class AbstractGenerator<Options extends CommonGeneratorOptions =
   public abstract render(model: CommonModel, inputModel: CommonInputModel): Promise<RenderOutput>;
 
   public process(input: Record<string, unknown>): Promise<CommonInputModel> {
-    return InputProcessor.processor.process(input, this.options.processorOption);
+    return InputProcessor.processor.process(input, this.options.processorOptions);
   }
 
   public async generate(input: Record<string, unknown> | CommonInputModel): Promise<OutputModel[]> {
--- a/src/generators/AbstractGenerator.ts
+++ b/src/generators/AbstractGenerator.ts
@@ -10,7 +10,7 @@ export interface CommonGeneratorOptions<P extends Preset = Preset> {
   };
   defaultPreset?: P;
   presets?: Presets<P>;
-  parserOptions?: ProcessOptions
+  processOptions?: ProcessOptions
 }
 
 export const defaultGeneratorOptions: CommonGeneratorOptions = {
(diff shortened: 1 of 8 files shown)


jonaslagoni:
Fixed code smell

jonaslagoni:
@smoya it is ready for review, if you are up for it 😄

src/generators/AbstractGenerator.ts
```diff
@@ -1,16 +1,15 @@
...
-import { RenderOutput } from '../models/RenderOutput';
-
 export interface CommonGeneratorOptions<P extends Preset = Preset> {
   indentation?: {
     type: IndentationTypes;
     size: number;
   };
   defaultPreset?: P;
   presets?: Presets<P>;
+  processorOption?: ProcessorOptions
```
smoya, line 12: Minor: does it make sense to rename the following?
```suggestion
  processorOptions?: ProcessorOptions
```

src/models/ProcessorOptions.ts
```diff
@@ -0,0 +1,6 @@
+export interface AsyncapiProcessorOptions {
+  path?: string
```
smoya, line 2: The reality is that I don't fully understand the use case you mentioned in the PR description.
I'm concerned we are adding this here because of a particular use case on our side (AsyncAPI) but the cost of it is strongly coupling with AsyncAPI itself. 

Would you mind clarifying the use case and perhaps add a real example?

src/generators/AbstractGenerator.ts
jonaslagoni: yea, makes sense 👍

src/models/ProcessorOptions.ts
jonaslagoni: Hmm, it is honestly a good point -  added a follow-up issue in the parser as well https://github.com/asyncapi/parser-js/issues/344

The example can be found in the follow-up issue. 

I think it is a difficult question, to answer whether we need to be this coupled, but do you see another approach? 🤔

smoya: Ok, i understand now the purpose of this change (thanks to the example you set in the issue).

A small suggestion (if makes sense): I see that if we ever support any other input beside the path to an AsyncApi document (e.g. any dsl) this could also be needed (depending on the parser used).
would it make sense to just rename the `asyncapi` property on the option object to a more generic one such as `input` or similar.

If you don't see it make sense, forget about it.

test/processors/InputProcessor.spec.ts
```diff
@@ -73,7 +73,22 @@ describe('InputProcessor', () => {
...
+      expect(asyncInputProcessor.process).toHaveBeenNthCalledWith(1, inputSchema, undefined);
+      expect(asyncInputProcessor.shouldProcess).toHaveBeenNthCalledWith(1, inputSchema);
+      expect(defaultInputProcessor.process).not.toHaveBeenCalled();
+      expect(defaultInputProcessor.shouldProcess).not.toHaveBeenCalled();
+    });
+    test('should be able to process AsyncAPI schema input with options', async () => {
+      const {processor, asyncInputProcessor, defaultInputProcessor} = getProcessors(); 
+      const options: ProcessorOptions = {
+        asyncapi: {
+          path: 'test'
```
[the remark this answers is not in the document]
smoya: Is there a way to add a test that uses the new option object so we can test the object is being passed all the way down?

src/models/ProcessorOptions.ts
jonaslagoni: Maybe, but adding it as generic already without seeing the use-case for it, might be the wrong call? 🤔

smoya: I dont think it is a wrong movement to just rename it for abstracting it from AsyncAPI but as mentioned, not strong opinion at this moment.

jonaslagoni: I think it is a pre-optimization that might never be needed elsewhere. Let's keep it as is, for now, we can always reevaluate it for v1

test/processors/InputProcessor.spec.ts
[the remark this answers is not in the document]
jonaslagoni: Should be, let me take a look.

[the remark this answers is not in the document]
jonaslagoni: Done 👍

PR #333, opened 2021-08-30
fix: c# generator should render types as nullable by default

jonaslagoni:
**Description**
This PR introduces a change to the underlying C# type rendering. It changes the rendering for `float`, `int` and `bool` to being optional primitives, instead of primitives. As we don't render required properties (yet, gonna do a follow-up issue) we should make them optional.

### DIFF
--- a/src/generators/csharp/CSharpRenderer.ts
+++ b/src/generators/csharp/CSharpRenderer.ts
@@ -76,11 +76,11 @@ export abstract class CSharpRenderer extends AbstractRenderer<CSharpOptions> {
     case 'string':
       return 'string';
     case 'integer':
-      return 'int';
+      return 'int?';
     case 'number':
-      return 'float';
+      return 'float?';
     case 'boolean':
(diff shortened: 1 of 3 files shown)


src/generators/csharp/CSharpRenderer.ts
```diff
@@ -76,11 +76,11 @@ export abstract class CSharpRenderer extends AbstractRenderer<CSharpOptions> {
     case 'string':
       return 'string';
```
magicmatatjahu, line 77: Why not string? I am dummy with C# so I ask 😅

jonaslagoni: String is a reference, which is by default nullable. There is something called nullable types in C# 8, but don't want to touch that too much, as I have no idea what it is yet 😆 https://stackoverflow.com/questions/32853482/how-can-i-make-my-string-property-nullable

magicmatatjahu (approved):

PR #334, opened 2021-08-30
feat: add c# serializer and deserializer functionality

jonaslagoni:
**Description**
This PR adds functionality to render serializer and deserializer functionality to C# class output. This will enable correct behavior from `System.Text.Json.JsonSerializer.Serialize` and `System.Text.Json.JsonSerializer.Deserialize`, so you can do the following in your code:
```c#
TestModel temp = new TestModel
{
    ...
};
string json = JsonSerializer.Serialize(temp);
TestModel output = JsonSerializer.Deserialize<TestModel>(json);
```

**Related issue(s)**
Partly solves https://github.com/asyncapi/modelina/issues/262

### DIFF
--- a/src/generators/csharp/presets/JsonSerializerPreset.ts
+++ b/src/generators/csharp/presets/JsonSerializerPreset.ts
@@ -3,8 +3,8 @@ import { CSharpPreset } from '../CSharpPreset';
 import { getUniquePropertyName, DefaultPropertyNames, FormatHelpers } from '../../../helpers';
 import { CommonModel } from '../../../models';
 
-function renderMarshalAdditionalProperties(model: CommonModel, renderer: CSharpRenderer) {
-  const marshalAdditionalProperties = '';
+function renderSerializeAdditionalProperties(model: CommonModel, renderer: CSharpRenderer) {
+  const serializeAdditionalProperties = '';
   if (model.additionalProperties !== undefined) {
     let additionalPropertyName = getUniquePropertyName(model, DefaultPropertyNames.additionalProperties);
     additionalPropertyName = FormatHelpers.upperFirst(renderer.nameProperty(additionalPropertyName, model.additionalProperties));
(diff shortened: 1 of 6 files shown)


jonaslagoni:
Blocked by #333

src/generators/csharp/presets/JsonSerializerPreset.ts
```diff
@@ -0,0 +1,247 @@
...
+      renderer.addDependency('using System.Text.Json;');
+      renderer.addDependency('using System.Text.Json.Serialization;');
+      renderer.addDependency('using System.Text.RegularExpressions;');
+      
+      const formattedModelName = renderer.nameType(model.$id);
+      const unmarshal = renderUnmarshal({renderer, model});
+      const marshal = renderMarshal({renderer, model});
+
+      return `[JsonConverter(typeof(${formattedModelName}Converter))]
+${await renderer.defaultSelf()}
```
magicmatatjahu, line 231: You can use the content from previous preset:

```suggestion
${content}
```

but then you must add `content` in the arguments  of `self` function.

magicmatatjahu (commented):
If you called functionality as serializer, then all TS functions should also be called `serialize`, not `marshal` :) or convert.

jonaslagoni:
> If you called functionality as serializer, then all TS functions should also be called `serialize`, not `marshal` :) or convert.

Hmm, I see your point, problem is, for C# it is part of the default dependencies `System.Text.Json` meaning it is called serializer, nothing I can change about that 😄 We could change the naming for C#, however in that language many refer to it as un/marshal. Not sure it is a problem? 🤔

src/generators/csharp/presets/JsonSerializerPreset.ts
jonaslagoni: Done 👍

magicmatatjahu:
> Hmm, I see your point, problem is, for C# it is part of the default dependencies System.Text.Json meaning it is called serializer, nothing I can change about that 😄 We could change the naming for C#, however in that language many refer to it as un/marshal. Not sure it is a problem? 🤔

We can left the `marshal` naming. It was my first concern when I saw marshal names in source code and `Serializer` in the output 😄

magicmatatjahu (approved):
LGTM

jonaslagoni:
> > Hmm, I see your point, problem is, for C# it is part of the default dependencies System.Text.Json meaning it is called serializer, nothing I can change about that 😄 We could change the naming for C#, however in that language many refer to it as un/marshal. Not sure it is a problem? 🤔
> 
> We can left the `marshal` naming. It was my first concern when I saw marshal names in source code and `Serializer` in the output 😄

Ahh! Let's change the naming inside the file, to make it consistent 👍 I thought you meant that the TS output has marshal and C# has sterilizer 😄

magicmatatjahu:
@jonaslagoni Yeah, exactly

magicmatatjahu (approved):

PR #337, opened 2021-09-01
fix: un/marshal TS class functions did not render references correctly

jonaslagoni:
**Description**
This PR fixes a couple of things: 
1. When enums are referenced, they should be un/marshaled as is, and not through helper functions
2. Ensured that pattern properties and additional properties render the correct un/marshal functionality based on type.

### DIFF
--- a/test/generators/typescript/preset/MarshallingPreset.spec.ts
+++ b/test/generators/typescript/preset/MarshallingPreset.spec.ts
@@ -14,7 +14,7 @@ const doc = {
   required: ['string prop'],
   properties: {
     'string prop': { type: 'string' },
-    enumProp: { $id: 'EnumTest', enum: ['Some enum String']},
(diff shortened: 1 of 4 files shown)


test/generators/typescript/preset/MarshallingPreset.spec.ts
```diff
@@ -3,19 +3,24 @@
...
+    }
+  },
   $id: 'Test',
   type: 'object',
-  additionalProperties: true,
+  additionalProperties: {$ref: '#/definitions/NestedTest'},
   required: ['string prop'],
   properties: {
     'string prop': { type: 'string' },
+    enumProp: { $id: 'EnumTest', enum: ['Some enum String']},
```
magicmatatjahu, line 17: Overall good, but please add second value in enum :) For the future.

jonaslagoni:
Thanks, @magicmatatjahu!

Issue #287, opened 2021-07-05
User-centric docs


jonaslagoni:
#### Reason/Context
At the moment all docs are contributor-centric and do not explain how a user can use the project.

#### Description
We need to add user-centric docs together with examples in generator language.

Should solve the following issues: https://github.com/asyncapi/modelina/issues/318 #296 https://github.com/asyncapi/modelina/issues/45

jonaslagoni:
The initial take on how we could change the documentation to be more user-centric. I declared each question as a potential use-case (or more specifically a feature in Modelina) that the user of the library can encounter. Each of these use-cases will be an individual document, that will explain how to accomplish it and the reason behind it.

1. Generate Go models
3. Generate C# models
4. Generate Java models
5. Generate TypeScript models
6. Generate JavaScript models
7. Integrate Modelina in a website
8. Integrate Modelina in an AsyncAPI generator template
9. Generate each model in the same file
10. Generate a model in separate files
11. Change the generated indentation type and size
12. Change the naming format for properties
12. Change the naming format for data models
13. Include a custom function in the data model
15. Use the models for data transfer
16. Extend the logic of an existing renderer
17. Build your own model renderer
18. Create your own models from the ground up, instead of a supported input
 
## Language-specific use-cases
### TypeScript
1. Generate an `interface` instead of `class`
2. Generate un/marshal functions for each class


### Java
1. Include `equals` function for the class
2. Include `hashCode` function for the class
3. Change the collection type for arrays
4. Include `toString` function for the class
5. Include `Java doc` for properties
6. Include `javax` validation constraints annotations for properties
7. Include `Jackson` annotations for the class

Anything missing?

derberg:
The majority of these docs topics are about coding, using Modelina API. Providing initial docs is easy, harder to maintain in the future. Did you plan to write just markdown or rather `examples` for each, examples that can be tested on a PR?

jonaslagoni:
> The majority of these docs topics are about coding, using Modelina API. Providing initial docs is easy, harder to maintain in the future. Did you plan to write just markdown or rather `examples` for each, examples that can be tested on a PR?

Good point 🤔 Most of these are actually already in place as tests. They are just all over the place (not structured to be easily discovered based on example), would you suggest we create an extra specific test per example, or restructure the tests, so there are no duplicates @derberg ?

derberg:
for user most important is example, not test itself so might be that tests are in one location, refactored to not have duplications. More important is to have "runnable examples"

[...]

PR #338, opened 2021-09-01
docs: initial setup for user-centric docs

jonaslagoni:
**Description**
This PR setup the entire structure of tests and docs, to enable `good-first-issue` label when splitting up #287 into smaller tasks.

**Related issue(s)**
Partly solves #287


jonaslagoni:
@alequetzalli would love your input (concerns, or otherwise, if you have any), for this approach we are taking with user-centric documentation 🙂

derberg (commented):
looks awesome,

- I would only consider separate `README.md` under `/docs` that you would link to from main readme, docs readme would list main documents and what can be find there
- I see example test but not example code that is tested 🤔  what is your plan here exactly. What I had in mind was something like https://github.com/asos-craigmorten/opine/tree/main/examples

docs/languages/TypeScript.md
```diff
@@ -0,0 +1,21 @@
+# TypeScript
+There are special use-cases that each language support, therefore this document is for anything related to the TypeScript models.
```
alequetzalli, line 2: There are special use-cases that each language supports; this document pertains to **TypeScript models**.

docs/usage.md
```diff
@@ -0,0 +1,72 @@
+# Usages
+Modelina can be used in many different contexts and has many features all depending on the output language, this document will show you the very basic usages of the library.
```
alequetzalli, line 2: Modelina can be used in many different contexts and has many features, all depending on the output language. This document will walk you through you the library's basic usages.

docs/usage.md
```diff
@@ -0,0 +1,72 @@
...
+- [Generate JavaScript models](#generate-javascript-models)
+
+<!-- tocstop -->
+
+## Understanding the output format
+TODO 
+
+## Generate models from AsyncAPI documents
+
+The library expects the `asyncapi` property for the document to be sat in order to understand the input format.
```
alequetzalli, line 28: typo

"to be set"

docs/usage.md
```diff
@@ -0,0 +1,72 @@
...
+- Generate from a pure JS object:
+
+```ts
+import { TypeScriptGenerator } from '@asyncapi/modelina';
+const generator = new TypeScriptGenerator();
+generator.generate({asyncapi: '2.1.0', ...});
+```
+
+## Generate models from JSON Schema draft 7 documents
+The library expects the `$schema` property for the document to be sat in order to understand the input format, by default if no other inputs are detected, it defaults to JSON Schema draft 7.
```
alequetzalli, line 49: typo :D 

"to be set"

alequetzalli, line 49: The library expects the `$schema` property for the document to be set in order to understand the input format. By default, if no other inputs are detected, it defaults to **JSON Schema draft 7**.

alequetzalli (commented):
Hello @jonaslagoni !! I went through and added my suggestions and ideas for improving the reading experience of the users. Please do feel free to ignore me if something I said really doesn't make sense. I hope this is helpful, do let me know :)

jonaslagoni:
> * I see example test but not example code that is tested 🤔  what is your plan here exactly. What I had in mind was something like https://github.com/asos-craigmorten/opine/tree/main/examples

@derberg you were right, was missing the README.md example, added 👍 

I have added an `example` folder under tests, where I have placed all the current examples as tests.
- Readme example: https://github.com/asyncapi/modelina/blob/072b3699abed1870e1b0be2709ac6a2eb4e0eb34/test/examples/readme.spec.ts
- Usage example: https://github.com/asyncapi/modelina/blob/072b3699abed1870e1b0be2709ac6a2eb4e0eb34/test/examples/usages.spec.ts
- Advanced example: https://github.com/asyncapi/modelina/blob/072b3699abed1870e1b0be2709ac6a2eb4e0eb34/test/examples/advanced.spec.ts

Would this setup not be sufficient?

README.md
```diff
@@ -5,75 +5,46 @@
...
+  <em>The official SDK for generating data models from AsyncAPI and other inputs</em>
 </p>
 
 [![Coverage Status](https://coveralls.io/repos/github/asyncapi/modelina/badge.svg?branch=master)](https://coveralls.io/github/asyncapi/modelina?branch=master)
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 [![All Contributors](https://img.shields.io/badge/all_contributors-8-orange.svg?style=flat-square)](#contributors-)
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 
-AsyncAPI Model SDK is a set of classes/functions for generating data models from JSON Schema and AsyncAPI spec.
+Modelina is used to generate different data models, such as Java, TypeScript classes, Go Structs, etc, for AsyncAPI message payloads among other supported inputs.
```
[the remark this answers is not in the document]
jonaslagoni: Moved to the top instead of https://github.com/asyncapi/modelina/pull/338#discussion_r700663121, too much duplicate text IMO, that explains nothing new 😄

README.md
```diff
@@ -5,75 +5,46 @@
...
-The generation process uses the predefined `CommonModel`s from the input stage to easily generate models regardless of input. 
-The list of supported output languages can be found [here](./docs/generators.md#supported-languages).
-
-Check out [the example](#example) to see how to use the library and [generators document](./docs/generators.md) for more info.
-
-> **NOTE**: Each implemented language has different options, dictated by the nature of the language.
-
-## Example
-
+You can then choose between all the generators, for example using the TypeScript generator, as in this example:
```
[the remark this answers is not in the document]
jonaslagoni: Left out `A generator is a unique kind of function that returns iterator objects with a sequence of values instead of returning single values.` as I feel like it needs more reasoning behind it. This will be explained in `./docs/usage.md#understanding-the-output-format`

docs/advanced.md
```diff
@@ -0,0 +1,78 @@
+# Advanced use-cases for Modelina
+
+<!-- toc is generated with GitHub Actions do not remove toc markers -->
+
+<!-- toc -->
+
+- [Generate each model in the same file](#generate-each-model-in-the-same-file)
+- [Generate a model in separate files](#generate-a-model-in-separate-files)
+- [Include a custom function in the data model](#include-a-custom-function-in-the-data-model)
+- [## Use the models for data transfer](#%23%23-use-the-models-for-data-transfer)
```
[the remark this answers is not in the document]
jonaslagoni: Ahh, good catch 👍

docs/languages/C#.md
```diff
@@ -0,0 +1,13 @@
+# C#
+There are special use-cases that each language support, therefore this document is for anything related to the C# models.
```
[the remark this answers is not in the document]
jonaslagoni: Guess I learned a new word today `pertains` 😄

jonaslagoni:
> Hello @jonaslagoni !! I went through and added my suggestions and ideas for improving the reading experience of the users. Please do feel free to ignore me if something I said really doesn't make sense. I hope this is helpful, do let me know :)

@alequetzalli quality review 🙇 Thanks!

docs/README.md
```diff
@@ -0,0 +1,51 @@
...
+- [Generators (`./generators.md`)](#generators-generatorsmd)
+- [Presets (`./presets.md`)](#presets-presetsmd)
+- [Interpretation of JSON Schema draft 7 (`./interpretation_of_JSON_Schema_draft_7.md`)](#interpretation-of-json-schema-draft-7-interpretation_of_json_schema_draft_7md)
+- [Languages](#languages)
+
+<!-- tocstop -->
+
+This document gives the overview of all the available documentation for Modelina.
+
+### [Usage (`./usage.md`)](./usage.md)
```
jonaslagoni, line 21: I am not sure I like this format.. But I feel like the link to the file should be visible somehow. What do you think? 🤔

alequetzalli:
> > Hello @jonaslagoni !! I went through and added my suggestions and ideas for improving the reading experience of the users. Please do feel free to ignore me if something I said really doesn't make sense. I hope this is helpful, do let me know :)
> 
> @alequetzalli quality review 🙇 Thanks!
> 
> Just a suggestion, when doing a comment about improvement, you can use the [suggestion box](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request):
> 
> ```suggestion
> some specific suggestions on this line.
> ```
> 
> This way it is easier to see which changes you are suggesting 🙂
> 
> Also, I cant re-request a review from you, but of course, feel free to review it again 😄

@jonaslagoni  got it, yes!! I knewww I had done it wrong. hahaha I will use the suggestion box feature moving fwd :D

docs/README.md
alequetzalli: Screen Shot 2021-09-02 at 10 16 51 AM

How about something like this? You still link to the markdown file from the title and the link is visible, but you do not include the file name in the title.

alequetzalli (commented):
Hello again @jonaslagoni !! I added just a few more commit suggestions, hopefully they are helpful. Let me know what you think :D

docs/README.md
jonaslagoni: I guess we have to stick with that yea, it looks better 😄

jonaslagoni:
> Hello again @jonaslagoni !! I added just a few more commit suggestions, hopefully they are helpful. Let me know what you think :D

@alequetzalli appreciate the suggestions 👍  

@derberg I still need your thoughts on the examples tests, I know we are still missing some of the examples from preset.md, etc. but will make a follow up issue for those 😄

derberg:
@jonaslagoni but I don't get where are these examples are supposed to work, I see examples' tests but they also include tested code, and do not test a separate example from another file. I'm probably missing something

jonaslagoni:
> @jonaslagoni but I don't get where are these examples are supposed to work, I see examples' tests but they also include tested code, and do not test a separate example from another file. I'm probably missing something

@derberg I don't follow 😄 The example tests are completely separate, from the rest of the tests, and located under `./test/examples/` that are directly ensuring that the example code runs as expected 🤔 Is that not what you are after?

derberg:
`that are directly ensuring that the example code runs as expected` -> where is the example code?

jonaslagoni:
@derberg https://github.com/asyncapi/modelina/pull/338#issuecomment-911449633

derberg:
exactly what I'm referring to 😄 so ` I see examples' tests but they also include tested code`. So your examples are part of tests, but they should not, example should be separated from test, just like -> https://github.com/asos-craigmorten/opine/tree/main/examples

Users should look only on example, with no noise around, later, if they want, they can check the test that validates the example, but this is optional.

jonaslagoni:
> exactly what I'm referring to 😄 so ` I see examples' tests but they also include tested code`. So your examples are part of tests, but they should not, example should be separated from test, just like -> https://github.com/asos-craigmorten/opine/tree/main/examples

@derberg why? 🙂 I don't see the benefits as the examples are included in the docs, are not examples that directly can be run? Their CLI allows the execution of them where we have no such thing - https://github.com/asos-craigmorten/opine#examples. From that point of view, it makes perfect sense to structure the examples they have, with associated tests. Also, they run a service they can test against, we don't. We would need to include the code exactly as it is now, in order to appropriately test it.

The example tests included here are exactly for one purpose, ensuring said example code in the markdown files is actually runnable. 

Maybe you see it differently? 🤔

derberg:
the problem with the current setup is that you duplicate efforts. You have an example in markdown and then you need an example in test:
- you duplicate work
- on PR you can run test and check if it works but on PR you won't be able to check if the example that you changed needs a test change, which makes it harder for contributors as they need to get direct feedback from human
- you rely on humans to remember the long term that whenever examples in MD files are updated, the tests must be updated too. I think you already know how that usually ends 😄 

It can only work in examples are separate mini-projects that you can easily run with `npm` and each is tested:
- when you modify the example, test will tell you if you break it
- I can locally run `npm run test` and make sure my changes don't break anything
- human is not much needed, only when new use case shows up and new example must be provided

mini projects? yeah, minimal `package.json`, example in separate `index.js` and test for it with `index.spec.js`  in the same dir or separate dir and super simple `readme.md` if needed, but I guess in most cases code comments will be better. Then you just need a TOC of all examples

derberg:
I guess you do not even need `package.json` 🤔 at the end instruction can be like -> run this example as `node examples/generate_some_stuff_whatever/index.js`

jonaslagoni:
@derberg I see your point, gonna convert one and see how it looks.

jonaslagoni:
@derberg checkout https://github.com/asyncapi/modelina/pull/338/commits/bf94e9a665a5f686cce3d5a68628a63efca3ab71, is this what you are looking for? 

@alequetzalli anything from your side in terms of this approach? 🤔

derberg:
@jonaslagoni 

> bit missing in some sense

why? it is perfect, example separated from other things, basically copy/paste. Just add note in main examples readme on how to run examples, and like in Deno example, add a big code comment to example `index.ts` on how to run the example locally

my only question is why there needs to be example-specific jest config + dedicated devDeps? why not use the main ones?

alequetzalli:
> @derberg checkout https://github.com/asyncapi/modelina/pull/338/commits/bf94e9a665a5f686cce3d5a68628a63efca3ab71, is this what you are looking for? I find the example code a bit missing in some sense, from the user perspective, as they are basically designed as a module 🤔 
> 
> 
> 
> @alequetzalli anything from your side in terms of this approach? 🤔 
> 
> 
> 
> 

Hey Jonas, I don't have any more comments or opinions at this point 😄

jonaslagoni:
@derberg converted all examples, please have a look at how I reference the examples in the docs, do you think that is fine?

Also the actual test for the example code, I merely ensure that it executes, in order to avoid duplicating a lot of duplicate expected content. IMO that should be continued to be tested in the `./test`'s,

examples/custom-logging/index.ts
```diff
@@ -0,0 +1,21 @@
...
+    house_number: { type: 'number' },
+    marriage: { type: 'boolean', description: 'Status if marriage live in given house' },
+    pet_names: { type: 'array', items: { type: 'string' } },
+    state: { type: 'string', enum: ['Texas', 'Alabama', 'California', 'other'] },
+  },
+  required: ['street_name', 'city', 'state', 'house_number', 'state'],
+};
+
+export async function generate(logger: ModelLoggingInterface) : Promise<OutputModel[]> {
+  Logger.setLogger(logger);
```
derberg, line 18: for logging example I would expect to actually see an example of ModelLoggingInterface

derberg (commented):
I left one specific comment for one of examples.

Also:
- what about the question I had `my only question is why there needs to be example-specific jest config + dedicated devDeps? why not use the main ones?`
- user still doesn't know, exactly how to run these examples after cloning the repo, no copy/paste command in readme or as part of the comment in the head of the example


> Also the actual test for the example code, I merely ensure that it executes, in order to avoid duplicating a lot of duplicate expected content. IMO that should be continued to be tested in the ./test's,

it does look super basic, the expected result would be an advantage for users to see what they get without running the example.
Why not running these tests as part of all the other tests?

examples/custom-logging/index.ts
jonaslagoni: I thought about doing that, but I could not figure out to write the test to verify it works if I can't control the input 🤔?

derberg: what do you mean you can't control the input? what input, you have a hardcoded doc here, or am I missing something? 🤔

jonaslagoni:
>     * what about the question I had `my only question is why there needs to be example-specific jest config + dedicated devDeps? why not use the main ones?`

@derberg I have removed the example-specific jest config.

>     * user still doesn't know, exactly how to run these examples after cloning the repo, no copy/paste command in readme or as part of the comment in the head of the example

When you say `run these examples`, do you mean how you can run the test? Cause there is no direct way to run the example code as is since everything is in memory output. 

> it does look super basic, the expected result would be an advantage for users to see what they get without running the example.

Problem is that once output change, it is a lot of "clutter" changes that all represent the same thing, meaning PR's will be cluttered with the same changes 😟 Especially since we have actual tests that ensure the `generate` function returns as expected. Not sure how you see we can handle this more efficiently?

> Why not running these tests as part of all the other tests?

They are, they are executed when running `npm run test`.

examples/custom-logging/index.ts
jonaslagoni: I use the jest mocked functions to ensure that the logger actually logs something - https://github.com/asyncapi/modelina/pull/338/files/7e6e530c96a65d52c7d6159b18181bf71c190f78#diff-582bf3a7b78962a1559eab563f529c43fe057b9519bde459e24d2f8d47678c8c

alequetzalli (approved):

examples/custom-logging/index.ts
jonaslagoni: But of course, this comes down to, what exactly am I asserting or ensuring works inside the example tests and I am not sure how to balance it 😬

derberg:
@jonaslagoni sorry for keeping you waiting. 

What I basically mean is to have examples that one can easily run and see results without seeing test run and this is possible.

- you have test for each example, take it to the advantage use them as integration tests, run them as any other tests, no need to have package.json for each example I think. So far I see only devDeps there in the example that could easily be picked up from main package.json
- enable user to run just the example, without really running the test, and whatever user must-see, use console.log for it
- have test actually rest the output, as you want to make sure that once users run the example, they actually see what you expect them to see

so tests might be duplicated, yes, but you gain super great onboarding experience. 

I can try to provide an example but need time until the end of the week

jonaslagoni:
@derberg so this is the only setup I could reliably get to work so the user can download the examples and run them (`npm run start`), while Modelina test script could still pick the tests up and run them. 

The only way this works at the moment is that any dependencies that the examples are using Modelina will also need, otherwise we need to recursively install dependencies in each example folder. I want to avoid this at all cost, as it will just clutter the Modelina setup. I dont think we will encounter any examples that will require dependencies that Modelina will not have regardless. So I think that will work out fine 🤞.

If we ever want to integrate something like https://github.com/asos-craigmorten/opine/tree/main/examples does, where the examples can be executed by Modelina itself, we will require #172 to be solved first. So for now I am pleased with this setup.

@derberg what are your thoughts here?

examples/asyncapi-from-object/index.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export async function generate(logCallback: (msg: string) => void): Promise<void> {
+  const models = await generator.generate(AsyncAPIDocument);
+  for (const model of models) {
+    logCallback(model.result);
+  }
+}
+
+generate(console.log);
```
derberg, line 38: I know it may sound strange but you literally caused collateral damage to my brain when I saw it. I needed a minute to figure out "what? how? why?" 😄 

I strongly suggest:
```ts
export async function generate(logCallback: (msg: string) => void): Promise<void> {
  const models = await generator.generate(AsyncAPIDocument);
  for (const model of models) {
    console.log(model);
  }
}

generate(model.result);
```

examples/asyncapi-from-parser/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "devDependencies": {
+    "@types/jest": "^26.0.15",
+    "cross-env": "^7.0.2",
+    "jest": "^26.6.3",
+    "ts-jest": "^26.4.4",
+    "ts-node": "^10.2.1"
+  },
+  "scripts": {
+    "start": "ts-node index.ts",
+    "test": "cross-env CI=true jest --config=../../jest.config.js ./examples/asyncapi-from-object/index.spec.ts"
```
derberg, line 18: do we need this script? and all the `dependencies` and `devDependencies` if at the we have `"test:examples": "cross-env CI=true jest ./examples",` in `package.json` + these dependencies are already there in the root?

can't we have just:
```json
{
  "scripts": {
    "start": "ts-node index.ts"
   }
}

File: examples/asyncapi-from-parser/index.spec.ts [hunk 13]
```diff
@@ -0,0 +1,15 @@
+import {generate} from './index';
+
+describe('Should be able to process AsyncAPI object from parser', () => {
+  afterAll(() => {
+    jest.clearAllMocks();
+  });
+  test('and should log expected output to console', async () => {
+    const callbackMock = jest.fn();
+    await generate(callbackMock);
+    const expectedConsoleLog = `export interface AnonymousSchema_1 {
```
derberg, line 10: I guess you're doing it this way because of how console.log is actually working. If example would write to file, and test would read the file and forward do snapshot testing, do you think it would be too much?

derberg (commented):
it's getting much better, few comments left

examples/asyncapi-from-object/index.ts
jonaslagoni: @derberg now you made my brain explode, can you clarify your example? 😆 

I guess you want the remove all of `logCallback` altogether to simplify the example right? If that's the case, it will be very difficult to test the expected output from the generation and in some cases completely impossible. If we can accept that, then yes we can do that.

examples/asyncapi-from-parser/package.json
jonaslagoni: Because I want users to stand inside the example repo, and write `npm run start` to run the example as well as tests `npm run test`. That won't work if it has no dependencies defined. 

And I dont want to overrun the root package.json with scripts.

examples/asyncapi-from-parser/index.spec.ts
jonaslagoni: I think we can use snapshots 🤔 Let me update the examples.

examples/asyncapi-from-object/index.ts
derberg: simplify to maximum is my goal always 😄 
then I guess the only way is files and snapshots

examples/asyncapi-from-parser/package.json
derberg: I just cloned your branch, did `npm install` in root only, then modified the package.json:

```json
{
  "scripts": {
    "start": "ts-node index.ts",
    "test": "cross-env CI=true jest --config=../../jest.config.js ./examples/asyncapi-from-object/index.spec.ts"
  }
}
```

and then `npm run test` works like a charm.

for instructions it is fine if user needs to install root project imho, otherwise just add another script `"install": "cd ../.. && npm i"` if you want to hide complexity. Just having these dependencies all over the place will complicate things a lot, management of dependencies.

jonaslagoni: Ahh right, cause the test script actually uses root dependencies. 

Did you try for the `asyncapi-from-parser` example, as it uses the parser dependency?

derberg: works like a charm, parser is at the root of the project, so in `node_modules`. why you think it should not work?

jonaslagoni: I dont get the setup you have, cause the `npm run start` cannot work with root dependencies and I dont like the requirement that you have to install `ts-node` globally, unless you have a setup I can't think of 😄

Can you provide the full example so you inside the example folder can successfully run `start` and `test` without defining dependencies for the example? 🤔 That also does not require `ts-node` to be globally installed separately.

derberg: `ts-node` must be added in the project root to `devDependencies`. It is a `devDependency` and I think it is completely fine approach to have it there

jonaslagoni: Now it snapped together how to achieve it, updated it 😄

examples/asyncapi-from-parser/index.ts
```diff
@@ -0,0 +1,39 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+};
+
+export async function generate(logCallback : (msg: string) => void) : Promise<void> {
```
derberg, line 32: sorry to be a pain here but I strongly thing this console.log callback is polluting the example. We should do this:

```js
  const parsedDoc = await parse(JSON.stringify(AsyncAPIDocument));
  const models = await generator.generate(parsedDoc as any);
  for (const model of models) {
    console.log(model.result);
  }
}

generate();
```

and test should do
```js
  test('and should log expected output to console', async () => {
    console.log = jest.fn();
    await generate();
    expect(console.log).toMatchSnapshot();
  });
```

I think it is much cleaner

derberg (commented):
small comment, sorry 😅

examples/asyncapi-from-parser/index.ts
jonaslagoni: I think there was a reason why I chose to use callback, but cannot remember it anymore 😆 Changed!

derberg (dismissed):
man, it took us some time to figure it out, not sure about you though but I think the result is super super cool 🚀 

Well done 👏🏼 

Now, I suggest every missing example gets an individual issue with label `hacktoberfest` and we promote it in October

jonaslagoni:
> man, it took us some time to figure it out, not sure about you though but I think the result is super super cool 🚀
> 
> Well done 👏🏼
> 
> Now, I suggest every missing example gets an individual issue with label `hacktoberfest` and we promote it in October

Agreed 🎉 Thanks for the reviews @derberg and @alequetzalli, gonna leave it open until tomorrow, so people have time to comment if they want, and I will start pouring out individual issues today and tomorrow 😄

jonaslagoni:
@derberg need another approval.

- Added template example for new contributors - https://github.com/asyncapi/modelina/pull/338/commits/3b2c15327ce2630a1b764f3d6f803fba72c58c89
- Made sure examples use default generator settings - https://github.com/asyncapi/modelina/pull/338/commits/56045e28a8ae717ffffc6e673adb4d5928444b19

derberg (approved):
loving the approach with the template!

jonaslagoni:
@derberg sorry, had to ignore the template example in tests 😅

derberg (approved):

jonaslagoni:
@all-contributors please add @alequetzalli for review, doc

Issue #244, opened 2021-06-11
Object enums are not rendered correctly


jonaslagoni:
#### Describe the bug
It is possible to define object enums but the generators cannot handle them.

#### How to Reproduce
Given the following JSON Schema:
```
{
  "type": "object",
  "properties": {
    "prop1": {
      "type": ["string", "object"],
      "enum": [{"test": 2}]
    }
  }
}
```

In TS it generates:

```
export enum Prop1 {
  OBJECT_OBJECT = "[object Object]",
}
```

#### Expected behavior
Expected an enum to be generated that makes sense to what I defined.

Issue #242, opened 2021-06-11
Number enums are not rendered correctly


jonaslagoni:
#### Describe the bug
If an enum is defined as a number it is not rendered correctly.

#### How to Reproduce
Given the following JSON Schema:
```
{
  "type": "object",
  "properties": {
    "prop1": {
      "type": ["integer", "string"],
      "enum": [0, 2, "test"]
    }
  }
}
```

The enum is generated as:
```
export enum Prop1 {
  0 = 0,
  2 = 2,
  TEST = "test"
}
```
which violates `An enum member cannot have a numeric name.`

#### Expected behavior
Expected to transpile the generated class successfully and use something similar to the generated Java enum `NUMBER_2`.

Issue #336, opened 2021-08-31
C# enum generator does not generate associated values


jonaslagoni:
#### Reason/Context
We have an enum generator, that just does not generate the associated values (such as strings, numbers, booleans, and objects).
`{enums: ["test"]}` will for example generate:

```
public enum <somename> {
  Test
}
```

Which have no knowledge of the associated values.

jonaslagoni:
Scratch that, we have an enum generator, that just does not generate the correct values (such as strings and numbers).

PR #349, opened 2021-09-03
fix: enum generation should handle union types

jonaslagoni:
**Description**
This PR addresses some of the issues revolving around enum generation.

C# need its enum generation fixed beforehand: https://github.com/asyncapi/modelina/issues/336

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/244
Fixes https://github.com/asyncapi/modelina/issues/242

### DIFF
--- a/src/generators/csharp/renderers/EnumRenderer.ts
+++ b/src/generators/csharp/renderers/EnumRenderer.ts
@@ -42,7 +42,7 @@ export const CSHARP_DEFAULT_ENUM_PRESET: EnumPreset<EnumRenderer> = {
   },
   item({ item }) {
     let itemName = String(item);
-    if (typeof item === 'number') {
+    if (typeof item === 'number' || typeof item === 'bigint') {
       itemName = `Number_${itemName}`;
     } else if (typeof item === 'object') {
       itemName = `${JSON.stringify(item)}`;
--- a/src/generators/csharp/renderers/EnumRenderer.ts
+++ b/src/generators/csharp/renderers/EnumRenderer.ts
@@ -44,7 +44,10 @@ export const CSHARP_DEFAULT_ENUM_PRESET: EnumPreset<EnumRenderer> = {
     let itemName = String(item);
     if (typeof item === 'number') {
       itemName = `Number_${itemName}`;
+    } else if (typeof item === 'object') {
+      itemName = `${JSON.stringify(item)}`;
     }
+
     return pascalCase(itemName);
   },
 };
--- a/src/generators/java/renderers/EnumRenderer.ts
+++ b/src/generators/java/renderers/EnumRenderer.ts
@@ -53,7 +53,7 @@ ${this.indent(this.renderBlock(content, 2))}
       return `"${value}"`;
     }
     if (typeof value === 'object') {
-      return `"${JSON.stringify(value).replaceAll('"', '\\"')}"`;
+      return `"${JSON.stringify(value).replace(/"/g, '\\"')}"`;
     }
     return String(value);
   }
--- a/src/generators/java/renderers/EnumRenderer.ts
+++ b/src/generators/java/renderers/EnumRenderer.ts
@@ -41,6 +41,9 @@ ${this.indent(this.renderBlock(content, 2))}
     case 'boolean': {
       return FormatHelpers.toConstantCase(`boolean ${value}`);
     }
+    case 'object': {
+      return FormatHelpers.toConstantCase(JSON.stringify(value));
+    }
     default: return FormatHelpers.toConstantCase(String(value));
     }
   }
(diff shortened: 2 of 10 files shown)


src/generators/typescript/renderers/EnumRenderer.ts
```diff
@@ -36,13 +36,59 @@ ${this.indent(this.renderBlock(content, 2))}
...
+  }
+
+  normalizeValue(value: any): any {
+    let normalizedValue;
+    switch (typeof value) {
+    case 'string':
+      normalizedValue = `"${value}"`;
+      break;
+    case 'boolean':
+      normalizedValue = `"${value}"`;
```
magicmatatjahu, line 66:
```suggestion
      normalizedValue = `${value}`;
```

src/generators/typescript/renderers/EnumRenderer.ts
```diff
@@ -36,13 +36,59 @@ ${this.indent(this.renderBlock(content, 2))}
...
+      break;
+    case 'bigint':
+    case 'number': {
+      normalizedValue = value;
+      break;
+    }
+    case 'object': {
+      normalizedValue = `'${JSON.stringify(value)}'`;
```
magicmatatjahu, line 74: Should object be stringified or treated as normal JS value (object value)? 🤔

src/generators/java/renderers/EnumRenderer.ts
```diff
@@ -49,6 +52,9 @@ ${this.indent(this.renderBlock(content, 2))}
     if (typeof value === 'string') {
       return `"${value}"`;
     }
+    if (typeof value === 'object') {
+      return `"${JSON.stringify(value).replace(/"/g, '\\"')}"`;
```
magicmatatjahu, line 56: Do we need two `\\` as I guess, one is for `"` char and the second is for what?

test/generators/csharp/__snapshots__/CSharpGenerator.spec.ts.snap
```diff
@@ -202,13 +202,13 @@ exports[`CSharpGenerator should render \`enum\` type $name should not be empty 2
 
 exports[`CSharpGenerator should render \`enum\` type $name should not be empty 3`] = `
 "public enum Things {
-  Texas, Number_1, False
+  Texas, Number_1, False, ObjectObject
```
magicmatatjahu, line 205: Why `ObjectObject`? 😄

magicmatatjahu (changes requested):

src/generators/typescript/renderers/EnumRenderer.ts
jonaslagoni: Cant be treated as a normal object value, wont allow me to.. So have to do some custom marshalling code to handle this 🙃

src/generators/java/renderers/EnumRenderer.ts
jonaslagoni: `JSON.stringify` returns `{"test": "test"}` but we wrap it in `"` meaning they have to be escaped 🙂

src/generators/typescript/renderers/EnumRenderer.ts
jonaslagoni: We are not allowed to mix and match booleans in enums... So they have to be rendered as a string rather than the value, therefore the wrap of `"`.

test/generators/csharp/__snapshots__/CSharpGenerator.spec.ts.snap
jonaslagoni: Good catch, changed 👍

src/generators/typescript/renderers/EnumRenderer.ts
magicmatatjahu: I thought that in TS it's possible to use the boolean in enums, sorry 😅

jonaslagoni: Thought the same 😄

src/generators/java/renderers/EnumRenderer.ts
magicmatatjahu: ahh... the second is for the first `\` 👌🏼

src/generators/csharp/renderers/EnumRenderer.ts
```diff
@@ -44,7 +44,10 @@ export const CSHARP_DEFAULT_ENUM_PRESET: EnumPreset<EnumRenderer> = {
     let itemName = String(item);
     if (typeof item === 'number') {
```
magicmatatjahu, line 45:
```suggestion
    if (typeof item === 'number' || typeof item === 'bigint') {
```

magicmatatjahu (changes requested):

magicmatatjahu (approved):

Issue #160, opened 2021-04-27
library does not work with node version < 11


jonaslagoni:
#### Describe the bug
We are using `flatMap` function which is not added before [node version 11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#browser_compatibility).

https://github.com/asyncapi/generator-model-sdk/blob/f52a7701d7ca4b89aae1a16940b647bfb69b6bd6/src/helpers/FormatHelpers.ts#L59

#### What should change
We need to decide if we want to add the requirement to the readme file and package file or whether we should remove this function use. If we decide the later, to what point should we stop being backward compatible?

PR #350, opened 2021-09-03
fix: restrict node version to >= 14

jonaslagoni:
**Description**
Because we at the moment don't restrict which node version, it is up to interpretation and can be wrongfully used. We use version 14 for all our GitHub actions, so let's cap it to that.

**Related issue(s)**
fixes https://github.com/asyncapi/modelina/issues/160

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- toc -->
 
-- [Requirement](#requirement)
+- [Requirements](#requirements)
 - [Installation](#installation)
 - [How it works](#how-it-works)
   * [The input process](#the-input-process)
@@ -43,7 +43,7 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- tocstop -->
 
-## Requirement
+## Requirements
 
 - [NodeJS](https://nodejs.org/en/) >= 14
 
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- toc -->
 
+- [Requirement](#requirement)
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,6 +7,9 @@
   "bugs": {
     "url": "https://github.com/asyncapi/modelina/issues"
   },
+  "engines": {
+    "node": "14"
```
magicmatatjahu, line 11: You should use the `>=14` version. It perhaps hardcodes the given version to the 14.

README.md
```diff
@@ -42,6 +43,11 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- tocstop -->
 
+## Requirement
+- Because of different features depending on from node, we require version >= v14
```
magicmatatjahu, line 47:
```suggestion

- [NodeJS](https://nodejs.org/en/) >= 14
```

package.json
jonaslagoni: Ah, right 👍

magicmatatjahu (approved):

README.md
```diff
@@ -42,6 +43,12 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- tocstop -->
 
+## Requirement
```
magicmatatjahu, line 46:
```suggestion
## Requirements
```

README.md
```diff
@@ -27,6 +27,7 @@ This package is under development and it has not reached version 1.0.0 yet, whic
 
 <!-- toc -->
 
+- [Requirement](#requirement)
```
magicmatatjahu, line 30:
```suggestion
- [Requirements](#requirements)
```

magicmatatjahu (changes requested):

magicmatatjahu (approved):

Issue #247, opened 2021-06-14
Handle only the latest non-falsey content in presets


magicmatatjahu:
#### Reason/Context

Regarding to the discussion https://github.com/asyncapi/modelina/pull/239#discussion_r650933318 we should return in presets only latest non-falsey value. It means that if preset will return `undefined`, `null` or even `""`, next preset should handle the previous `content`.

We should change the line https://github.com/asyncapi/modelina/blob/master/src/generators/AbstractRenderer.ts#L50 to:

```ts
        content = await preset[String(functionName)]({ 
            ...params,           
            renderer: this,           
            content,           
            options,           
            model: this.model,           
            inputModel: this.inputModel        
        }) || content;
```

Issue #125, opened 2021-03-02
Bidirectional preset flow


magicmatatjahu:
At the moment we have only one way flow in presets. It means that render starts from `defaultPreset`, goes through every custom preset and return to output final string. Sometimes we need to adjust to final part of model something after rendering the rest of custom presets:

e.g. we have `CUSTOM_PRESET_1`, `CUSTOM_PRESET_2`, in `CUSTOM_PRESET_1` I wanna run first `CUSTOM_PRESET_2` and then add something to final string by `CUSTOM_PRESET_1`. Of course I can change order of `CUSTOM_PRESET_2` with `CUSTOM_PRESET_1` but some actions cannot be achieved by changing order - we wanna render some property, go through by every preset for property, but in final string, render something additional above the property... in this case we have problems:
- as a preset creator I don't know how many presets there are when rendering.
- sometimes subsequent presets may render the final model incorrectly - like here https://github.com/asyncapi/java-spring-template/pull/121/files#diff-812b0ef67b54ce5586eb2192840e7dbdb0c16608074f37f67a896fe733ff9436R110 - I must add  `@Valid` annotation to property which uses inline enum - by directional preset I can avoid repetitive boilerplate.
- related to the second point. Some presets can be completely overwritten and to run flow from beginning you need to run the preset (like `property`) in the another preset, but this can result in circular recursion.

Of course bidirectional preset flow doesn't fix all problem related to overwriting presets result - some preset must be render as last - but we can avoid some problems.

Maybe bidirectional preset flow isn't a good thing, how do you think about that? Maybe another suggestion to resolve above problems?

magicmatatjahu:
`next` function like in middleware system should helps. Example from the express framework -  https://expressjs.com/en/guide/using-middleware.html

PR #352, opened 2021-09-03
fix: run presets should be geared towards string output

jonaslagoni:
**Description**
This PR changes the default of the preset rendering, to use the latest non-falsey preset content.

**Related issue(s)**
fixes https://github.com/asyncapi/modelina/issues/247

### DIFF
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -71,7 +71,7 @@ export abstract class AbstractRenderer<
           model: this.model, 
           inputModel: this.inputModel
         });
-        if (presetRenderedContent === undefined || presetRenderedContent === null) {
+        if (presetRenderedContent === undefined || presetRenderedContent === null || typeof presetRenderedContent !== 'string') {
           content = '';
           continue;
         } 
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -56,10 +56,10 @@ export abstract class AbstractRenderer<
     return this.runPreset('additionalContent');
   }
   
-  async runPreset<RT = string>(
+  async runPreset(
     functionName: string,
     params: Record<string, unknown> = {},
-  ): Promise<RT> {
+  ): Promise<string> {
     let content = '';
     for (const [preset, options] of this.presets) {
       if (typeof preset[String(functionName)] === 'function') {
@@ -71,11 +71,11 @@ export abstract class AbstractRenderer<
           model: this.model, 
           inputModel: this.inputModel
         });
-        if (typeof presetRenderedContent === 'string' && presetRenderedContent !== undefined) {
-          content = presetRenderedContent;
+        if (presetRenderedContent === undefined || presetRenderedContent === null) {
+          content = '';
           continue;
-        }
-        content = '';
+        } 
+        content = presetRenderedContent;
       }
     }
     return content;
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -71,11 +71,11 @@ export abstract class AbstractRenderer<
           model: this.model, 
           inputModel: this.inputModel
         });
-        if (presetRenderedContent === undefined || presetRenderedContent === null || typeof presetRenderedContent !== 'string') {
+        if (typeof presetRenderedContent === 'string') {
+          content = presetRenderedContent;
+        } else {
           content = '';
-          continue;
-        } 
-        content = presetRenderedContent;
+        }
       }
     }
     return content;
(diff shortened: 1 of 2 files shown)


jonaslagoni:
Would there really not be a use-case where you want it to not render the previously rendered content? 🤔

src/generators/AbstractRenderer.ts
```diff
@@ -70,7 +70,7 @@ export abstract class AbstractRenderer<
           options, 
           model: this.model, 
           inputModel: this.inputModel
-        });
+        }) || content;
```
magicmatatjahu, line 73: What about if someone want to disable the given preset chain and return the empty string, what then?

jonaslagoni: That is the question isn't it (https://github.com/asyncapi/modelina/pull/352#issuecomment-912511672) 😄

magicmatatjahu: Question 😄  I think we should only do fallback for `null` and `undefined`, empty string should remove preset chain.

jonaslagoni: @magicmatatjahu are you sure we even need to handle this at all? 🤔

magicmatatjahu: Yes :)

jonaslagoni: @magicmatatjahu now that I think about it, null or undefined, should never be possible 🤔 https://github.com/asyncapi/modelina/blob/253462b302e67cd1f3021059f0c36569cdd4155e/src/models/Preset.ts#L31

magicmatatjahu: In TS, but in JS you can return the null or undefined. I think that we should check in the function https://github.com/asyncapi/modelina/blob/2f4f8631801f92ade0324f2178180237cd4edf41/src/generators/AbstractRenderer.ts#L59 if given output is string or not, and then save or not the result.

jonaslagoni: @magicmatatjahu right. 

I can't figure out if it makes more sense that `null` or `undefined` remove preset, or empty string 🤔 I am kinda leaning towards `null` or `undefined`

jonaslagoni: ... I dont know 😆

magicmatatjahu: Consider that presets act like middlewars in http frameworks. If a given middleware returns value without running the `next` function, then you break the middleware chain. It should be the same here - an empty string should also break the chain of presets. `null`, `undefined` and other non-string values we should treat as empty string.

jonaslagoni: So the following is what you are proposing, in terms of return values from presets?
- empty string: default back to previous rendered preset.
- null, undefined and other non-string values: render preset as an empty string;

magicmatatjahu: - null, undefined and other non-string values: render preset as an empty string.
- empty string should be also saved to the content as in present solution.

The 1st point is a part of the 2nd point.

magicmatatjahu: I think that here we have problem with direction of the presets flow. At the we have only a pure, one direction flow - from first to the last. Maybe we should consider make the preset flow as bi-directiional as described [in this issue](https://github.com/asyncapi/modelina/issues/125).

Why is this a problem? Because if someone want to disable the given preset must add additional custom preset to the end of the flow and return the empty string. Of course there is also another (more needed) use cases for the bi-directional presets Maybe we should switch to the logic like in native middleware with `next` function? @jonaslagoni What do you think?

jonaslagoni: I think that this PR is not about bi-directional presets 😄

I have no use-case where I see this as being useful at the moment, so I am just trying to address the associated issue 😄

src/generators/AbstractRenderer.ts
```diff
@@ -56,21 +56,26 @@ export abstract class AbstractRenderer<
...
           content, 
           options, 
           model: this.model, 
           inputModel: this.inputModel
         });
+        if (presetRenderedContent === undefined || presetRenderedContent === null || typeof presetRenderedContent !== 'string') {
+          content = '';
+          continue;
+        } 
+        content = presetRenderedContent;
```
magicmatatjahu, line 78:
```suggestion
        if (typeof presetRenderedContent === 'string') {
          content = presetRenderedContent;
        } else {
          content = '';
        }
```

magicmatatjahu (changes requested):

magicmatatjahu (approved):

PR #355, opened 2021-09-06
feat: csharp enum rendering associated values

jonaslagoni:
**Description**
This PR adds an extension to enums, which allows the retrieval of the associated enum values.

**Related issue(s)**
fixes https://github.com/asyncapi/modelina/issues/336

### DIFF
--- a/src/generators/csharp/renderers/EnumRenderer.ts
+++ b/src/generators/csharp/renderers/EnumRenderer.ts
@@ -11,20 +11,31 @@ export class EnumRenderer extends CSharpRenderer {
   async defaultSelf(): Promise<string> {
     const enumItems = await this.renderItems();
     const formattedName = this.nameType(this.model.$id);
-    const caseItemValues = await this.renderCaseItemValues();
+    const getValueCaseItemValues = await this.getValueCaseItemValues();
+    const toEnumCaseItemValues = await this.toEnumCaseItemValues();
     return `public enum ${formattedName} {
 ${this.indent(enumItems)}
 }
 public static class ${formattedName}Extensions {
-  public static dynamic GetValue(this ${formattedName} val)
+  public static dynamic GetValue(this ${formattedName} enumValue)
   {
-    switch (val)
+    switch (enumValue)
     {
-${this.indent(caseItemValues, 6)}
+${this.indent(getValueCaseItemValues, 6)}
+    }
+    return null;
+  }
+
+  public static ${formattedName}? To${formattedName}(dynamic value)
+  {
+    switch (value)
+    {
+${this.indent(toEnumCaseItemValues, 6)}
     }
     return null;
   }
 }
+
 `;
   }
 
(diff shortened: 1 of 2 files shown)


jonaslagoni:
These functions are primarily for #356, however, I feel they are needed in the standard enum generation 🤔 Let me know if you disagree!

test/generators/csharp/__snapshots__/CSharpGenerator.spec.ts.snap
```diff
@@ -191,25 +191,129 @@ exports[`CSharpGenerator should render \`class\` type 2`] = `
...
-}"
+}
+public static class StatesExtensions {
+  public static dynamic GetValue(this States enumValue)
+  {
+    switch (enumValue)
+    {
+      case States.Texas: return \\"Texas\\";
+      case States.Alabama: return \\"Alabama\\";
+      case States.California: return \\"California\\";
```
magicmatatjahu, line 202: I think that there is a bug. If something is a string, then it should be treated as native string in language, not as wrapped string with `""` characters. I don't know why, but it is rendered as object, not as string - please check also another test cases - you have this bug in every place.

magicmatatjahu (changes requested):

src/generators/csharp/renderers/EnumRenderer.ts
```diff
@@ -9,13 +9,34 @@ import { pascalCase } from 'change-case';
...
+    const enumItems = await this.renderItems();
     const formattedName = this.nameType(this.model.$id);
+    const getValueCaseItemValues = await this.getValueCaseItemValues();
+    const toEnumCaseItemValues = await this.toEnumCaseItemValues();
     return `public enum ${formattedName} {
-${this.indent(this.renderBlock(content, 2))}
-}`;
+${this.indent(enumItems)}
+}
+public static class ${formattedName}Extensions {
```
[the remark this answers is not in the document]
jonaslagoni: It is a pattern in C#, not changeable 😄

test/generators/csharp/__snapshots__/CSharpGenerator.spec.ts.snap
jonaslagoni: It is just the snapshot :) They are rendered as: 
```
      case States.Texas: return "Texas";
      case States.Alabama: return "Alabama";
      case States.California: return "California";
```
Which is correct.

src/generators/csharp/renderers/EnumRenderer.ts
```diff
@@ -9,13 +9,34 @@ import { pascalCase } from 'change-case';
...
+  public static dynamic GetValue(this ${formattedName} enumValue)
```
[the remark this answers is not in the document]
jonaslagoni: I know it is a weird format, but 🤷 - https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods

test/generators/csharp/__snapshots__/CSharpGenerator.spec.ts.snap
magicmatatjahu: and it's one of the cases, why I don't like snapshots 😄

jonaslagoni: You just gotta get used to it 😆

src/generators/csharp/renderers/EnumRenderer.ts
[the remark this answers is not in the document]
magicmatatjahu: Ok, it's something like trait.

magicmatatjahu (approved):

jonaslagoni:
@magicmatatjahu `fix` or `feat`? 🤔 Kinda thinking it should be `feat`

PR #366, opened 2021-09-08
feat: csharp equal and hashcode methods

jonaslagoni:
**Description**
This adds a new Common preset for C#, which includes equal and hashcode methods to the class.

### DIFF
--- a/test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
+++ b/test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
@@ -1,0 +1,147 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CSHARP_COMMON_PRESET should render Equals support function 1`] = `
+"public class Test {
+  private string stringProp;
+  private float? numberProp;
+  private NestedTest objectProp;
+  private Dictionary<string, dynamic> additionalProperties;
+  private Dictionary<string, string> sTestPatternProperties;
+
+  public string StringProp 
+  {
+    get { return stringProp; }
+    set { stringProp = value; }
+  }
+
+  public float? NumberProp 
+  {
+    get { return numberProp; }
+    set { numberProp = value; }
+  }
+
+  public NestedTest ObjectProp 
+  {
+    get { return objectProp; }
+    set { objectProp = value; }
+  }
+
+  public Dictionary<string, dynamic> AdditionalProperties 
+  {
+    get { return additionalProperties; }
+    set { additionalProperties = value; }
+  }
+
+  public Dictionary<string, string> STestPatternProperties 
+  {
+    get { return sTestPatternProperties; }
+    set { sTestPatternProperties = value; }
+  }
+
+  public override bool Equals(object obj)
+  {
+    return obj is Test model &&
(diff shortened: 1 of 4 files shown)


test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
```diff
@@ -0,0 +1,148 @@
...
+
+  public Dictionary<string, string> STestPatternProperties 
+  {
+    get { return sTestPatternProperties; }
+    set { sTestPatternProperties = value; }
+  }
+
+  public override bool Equals(object obj)
+  {
+    return obj is Test model &&
```
magicmatatjahu, line 43: This language is surprise me every time :D I should learn it.

magicmatatjahu: it seems to me that if it is true then the rest doesn't make sense, because it points to the same reference. 

You should make something similar like in JAVA: https://github.com/asyncapi/modelina/blob/master/test/generators/java/presets/CommonPreset.spec.ts#L28

so first check reference to object and then (if not be true) check the fields.

magicmatatjahu (changes requested):

test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
jonaslagoni: @magicmatatjahu hahaha, yea, this `obj is Test model` with auto-sign to a variable is as short as it can be 😆 

If the first condition fails the rest of the conditions are never checked 🙂 This is what Visual Studio auto-generates when you ask for it, so I dont see any issues here 😄

jonaslagoni: If you mean in terms of whether I can do `obj == model` then no, pretty sure it will use the equals method for that check, so not sure how or if that will work 😆

magicmatatjahu: aaa ok... it's casting... but I mean that if the reference to the `obj` is the same, then you should  return `true` without checking rest of things, like in Java - if these objects share this same reference, why to check the rest fields (it will be always true). Also (as in Java) you should check if the object this same, but reference is another - https://github.com/asyncapi/modelina/blob/master/test/generators/java/presets/CommonPreset.spec.ts#L35

jonaslagoni: Right, I actually thought `==` [automatically called](https://stackoverflow.com/questions/10109738/why-does-default-implementation-not-call-equals) `Equals`, but it checks references as you said, so gonna add that 👍 

I also need to test something, as I am not sure equals method works accurately for the dictionaries, enum and class checks 🧐

magicmatatjahu: > I also need to test something, as I am not sure equals method works accurately for the dictionaries, enum and class checks 🧐

For that probably checks the references, not exactly values (maybe only for Enums it checks values).

magicmatatjahu (approved):
LGTM

Issue #360, opened 2021-09-06
Swagger / OpenAPI input


voxpelli:
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
  - Swagger / OpenAPI is very similar to AsyncAPI and also builds upon JSON Schema style model definitions
- How will this change help?
  - Enable more types of conversions and a more complete toolchain for models of all types of sources.

#### Description
Please try answering few of those questions
- What changes have to be introduced?
  - Probably the same that was needed to support AsyncAPI on top of JSON Schema
- Will this be a breaking change?
  - Nope
- How could it be implemented/designed?
  - Probably the same way that AsyncAPI was

jonaslagoni:
Just leaving this for later:

- [OpenAPI 3.0.3](https://spec.openapis.org/oas/v3.0.3) - superset of [draft-wright-json-schema-00](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-00)
- [OpenAPI 3.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schemaObject) - superset of [2020-12 draft](https://tools.ietf.org/html/draft-bhutton-json-schema-00)
- [Swagger 2.0](https://swagger.io/specification/v2/) - superset of [JSON Schema Specification Draft 4](http://json-schema.org/specification-links.html#draft-4)

jonaslagoni:
@voxpelli do you by any chance have any Swagger/OpenAPI documents you can share, that I can test with? 🙂 

Also, do you mind if I ping you for giving a review once the features are complete?

voxpelli:
@jonaslagoni Unfortunately nothing public, but I do remember the file I was looking at this time as being a Swagger 2.0 file, so #378 would solve it.

Please ping me when you want a review!

jonaslagoni:
@voxpelli the PR is ready from our side to be merged, if you want to give your review, now is the time 😄 Let me know if you don't have time to review it, so we don't wait in vain 👍

I suggest to download the branch and adapt the example to fit your need - https://github.com/jonaslagoni/generator-model-sdk/tree/feature/swagger_new/examples/swagger2.0-from-object

voxpelli:
Still relevant :)

voxpelli:
Still relevant :)

AnimeshKumar923:
still valid? @jonaslagoni @voxpelli

voxpelli:
I guess this still holds true:

[...]

PR #372, opened 2021-09-14
feat: support swagger 2.0 and JSON Schema draft 4 inputs

### LINKED ISSUE #86: originalSchema for CommonModel should be called originalInput
#### Reason/Context
The reason why we should make this change is that the input does not necessarily have be JSON Schema but could be something completely unrelated.

jonaslagoni:
**Description**
This PR refactors some of the core input processor code and adds new schema models for draft 4 and swagger 2.0 schema.

One question you might have looking at this, "why not use already pre-defined types for these schemas?", well they are faulty and unprecise in most cases. This way we are sure on the content on the schemas files.

Related document for seeing which schema properties are related:
spec schema sharing drawio

**Related issue(s)**
Related to #360 
Fixes #86


jonaslagoni:
Split out PR into 3:
- https://github.com/asyncapi/modelina/pull/377
- https://github.com/asyncapi/modelina/pull/378
- https://github.com/asyncapi/modelina/pull/379

Issue #365, opened 2021-09-08
Improve reuseability and preset developer experience


jonaslagoni:
### Reason/Context
This is one big braindump, that will continue to be iterated until the CommonModel enables a better behavior in the following use-cases.

### Issues

- **You need to know which support functions you need to call to render something like types**
- **Iterating pattern properties are cumbersome and require knowledge of what the name for the property is.**

Currently, many languages uses the following syntax to iterate pattern properties that may be present:
```ts
for (const [pattern, patternModel] of Object.entries(model.patternProperties)) {
  let patternPropertyName = getUniquePropertyName(model, `${pattern}${DefaultPropertyNames.patternProperties}`);
  patternPropertyName = renderer.nameProperty(patternPropertyName, patternModel);
```
- **Additional properties property are hard to access**

Somewhat the same ordeal as pattern properties.
```ts
if (model.additionalProperties !== undefined) {
  let additionalPropertyName = getUniquePropertyName(model, DefaultPropertyNames.additionalProperties);
  additionalPropertyName = FormatHelpers.upperFirst(renderer.nameProperty(additionalPropertyName, model.additionalProperties));
```
- **Tight coupling with JSON Schema draft 7**

The CommonModel is tightly coupled with JSON Schema draft 7 as we wanted to see how it all played out with the interpretation of schema files, as well as other inputs before making a change. 

We need to remember that different inputs have different effects on the output. For example to enable a property for `additionalProperties` in the model from a JSON Schema draft 7 input, for serialization, it needs to NOT be rendered as JSON property, but unwrapped. 

- **Iterating all properties for a model is though**

This goes back to the 3 aforementioned issues, we have no way to say `for (const [propertyName, propertyModel] of model.properties) {` and then receive all the expected properties.

- **Figuring out a referenced model's type**

To access any referenced models you need to have access to the `CommonInputModel` and then look it up, for then to determine the type of referenced model, such as in this case, where we need to do something specific for referenced Enums.

```ts
  if (model.$ref) {
    const resolvedModel = inputModel.models[model.$ref];
    const propertyModelKind = TypeHelpers.extractKind(resolvedModel);
    //Referenced enums is the only one who need custom serialization
    if (propertyModelKind === ModelKind.ENUM) {
      value = `${value}.GetValue()`;
    }
  }
```

- **Figuring out a property's type that should be generated can be cluttered**

[...]

PR #377, opened 2021-09-17
refactor: changing input processor and interpreter to support more schemas

### LINKED ISSUE #86: originalSchema for CommonModel should be called originalInput

jonaslagoni:
**Description**
As more input schemas are to be supported, we need to split out the schemas even more than we are now. 

The changes:
- It removes the generic `Schema` and `CommonSchema` completely and introduces the more specific `Draft7Schema` and `AsyncapiV2Schema`.
- It changes the `originalSchema` property from the `CommonModel` to `originalInput` - see #86
- Refactored the `toX` functions for `Draft7Schema` and `CommonModel` to be more generic, I did not find any way to generalize it anymore between them, so if you have any suggestion, let me know 🙏 

**Related issue(s)**
Related to #360 
Fixes #86


jonaslagoni:
@magicmatatjahu we can ignore the bug as #379 removes it.

src/models/AsyncapiV2Schema.ts
```diff
@@ -0,0 +1,135 @@
...
+    throw new Error('Could not convert input to expected copy of AsyncapiV2Schema');
+  }
+  private static internalToSchema(object: any, seenSchemas: Map<any, AsyncapiV2Schema> = new Map()): any {
+    // if primitive types return as is
+    if (null === object || 'object' !== typeof object) {
+      return object;
+    }
+
+    if (seenSchemas.has(object)) {
+      return seenSchemas.get(object) as any;
```
magicmatatjahu, line 103:
```suggestion
      return seenSchemas.get(object);
```

src/processors/JsonSchemaInputProcessor.ts
```diff
@@ -49,144 +53,155 @@ export class JsonSchemaInputProcessor extends AbstractInputProcessor {
...
         occurrence++;
       }
       const inferredName = occurrence ? `${name}_${occurrence}` : name;
-      schema[this.MODELGEN_INFFERED_NAME] = inferredName;
+      (schema as any)[this.MODELGEN_INFFERED_NAME] = inferredName;
     }
 
     if (schema.allOf !== undefined) {
-      schema.allOf = schema.allOf.map((item, idx) => this.reflectSchemaNames(item, namesStack, this.ensureNamePattern(name, 'allOf', idx)));
+      schema.allOf = (schema.allOf as any[]).map((item: any, idx: number) => JsonSchemaInputProcessor.reflectSchemaNames(item, namesStack, this.ensureNamePattern(name, 'allOf', idx)));
```
magicmatatjahu, line 125: You can use `this` keyword for static methods/properties in static methods/properties, so why you changed it to class reference?

jonaslagoni: Thaaats a good question, I think it was because I tried different restructures first and then this happened 😄 Changed!

src/models/AsyncapiV2Schema.ts
jonaslagoni: Changed

magicmatatjahu (commented):
@jonaslagoni I understand that we are still sticking with CommonModel etc? Will this be changed in the future?

> Refactored the toX functions for Draft7Schema and CommonModel to be more generic, I did not find any way to generalize it anymore between them, so if you have any suggestion, let me know 🙏

It seems like too much boilerplate to me, but it would take a lot of work to change it and that's the easiest way to do it now...  😅 

Also, you have one bug in sonar cloud.

jonaslagoni:
> @jonaslagoni I understand that we are still sticking with CommonModel etc? Will this be changed in the future?

For now, however I have started to rework that part, but it is a WIP, as I still need test different things before making some noise around it - https://github.com/asyncapi/modelina/issues/365#issuecomment-922461877

> It seems like too much boilerplate to me, but it would take a lot of work to change it and that's the easiest way to do it now... 😅

Yea, exactly, dont have any better alternative at the moment. At least I dont see it.

magicmatatjahu (approved):
LGTM :)

PR #378, opened 2021-09-17
feat: support Swagger 2.0 as input

jonaslagoni:
**Description**
This PR introduces the new input processor for Swagger 2.0 documents.

**Related issue(s)**
Partly solves #360


jonaslagoni:
Blocked by https://github.com/asyncapi/modelina/pull/377

docs/usage.md
```diff
@@ -25,20 +25,36 @@ TODO
...
+## Generate models from Swagger 2.0 documents
+When providing an AsyncAPI document, Modelina iterates the entire document and generate models for all defined body parameters and responses. If any other kind of iteration is wanted, feel free to create a [feature request](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md).
+
+There are one way to generate models from a Swagger 2.0 document
+
+- [Generate from a pure JS object](../examples/swagger2.0-from-object)
+
+The Swagger input processor expects that the property `swagger` is defined in order to know it should be processed.
+
+The response payload and parameters, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
```
magicmatatjahu, line 56:
```suggestion
The response payload and `body` parameters, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
```

docs/usage.md
```diff
@@ -25,20 +25,36 @@ TODO
...
 
 There are one way to generate models from a JSON Schema draft 7 document.
 
 - [Generate from a pure JS object](../examples/json-schema-draft7-from-object)
 
-The library expects the `$schema` property for the document to be set in order to understand the input format. By default, if no other inputs are detected, it defaults to `JSON Schema draft 7`.
+The library expects the `$schema` property for the document to be set in order to understand the input format. By default, if no other inputs are detected, it defaults to `JSON Schema draft 7`. The process of interpreting a JSON Schema to a model can be read [here](./interpretation_of_JSON_Schema.md).
+
+## Generate models from Swagger 2.0 documents
+When providing an AsyncAPI document, Modelina iterates the entire document and generate models for all defined body parameters and responses. If any other kind of iteration is wanted, feel free to create a [feature request](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md).
```
magicmatatjahu, line 48:
```suggestion
When providing an AsyncAPI document, Modelina iterates the entire document and generate models for all defined `body` parameters and responses. If any other kind of iteration is wanted, feel free to create a [feature request](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md).
```

src/processors/InputProcessor.ts
```diff
@@ -12,6 +13,7 @@ export class InputProcessor {
 
   constructor() {
     this.setProcessor('asyncapi', new AsyncAPIInputProcessor()); 
+    this.setProcessor('swagger', new SwaggerInputProcessor());
```
magicmatatjahu, line 16: I wonder if we need to create a swagger processor by default. This is such a rare case that the user should create it only if he/she wants to - why is this so important to me? For this same problem like we have with `fs` package in another PR - for tree-shaking. Whenever someone wants to use the generation just for AsyncAPI or plain JSON Schema, they will add dependencies for swagger to the resulting output bundle (case for Modelina playground on the website and in the future for the Studio) - which isn't necessary and only add additional, not used code to bundle. What do you think?

src/processors/SwaggerInputProcessor.ts
```diff
@@ -0,0 +1,109 @@
...
+      //Remove any pre-pending '/'
+      formattedPathName = formattedPathName.replace(/\//, '');
+      //Replace all segment separators '/'
+      formattedPathName = formattedPathName.replace(/\//gm, '_');
+      this.processOperation(pathObject.get, `${formattedPathName}_get`, common);
+      this.processOperation(pathObject.put, `${formattedPathName}_put`, common);
+      this.processOperation(pathObject.post, `${formattedPathName}_post`, common);
+      this.processOperation(pathObject.options, `${formattedPathName}_options`, common);
+      this.processOperation(pathObject.head, `${formattedPathName}_head`, common);
+      this.processOperation(pathObject.patch, `${formattedPathName}_patch`, common);
```
magicmatatjahu, line 40: I don't know so I ask. Does Swagger also have something like components, where you can have reusable definitions for parameters and response?

magicmatatjahu: Also, maybe you should use as fallback the `id` field for model name?

magicmatatjahu (commented):
These are mostly comments for starting the discussion, so I don't give a reject. Btw. awesome work with new processor :)

src/processors/InputProcessor.ts
jonaslagoni: Hmm, I definitely see the use case, however, when would you want to see this being addressed, now, or later?

Cause I assume we would need to split up some things for this to work, or is it painless? 🤔

magicmatatjahu: We can later. Probably when we will integrate the Modelina in the Studio (not sure when 😆 )

src/processors/SwaggerInputProcessor.ts
jonaslagoni: They have something called `definitions` rather than `components`. Do you think we should include such a "complex" id reflection here? 🤔 

I was thinking it was best with a basic implementation, at least for now.

magicmatatjahu: > I was thinking it was best with a basic implementation, at least for now.

If you wanna go with next PRs, current solution is good for me :)

jonaslagoni: @magicmatatjahu if the swagger document has lonely definitions (that are not referenced anywhere), the data models will not be generated - and I am not planning to add any PR after this one, this PR is more a primer to include more users that can use it.

magicmatatjahu (approved):
For me good. @alequetzalli Do you wanna make review? :)

jonaslagoni:
In order to not block any further let's merge the PR, @voxpelli if you find any problems etc, please create an issue 🙂

Issue #396, opened 2021-09-25
Generating models to files directly for Java


jonaslagoni:
#### Reason/Context
As a precursor to #172 we need to give the user the option to generate models directly to files.

This is related to #288 as the full output is needed to write it to a file.

Issue #288, opened 2021-07-05
Generate full output


jonaslagoni:
#### Reason/Context
At the moment the default behavior of the generator is to return the result as a segmented output. i.e. it returns the model as a bare minimum where you need to "look" for dependencies and for java add your own package.

We need to ensure that there is an option to generate everything together to allow as low customization as possible to use the library as is.

jonaslagoni:
Still being worked on.

PR #401, opened 2021-09-25
feat: generate Java models to files and full output

jonaslagoni:
**Description**
This PR enables the user to generate Java models directly to files as well as full output.
 
**Related issue(s)**
Partly solves #288 
Fixes https://github.com/asyncapi/modelina/issues/396

### DIFF
--- a/src/generators/java/JavaGenerator.ts
+++ b/src/generators/java/JavaGenerator.ts
@@ -37,34 +37,6 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
     return this.renderClass(model, inputModel);
   }
 
-  /**
-   * Generates all the models to the same output directory. 
-   * 
-   * This function is invasive, as it overwrite any existing files with the same name as the model.
-   * 
-   * @param input
-   * @param outputDirectory where you want the models generated to
-   * @param packageName for the models
-   */
-  public async generateToFile(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, packageName: string): Promise<OutputModel[]> {
-    if (isReservedJavaKeyword(packageName)) {
-      return Promise.reject(`You cannot use reserved java keyword (${packageName}) as package name, please use another.`);
-    }
-
-    const generatedModels = await this.generate(input);
-    for (const outputModel of generatedModels) {
-      const outputFilePath = path.resolve(outputDirectory, `${FormatHelpers.toPascalCase(outputModel.model.$id || 'undefined')}.java`);
-      const modelDependencies = outputModel.model.getNearestDependencies().map((dependencyModel) => { return `import ${packageName}.${FormatHelpers.toPascalCase(dependencyModel || 'undefined')};`;});
-      const outputContent = `package ${packageName};
-${modelDependencies.join('\n')}
-${outputModel.dependencies.join('\n')}
-${outputModel.result}
-`;
-      await FileHelpers.writeToFile(outputContent, outputFilePath);
-    }
-    return Promise.resolve(generatedModels);
-  }
-
   async renderClass(model: CommonModel, inputModel: CommonInputModel): Promise<RenderOutput> {
     const presets = this.getPresets('class');
     const renderer = new ClassRenderer(this.options, this, presets, model, inputModel);
@@ -77,5 +49,51 @@ ${outputModel.result}
     const renderer = new EnumRenderer(this.options, this, presets, model, inputModel);
     const result = await renderer.runSelfPreset();
     return RenderOutput.toRenderOutput({result, dependencies: renderer.dependencies});
+  }
+
+  /**
+   * Generates all the models to an output directory. 
+   * 
+   * This function is invasive, as it overwrite any existing files with the same name as the model.
+   * 
+   * @param input
+   * @param outputDirectory where you want the models generated to
+   * @param packageName for the models
+   */
+  public async generateToFile(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, packageName: string): Promise<OutputModel[]> {
+    const generatedModels = await this.generateFullOutput(input, packageName);
+    for (const outputModel of generatedModels) {
+      const outputFilePath = path.resolve(outputDirectory, `${FormatHelpers.toPascalCase(outputModel.model.$id || 'undefined')}.java`);
+      await FileHelpers.writeToFile(outputModel.result, outputFilePath);
+    }
+    return Promise.resolve(generatedModels);
+  }
+
+  /**
+   * Generates the full output of a model, instead of a scattered model.
+   * 
+   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
+   * 
+   * @param input 
+   * @param packageName for the models
+   */
+  public async generateFullOutput(input: Record<string, unknown> | CommonInputModel, packageName: string): Promise<OutputModel[]> {
+    if (isReservedJavaKeyword(packageName)) {
+      return Promise.reject(`You cannot use reserved Java keyword (${packageName}) as package name, please use another.`);
+    }
+
+    const fullOutputModels = [];
+    const generatedModels = await this.generate(input);
+    for (const outputModel of generatedModels) {
+      const modelDependencies = outputModel.model.getNearestDependencies().map((dependencyModel) => { return `import ${packageName}.${FormatHelpers.toPascalCase(dependencyModel || 'undefined')};`;});
+      const outputContent = `package ${packageName};
+${modelDependencies.join('\n')}
+${outputModel.dependencies.join('\n')}
+${outputModel.result}
+`; 
+      const newOutputModel = new OutputModel(outputContent, outputModel.model, outputModel.modelName, outputModel.inputModel, outputModel.dependencies);
+      fullOutputModels.push(newOutputModel);
+    }
+    return Promise.resolve(fullOutputModels);
   }
 }
--- a/src/generators/java/JavaGenerator.ts
+++ b/src/generators/java/JavaGenerator.ts
@@ -82,18 +82,14 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
       return Promise.reject(`You cannot use reserved Java keyword (${packageName}) as package name, please use another.`);
     }
 
-    const fullOutputModels = [];
     const generatedModels = await this.generate(input);
-    for (const outputModel of generatedModels) {
+    return Promise.resolve(generatedModels.map((outputModel) => {
       const modelDependencies = outputModel.model.getNearestDependencies().map((dependencyModel) => { return `import ${packageName}.${FormatHelpers.toPascalCase(dependencyModel || 'undefined')};`;});
       const outputContent = `package ${packageName};
 ${modelDependencies.join('\n')}
 ${outputModel.dependencies.join('\n')}
-${outputModel.result}
-`; 
-      const newOutputModel = new OutputModel(outputContent, outputModel.model, outputModel.modelName, outputModel.inputModel, outputModel.dependencies);
-      fullOutputModels.push(newOutputModel);
-    }
-    return Promise.resolve(fullOutputModels);
+${outputModel.result}`; 
+      return new OutputModel(outputContent, outputModel.model, outputModel.modelName, outputModel.inputModel, outputModel.dependencies);
+    }));
   }
 }
--- a/src/generators/java/JavaGenerator.ts
+++ b/src/generators/java/JavaGenerator.ts
@@ -3,11 +3,13 @@ import {
   CommonGeneratorOptions,
   defaultGeneratorOptions,
 } from '../AbstractGenerator';
-import { CommonModel, CommonInputModel, RenderOutput } from '../../models';
-import { CommonNamingConvention, CommonNamingConventionImplementation, ModelKind, TypeHelpers } from '../../helpers';
+import { CommonModel, CommonInputModel, RenderOutput, OutputModel } from '../../models';
+import { CommonNamingConvention, CommonNamingConventionImplementation, ModelKind, TypeHelpers, FileHelpers, FormatHelpers } from '../../helpers';
 import { JavaPreset, JAVA_DEFAULT_PRESET } from './JavaPreset';
 import { ClassRenderer } from './renderers/ClassRenderer';
 import { EnumRenderer } from './renderers/EnumRenderer';
+import * as path from 'path';
+import { isReservedJavaKeyword } from './Constants';
 export interface JavaOptions extends CommonGeneratorOptions<JavaPreset> {
   collectionType?: 'List' | 'Array';
   namingConvention?: CommonNamingConvention;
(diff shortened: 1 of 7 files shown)


src/generators/java/JavaGenerator.ts
```diff
@@ -3,11 +3,13 @@ import {
...
   defaultGeneratorOptions,
 } from '../AbstractGenerator';
-import { CommonModel, CommonInputModel, RenderOutput } from '../../models';
-import { CommonNamingConvention, CommonNamingConventionImplementation, ModelKind, TypeHelpers } from '../../helpers';
+import { CommonModel, CommonInputModel, RenderOutput, OutputModel } from '../../models';
+import { CommonNamingConvention, CommonNamingConventionImplementation, ModelKind, TypeHelpers, FileHelpers, FormatHelpers } from '../../helpers';
 import { JavaPreset, JAVA_DEFAULT_PRESET } from './JavaPreset';
 import { ClassRenderer } from './renderers/ClassRenderer';
 import { EnumRenderer } from './renderers/EnumRenderer';
+import * as path from 'path';
```
magicmatatjahu, line 11: Did you try to generate the models on the website with this PR? I don't know how NextJS treats NodeJS related packages in the browser, but we should check it before merge to avoid broken rendering.

src/generators/java/JavaGenerator.ts
```diff
@@ -48,4 +50,46 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
...
+   * @param outputDirectory where you want the models generated to
+   * @param packageName for the models
+   */
+  public async generateToFile(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, packageName: string): Promise<OutputModel[]> {
+    const generatedModels = await this.generateFullOutput(input, packageName);
+    for (const outputModel of generatedModels) {
+      const outputFilePath = path.resolve(outputDirectory, `${FormatHelpers.toPascalCase(outputModel.model.$id || 'undefined')}.java`);
+      await FileHelpers.writeToFile(outputModel.result, outputFilePath);
+    }
+    return Promise.resolve(generatedModels);
```
magicmatatjahu, line 69: Why do you call `Promise.resolve` even function is async? You can just return `generatedModels` without wrapping with Promise.

src/generators/java/JavaGenerator.ts
```diff
@@ -48,4 +50,46 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
...
+   * @param input 
+   * @param packageName for the models
+   */
+  public async generateFullOutput(input: Record<string, unknown> | CommonInputModel, packageName: string): Promise<OutputModel[]> {
+    if (isReservedJavaKeyword(packageName)) {
+      return Promise.reject(`You cannot use reserved Java keyword (${packageName}) as package name, please use another.`);
+    }
+
+    const generatedModels = await this.generate(input);
+    return Promise.resolve(generatedModels.map((outputModel) => {
```
magicmatatjahu, line 86: Also here. Why do you call Promise.resolve even function is async?

src/generators/java/JavaGenerator.ts
```diff
@@ -48,4 +50,46 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
...
+   * Generates the full output of a model, instead of a scattered model.
+   * 
+   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
+   * 
+   * @param input 
+   * @param packageName for the models
+   */
+  public async generateFullOutput(input: Record<string, unknown> | CommonInputModel, packageName: string): Promise<OutputModel[]> {
+    if (isReservedJavaKeyword(packageName)) {
+      return Promise.reject(`You cannot use reserved Java keyword (${packageName}) as package name, please use another.`);
```
magicmatatjahu, line 82: You can throw native error, without wrapping it with Promise - function is async.

src/helpers/FileHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+    const outputFilePath = path.resolve(outputFile);
+    try {
+      // eslint-disable-next-line security/detect-non-literal-fs-filename
+      await fs.mkdir(path.dirname(outputFilePath), { recursive: true });
+      // eslint-disable-next-line security/detect-non-literal-fs-filename
+      await fs.writeFile(outputFilePath, content);
+    } catch (e: any) {
+      return Promise.reject(e);
+    }
+    return Promise.resolve();
```
magicmatatjahu, line 15: It's unnecessary. See comments above.

src/helpers/FileHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+export class FileHelpers {
+  static async writeToFile(content: string, outputFile: string): Promise<void> {
+    const outputFilePath = path.resolve(outputFile);
+    try {
+      // eslint-disable-next-line security/detect-non-literal-fs-filename
+      await fs.mkdir(path.dirname(outputFilePath), { recursive: true });
+      // eslint-disable-next-line security/detect-non-literal-fs-filename
+      await fs.writeFile(outputFilePath, content);
+    } catch (e: any) {
+      return Promise.reject(e);
```
magicmatatjahu, line 13: Just rethrow error.

src/generators/java/JavaGenerator.ts
```diff
@@ -48,4 +50,46 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
...
+    }
+
+    const generatedModels = await this.generate(input);
+    return Promise.resolve(generatedModels.map((outputModel) => {
+      const modelDependencies = outputModel.model.getNearestDependencies().map((dependencyModel) => { return `import ${packageName}.${FormatHelpers.toPascalCase(dependencyModel || 'undefined')};`;});
```
magicmatatjahu, line 87: The `undefined` fallback can produces something like:

```java
import somePackage.Undefined
```

and I don't think so that it's good solution. Please give me one case where such a situation will be possible. If not, please filter undefined models.

src/generators/java/JavaGenerator.ts
```diff
@@ -48,4 +50,46 @@ export class JavaGenerator extends AbstractGenerator<JavaOptions> {
...
+   * This function is invasive, as it overwrite any existing files with the same name as the model.
+   * 
+   * @param input
+   * @param outputDirectory where you want the models generated to
+   * @param packageName for the models
+   */
+  public async generateToFile(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, packageName: string): Promise<OutputModel[]> {
+    const generatedModels = await this.generateFullOutput(input, packageName);
+    for (const outputModel of generatedModels) {
+      const outputFilePath = path.resolve(outputDirectory, `${FormatHelpers.toPascalCase(outputModel.model.$id || 'undefined')}.java`);
```
magicmatatjahu, line 66: Like in another comment - it can produce something like `undefined.java` and you can end with several undefined models, so please give me one case where such a situation will be possible. If not, please filter undefined models.

magicmatatjahu (changes requested):
In addition to the code changes, I have a problem with the functions placement itself.

I think that `generateToFile` (aka `generateFile`) function should be written in the `AbstractGenerator` class and should use some abstract function (like in your case `generateFullOutput`, but I think that `renderFile` is better name), similar to `generate` and `render` functions:
- `generateFile` is place in the `AbstractGenerator`
- `renderFile` is implemented in each generator and `generateFile` uses it

src/generators/java/JavaGenerator.ts
jonaslagoni: `path` is okay, `fs` however, is a problem yes. Solvable in some cases - https://maikelveen.com/blog/how-to-solve-module-not-found-cant-resolve-fs-in-nextjs

Not sure where we want to draw the line? Should we just include `generateFullOutput` and don't help out with file creation? 🤔

src/generators/java/JavaGenerator.ts
jonaslagoni: Not sure why, but had some problems with it, so just opted for a more explicit resolvement 😅 Solved!

src/generators/java/JavaGenerator.ts
magicmatatjahu: The idea with only `generateFullOutput` is nice. Hmm 🤔  Or maybe create separate class like `JavaFileGenerator` where you will have such a logic to generate and save to the files? Then everything will be tree-shakable, and NextJS should remove unused code in the website and other front-end tools, what do you think?

jonaslagoni: Hmmm... Not sure whether the user experience is hurt by that. In most cases, I would not want to manually have to choose between the two 🤔 

What if we created a file interface that has to be provided to the function, so we dont include the `fs` module? 

i.e. we do something like:

```
export interface FileWriter {
  write(content: string, toFile: string): Promise<void>; 
}
  public async generateFile(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, options: JavaRenderFullOptions, fileWriter: FileWriter): Promise<OutputModel[]> {
    let generatedModels = await this.generateFull(input, options);
    generatedModels = generatedModels.filter((outputModel) => { return outputModel.model.$id !== undefined; });
    for (const outputModel of generatedModels) {
      const filePath = path.resolve(outputDirectory, `${outputModel.modelName}.java`);
      await fileWriter.write(outputModel.result, filePath);
    }
    return generatedModels;
  }
```

Thoughts?

magicmatatjahu: But then someone has to implement that interface. With the new class I meant rather to create helper class which should have as argument to the constructor base generator of given language, so you will end with something like:

```ts
class JavaFileGenerator {
  constructor(
    private generator: JavaGenerator
  ) {}
  
  generateFile() { .... }
  ...
}
```

and then you have another module in which this fs would be separated and could be easily removed

magicmatatjahu: @jonaslagoni It's an another option, but then people will write such an code:

```js
async (content, file) => {await fs.promises.writeFile(file, content);}
```

every time. For me it makes more boilerplate than the second extra class :)

Also from our conversation in slack (https://asyncapi.slack.com/archives/CQVJXFNQL/p1633351369306600) you gave me one idea - we can create `FileGenerator` class for every generators, but it will work in this way that appropriate generator (e.g. JavaGenerator) will produce by some function (generateToFiles) array with context and `FileGenerator` will only render content to file and save it. What do you think? Do you understand my point of view?

jonaslagoni: I decided to go with the inheritance option. Did not find a specific reason why you would need to have access to two generators 😄 

i.e. you use:

```ts
const generator = new JavaFileGenerator();
const outputFolder = './examples/generate-to-files/output';
const modelGenerationOptions = {
  packageName: 'TestPackageName'
};
const models = await generator.generateToSeparateFiles(jsonSchemaDraft7, outputFolder, modelGenerationOptions);
```

I really like this approach 😄

src/generators/AbstractGenerator.ts
```diff
@@ -34,28 +34,58 @@ export abstract class AbstractGenerator<Options extends CommonGeneratorOptions =
...
-    return this.generateModels(model);
+  /**
+   * Generates the full output of a model, instead of a scattered model.
+   * 
+   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
+   * 
+   * @param input 
+   * @param options to use for rendering full output
+   */
+  public async generateFull(input: Record<string, unknown> | CommonInputModel, options: RenderFullOptions): Promise<OutputModel[]> {
```
jonaslagoni, line 51: I still can't figure out if this is the `right` name for the function 🤷

docs/advanced.md
```diff
@@ -22,8 +22,20 @@ This document contains many of the advanced use-cases that you may stumble upon
 ## Generate each model in the same file
 TODO 
 
-## Generate a model in separate files
-TODO 
+## Generate models to separate files
+
+The standard generator only allows you to generate the raw models which you can implement your own logic for generating the models to separate files. We have however create simple wrapper generators to use.
+
+The reason for splitting the functionality is because in certain environments, generating to a file is not needed.
```
magicmatatjahu, line 29:
```suggestion
The reason for splitting the functionality is because in certain environments (like pure front-end application), generating to a file is not needed.
```

src/generators/AbstractFileGenerator.ts
```diff
@@ -0,0 +1,10 @@
+import { CommonInputModel, OutputModel } from '../models';
+
+export type FileGenerator = (content: string, toFile: string) => Promise<void>; 
+
+/**
+ * Abstract file generator which each file generator much implement.
+ */
+export interface AbstractFileGenerator<RenderFullOptions> {
+  generateToSeparateFiles(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, options: RenderFullOptions): Promise<OutputModel[]>;
```
magicmatatjahu, line 9:
```suggestion
  generateToFiles(input: Record<string, unknown> | CommonInputModel, outputDirectory: string, options: RenderFullOptions): Promise<OutputModel[]>;
```

I think that `Seperate ` is not needed :)

src/generators/AbstractGenerator.ts
```diff
@@ -34,28 +34,58 @@ export abstract class AbstractGenerator<Options extends CommonGeneratorOptions =
   }
 
   public abstract render(model: CommonModel, inputModel: CommonInputModel): Promise<RenderOutput>;
+  public abstract renderFull(model: CommonModel, inputModel: CommonInputModel, options: RenderFullOptions): Promise<RenderOutput>;
```
magicmatatjahu, line 37:
```suggestion
  public abstract renderFileContent(model: CommonModel, inputModel: CommonInputModel, options: RenderFullOptions): Promise<RenderOutput>;
```

It should say more that `renderFull`.

src/generators/AbstractGenerator.ts
magicmatatjahu: Maybe `generateFullContent`?

src/generators/AbstractGenerator.ts
```diff
@@ -34,28 +34,58 @@ export abstract class AbstractGenerator<Options extends CommonGeneratorOptions =
...
+   * @param input 
+   */
+  private async processInput(input: Record<string, unknown> | CommonInputModel): Promise<CommonInputModel> {
+    let inputModel: CommonInputModel;
+    if (input instanceof CommonInputModel) {
+      inputModel = input;
+    } else {
+      inputModel = await this.process(input);
+    }
+    return inputModel;
```
magicmatatjahu, line 86:
```suggestion
    if (input instanceof CommonInputModel) {
      return input;
    }
    return this.process(input);
```

src/generators/AbstractFileGenerator.ts
magicmatatjahu: Or only `generateFiles` :)

magicmatatjahu (commented):
Awesome! I also very like this idea, but I have some suggestions, only related to function names :)

src/generators/AbstractGenerator.ts
jonaslagoni: see https://github.com/asyncapi/modelina/pull/401#discussion_r723278001

src/generators/AbstractGenerator.ts
magicmatatjahu: Ok, it can be :)

package.json
```diff
@@ -27,6 +27,7 @@
   "main": "lib/cjs/index.js",
   "module": "lib/esm/index.js",
   "types": "lib/types/index.d.ts",
+  "sideEffects": ["./src/generators/*FileGenerator.ts", "./src/helpers/FileHelpers.ts"],
```
magicmatatjahu, line 30: We read in webpack docs:

> All the code noted above does not contain side effects, so we can mark the property as false to inform webpack that it can safely prune unused exports.

so, we should change it to:

```suggestion
  "sideEffects": false
```

`"./src/generators/*FileGenerator.ts", "./src/helpers/FileHelpers.ts"]` haven't any side effects for other code in package - these files only use another code and aren't used in other place :)

magicmatatjahu (changes requested):

package.json
```diff
@@ -27,6 +27,7 @@
   "main": "lib/cjs/index.js",
   "module": "lib/esm/index.js",
   "types": "lib/types/index.d.ts",
+  "sideEffects": false
```
magicmatatjahu, line 30:
```suggestion
  "sideEffects": false,
```

Sorry 😅

src/generators/AbstractGenerator.ts
```diff
@@ -34,28 +34,55 @@ export abstract class AbstractGenerator<Options extends CommonGeneratorOptions =
...
-    return this.generateModels(model);
+  /**
+   * Generates the full output of a model, instead of a scattered model.
+   * 
+   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
+   * 
+   * @param input 
+   * @param options to use for rendering full output
+   */
+  public async generateCompleteModel(input: Record<string, unknown> | CommonInputModel, options: RenderCompleteModelOptions): Promise<OutputModel[]> {
```
magicmatatjahu, line 51: And last thing. Sorry 😅 

```suggestion
  public async generateCompleteModels(input: Record<string, unknown> | CommonInputModel, options: RenderCompleteModelOptions): Promise<OutputModel[]> {
```

magicmatatjahu (approved):

PR #470, opened 2021-10-14
feat: adding enumType options for TS generator to generate union or regular enum model

### LINKED ISSUE #464: Enum in TS for const types
#### Describe the bug
For a constant property a new enum is created with only this one key. Now I cannot simply assign the constant value but rather I have to use the enum and the key in the enum. This is problematic as I cannot predict the enum name beforehand (sometimes it is AnonymousSchema_X). It would be easier to use if the const is translated to a constant typescript type.

For example I have a constant property event with only possible key 'ping'. This generated code, desired generated code and my desired usage of the code is displayed below:
```
Generated:
export interface Root {
    event: Event;
}
export enum Event {
    PING = "ping",
}

Desired:
export interface Root {
    event: "ping";
}

Usage:
let test: Root = {
    event: "ping"
}
```

#### How to Reproduce
Use this jsonSchema in the typescript-interface example
```
const jsonSchemaDraft7 = {
  type: 'object',
  additionalProperties: false,
  "properties": {
    "event": {
      "type": "string",
      "const": "ping"
    }
  },
  "required": [
    "event"
  ]
};
```

panwauu:
**Description**

- Adding option enumType to TSoptions
- This lets enums be rendered as unions
- Added test
- Added doc in options and typescript specific docs

**Related issue(s)**
Resolves #464


jonaslagoni:
And you have a linting error: `npm run lint`

jonaslagoni:
@all-contributors please add @panwauu for code, test, example and doc

PR #478, opened 2021-10-15
fix: correctly formatting property and model names

jonaslagoni:
**Description**

- ...
- ...
- ...

**Related issue(s)**
Fixes #456
Fixes #452
Fixes #449
Fixes #388

PR #493, opened 2021-11-01
docs: show how to change default data model naming convention

jonaslagoni:
Description
As I was changing the naming convention for data models in #478, I need to be able to show how it can be changed.

Related issue(s)
Fixes https://github.com/asyncapi/modelina/issues/415


jonaslagoni:
With the new MetaModel, I will recreate this.

PR #497, opened 2021-11-02
fix: special characters in enum values

### LINKED ISSUE #450: Special cases in enum values fails to generate accurate enum output
#### Describe the bug
Given the schema:
```
{ "enum": ["medium", "medium+"] } 
```
It would generate the enum output 

```
export enum X {
  MEDIUM = "medium",
  MEDIUM = "medium+"
}
```

Which cannot happen as no duplicate keys may be generated

panwauu:
**Description**
I added a Helper Function to resolve ascii characters that are not numbers or letters ([0-9a-zA-Z]) and translate them to english. - becomes minus, + becomes plus ...
The function also accepts characters to be ignored and a seperator string to seperate the translated characters from the rest.
Applied this function to the enum renderer to resolve the issue

**Related issue(s)**
Fixes #450

**Tests**
In #450 a blacklisted test is mentioned

### DIFF
--- a/src/generators/typescript/renderers/EnumRenderer.ts
+++ b/src/generators/typescript/renderers/EnumRenderer.ts
@@ -50,7 +50,7 @@ ${this.indent(this.renderBlock(content, 2))}
       break;
     }
     default: {
-      key = FormatHelpers.replaceSpecialCharacters(String(value), { exclude: [' '], separator: '_' })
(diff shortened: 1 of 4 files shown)


jonaslagoni:
@panwauu you have one sonarcloud code smell :)

jonaslagoni (approved):
This looks great, good job! 🎉 

Do you want to create the follow-up issue, that `enum: ['test+', 'test_plus']` causes clashes, as both will generate `TEST_PLUS` name?

PR #511, opened 2021-11-15
fix: ensure JavaCommonPreset can be compiled 

### LINKED ISSUE #510: Java equals and hashcode breaks compilation
When generating a java model including equals and hashcode settings, the resulting files break compilation.

The reason is that equals is generated using `Objects.equals` util method, but the relative `import java.util.Objects` is omitted.

Sample repo: https://github.com/ThanksForAllTheFish/modelinabug

How to reproduce once cloned:

- `npm ci`
- `npm run generate`
- check files in `java/org/t4atf/asyncapi`, equals and hashcode are properly implemented, but the import is missing

Verified against latest modelina npm package (0.39.4)

jonaslagoni:
**Description**
This PR cleans up the JavaCommonPreset a bit and fixes a couple of bugs.

- Fixed that we always assumed `additionalProperties` where needed.
- Fixed that if no properties are present it renders accurately.
- Fixed properties with type array did not work.
- Added BlackBox tests for different generator options, which now tests the preset.

**Related issue(s)**
Fixes #510

### DIFF
--- a/examples/java-generate-tostring/__snapshots__/index.spec.ts.snap
+++ b/examples/java-generate-tostring/__snapshots__/index.spec.ts.snap
@@ -12,7 +12,6 @@ Array [
   public String toString() {
     return \\"class Root {\\\\n\\" +   
       \\"    email: \\" + toIndentedString(email) + \\"\\\\n\\" +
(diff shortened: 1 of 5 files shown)


magicmatatjahu (approved):
LGTM!

jonaslagoni:
Thanks @magicmatatjahu 🙏

PR #526, opened 2021-12-13
fix: special characters in java enum values (#502)

### LINKED ISSUE #502: Special cases in enum values fails to generate accurate enum keys for Java generator
#### Describe the bug
Thanks to @panwauu, we already have it solved in the TS generator, see https://github.com/asyncapi/modelina/issues/450 and fix in https://github.com/asyncapi/modelina/pull/497.

We need to apply the same fix we did here:
https://github.com/asyncapi/modelina/blob/2a8659da465f53f9cfe1097a622ed82d0fa7594c/src/generators/typescript/renderers/EnumRenderer.ts#L53
to this line:
https://github.com/asyncapi/modelina/blob/2a8659da465f53f9cfe1097a622ed82d0fa7594c/src/generators/java/renderers/EnumRenderer.ts#L52

quadrrem:
**Description**
I changed the Java EnumRenderer to sanitize the enum keys by using `FormatHelpers.replaceSpecialCharacters` function in the same way as the Typescript EnumRender already does. 
I also duplicated the test to check that the keys are rendered properly.

**Related issue(s)**
Fixes #502

### DIFF
--- a/src/generators/java/renderers/EnumRenderer.ts
+++ b/src/generators/java/renderers/EnumRenderer.ts
@@ -49,7 +49,7 @@ ${this.indent(this.renderBlock(content, 2))}
       break;
     }
     default: {
-      key = String(value);
+      key = FormatHelpers.replaceSpecialCharacters(String(value), { exclude: [' '], separator: '_' }); 
       //Ensure no special char can be the beginning letter 
       if (!(/^[a-zA-Z]+$/).test(key.charAt(0))) {
         key = `string_${key}`;
--- a/test/generators/java/JavaGenerator.spec.ts
+++ b/test/generators/java/JavaGenerator.spec.ts
@@ -376,6 +376,54 @@ public enum CustomEnum {
     expect(enumModel.dependencies).toEqual(expectedDependencies);
   });
 
+  test('should render enums with translated special characters', async () => {
+    const doc = {
+      $id: 'States',
+      enum: ['test+', 'test', 'test-', 'test?!', '*test']
+    };
+    const expected = `public enum States {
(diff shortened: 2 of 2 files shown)


src/generators/java/renderers/EnumRenderer.ts
```diff
@@ -49,7 +49,7 @@ ${this.indent(this.renderBlock(content, 2))}
       break;
     }
     default: {
-      key = String(value);
+      key = FormatHelpers.replaceSpecialCharacters(String(value), { exclude: [' '], separator: '_' }); 
       //Ensure no special char can be the beginning letter 
       if (!(/^[a-zA-Z]+$/).test(key.charAt(0))) {
         key = `string_${key}`;
```
jonaslagoni, line 55: You can remove the old special chars code 🙂

jonaslagoni (changes requested):
Only have one small comment, otherwise looks good 🎉

Thanks for solving this!

src/generators/java/renderers/EnumRenderer.ts
quadrrem: I think this is still needed since even with the new escape function it can happen that the enum key starts with a number and that would lead to compile errors. 

The same code is also still present in the typescript `EnumRenderer`
https://github.com/asyncapi/modelina/blob/2a8659da465f53f9cfe1097a622ed82d0fa7594c/src/generators/typescript/renderers/EnumRenderer.ts#L55

Maybe I can move the check to a `FormatHelpers` function to remove duplication?

jonaslagoni: You are right, good catch!

If you see it as being beneficial and it removes duplicated code, feel free to do so! 🙂

jonaslagoni (approved):
LGTM 👍

jonaslagoni:
@all-contributors please add @quadrrem for code, test

Issue #503, opened 2021-11-03
Special cases in enum values fails to generate accurate enum keys for C# generator


jonaslagoni:
#### Describe the bug
Thanks to @panwauu, we already have it solved in the C# generator, see https://github.com/asyncapi/modelina/issues/450 and fix in https://github.com/asyncapi/modelina/pull/497.

We need to apply the same fix we did here:
https://github.com/asyncapi/modelina/blob/2a8659da465f53f9cfe1097a622ed82d0fa7594c/src/generators/typescript/renderers/EnumRenderer.ts#L53
to this line:
https://github.com/asyncapi/modelina/blob/2a8659da465f53f9cfe1097a622ed82d0fa7594c/src/generators/csharp/renderers/EnumRenderer.ts#L110

ron-debajyoti:
Hi @jonaslagoni , let me take this up and try to solve this issue !

jonaslagoni:
No need to ask for permission, go right ahead @ron-debajyoti 🙂

ron-debajyoti:
Hi @jonaslagoni !
I followed the same process used to fix #497. Please review the pull request and the test snapshots I added.

Issue #460, opened 2021-10-13
Go generates reserved keyword properties


jonaslagoni:
#### Describe the bug

Given the input:

```
{
  properties: {
    "break": {type: "string"}
  }
}
```

It will generate a struct with the reserved keyword `break` - https://notes.shichao.io/golang/

jonaslagoni:
Still relevant

jonaslagoni:
We need to create a list of reserved keywords for Go, as seen here for Java: https://github.com/asyncapi/modelina/blob/master/src/generators/java/Constants.ts

All of this functionality is somewhat already located for Java as you can use it as a reference 🙂 

Next we need to use it for our naming convention: https://github.com/asyncapi/modelina/blob/10aca3b10d985596241761f8976697a72dc360e1/src/generators/go/GoGenerator.ts#L17

Use the newly list of reserved keywords: https://github.com/asyncapi/modelina/blob/master/src/generators/go/GoRenderer.ts#L59
https://github.com/asyncapi/modelina/blob/master/src/generators/go/GoRenderer.ts#L71

Change default naming convention: https://github.com/asyncapi/modelina/blob/10aca3b10d985596241761f8976697a72dc360e1/src/generators/go/GoGenerator.ts#L25


And then just adapt tests of course 👍

Issue #456, opened 2021-10-11
Numbers cannot be used as names of classes


jonaslagoni:
#### Describe the bug
Given a schema such as:

```json
{
   "$id": "1234Test",
    "type": "object",
    "properties": {
        "test": {
            "type": "string"
        }
    }
}
```

it renders the name of the class `export class 1234Test {` which is not supported in most if not all languages.

jonaslagoni:
still relevant

jonaslagoni:
Already fixed in `next` as part of the constraint logic.

Issue #452, opened 2021-10-11
Numbers cannot be used as identifiers for properties


jonaslagoni:
#### Describe the bug
Given a schema such as 
```
{
    "type": "object",
    "description": "One or more icons that represent the extension, app, or theme. Recommended format: PNG; also BMP, GIF, ICO, JPEG.",
    "minProperties": 1,
    "properties": {
        "16": {
            "type": "string",
            "description": "Used as the favicon for an extension's pages and infobar."
        },
        "48": {
            "type": "string",
            "description": "Used on the extension management page (chrome://extensions)."
        },
        "128": {
            "type": "string",
            "description": "Used during installation and in the Chrome Web Store."
        },
        "256": {
            "type": "string",
            "description": "Used during installation and in the Chrome Web Store."
        }
    }
}
```

The following is generated:

```ts
export class Icons {
  private _16?: string;
  private _48?: string;
  private _128?: string;
  private _256?: string;
  private _additionalProperties?: Map<String, object | string | number | Array<unknown> | boolean | null>;

  constructor(input: {
    16?: string,
    48?: string,
    128?: string,
    256?: string,
  }) {
    this._16 = input.16;
    this._48 = input.48;
    this._128 = input.128;
    this._256 = input.256;
  }

  get 16(): string | undefined { return this._16; }
  set 16(16: string | undefined) { this._16 = 16; }

  get 48(): string | undefined { return this._48; }
  set 48(48: string | undefined) { this._48 = 48; }

  get 128(): string | undefined { return this._128; }
  set 128(128: string | undefined) { this._128 = 128; }

  get 256(): string | undefined { return this._256; }
  set 256(256: string | undefined) { this._256 = 256; }

  get additionalProperties(): Map<String, object | string | number | Array<unknown> | boolean | null> | undefined { return this._additionalProperties; }
  set additionalProperties(additionalProperties: Map<String, object | string | number | Array<unknown> | boolean | null> | undefined) { this._additionalProperties = additionalProperties; }
}
```

Related to https://github.com/asyncapi/modelina/issues/449

jonaslagoni:
Related BlackBox tests that currently is being blacklisted -
https://github.com/asyncapi/modelina/blob/3f14b6cc44eb10af4c74fbba9d8d12037ae66ad4/test/blackbox/blackbox.spec.ts#L57

PR #530, opened 2021-12-15
docs: new core data model

jonaslagoni:
> The change I am proposing can be read about here: https://github.com/asyncapi/modelina/blob/2769d6f20bd7f3ad0cd7ac525b00d4c7844e9246/docs/processing.md

## The why
These are the issues that are triggering the change, that simply just shows the problems with the current **CommonModel** structure. We need a more precise structure to solve and prevent these issues from keep popping up:

- https://github.com/asyncapi/modelina/issues/243
- https://github.com/asyncapi/modelina/issues/367
- https://github.com/asyncapi/modelina/issues/388
- https://github.com/asyncapi/modelina/issues/390
- https://github.com/asyncapi/modelina/issues/449
- https://github.com/asyncapi/modelina/issues/452
- https://github.com/asyncapi/modelina/issues/456
- https://github.com/asyncapi/modelina/issues/460
- https://github.com/asyncapi/modelina/issues/503
- Code injection problem - https://github.com/asyncapi/modelina/security/advisories/GHSA-4jg2-84c2-pj95

The following issues highlight the need for better UX:
- https://github.com/asyncapi/modelina/discussions/487#discussioncomment-1520496
- https://github.com/asyncapi/modelina/issues/181
- https://github.com/asyncapi/modelina/issues/365 

## The rewards

- No tight coupling to JSON Schema structure.
- Better UX for preset and generator developers. 
- Better UX for users of Modelina, as it will be easier to understand the underlying data model.
- Abstraction between data model types.
- Enable easier representation of other input types, such as ProtoBuf, XML, JSON Type Definition, and many many more.

## The code changes
So with these changes, how does it simplify the generators and presets?

For the examples below, I only focus on TypeScript but keep in mind these changes are relevant for each generator and preset. 

<details>
<summary>Rendering of properties</summary>
<p>

[Old implementation can be found here](https://github.com/asyncapi/modelina/blob/d888adc86d89b989e4f2c2df5282b384a0b72cfe/src/generators/typescript/TypeScriptRenderer.ts#L123-L147). 

New simplified implementation becomes straight forward because `patternProperties` and `additionalProperties` are already resolved directly as **ObjectModel** properties using the new `DictionaryModel`.
```ts
  async renderProperties(): Promise<string> {
    const properties = this.model.properties || {};
    const content: string[] = [];

    for (const [propertyName, property] of Object.entries(properties)) {
      const rendererProperty = await this.runPropertyPreset(propertyName, property);
      content.push(rendererProperty);
    }
    return this.renderBlock(content);
  }
```

</p>
</details>  

<details>
<summary>Serialization preset</summary>
<p>

[Old implementation can be found here](https://github.com/asyncapi/modelina/blob/d888adc86d89b989e4f2c2df5282b384a0b72cfe/src/generators/typescript/presets/CommonPreset.ts#L26-L79). 

New simplified implementation for de/serialization functionality make use of the `DictionaryModel` and `serializationType: 'unwrap' | 'normal'` to support this behavior:

```ts
function renderMarshalProperty(modelInstanceVariable: string, model: ConstrainedMetaModel) {
  if (model instanceof ConstrainedReferencedModel) {
    const resolvedModel = model.referencedModel;
    if (resolvedModel instanceof ConstrainedObjectModel) {
      return `$\{${modelInstanceVariable}.marshal()}`;
    }
  }
  return realizePropertyFactory(modelInstanceVariable);
}
function renderMarshalProperties(model: ObjectModel, renderer: TypeScriptRenderer) {
  const properties = model.properties || {};
  const propertyKeys = [...Object.entries(properties)];
  const marshalProperties = propertyKeys.map(([prop, propModel]) => {
    const formattedPropertyName = renderer.nameProperty(prop, propModel);
    if (propModel.propertyModel instanceof ConstrainedDictionaryModel && propModel.propertyModel.serializationType === 'unwrap') {
      const modelInstanceVariable = 'value';
      const patternPropertyMarshalCode = renderMarshalProperty(modelInstanceVariable, propModel.propertyModel.keyType);
      const marshalCode = `json += \`"$\{key}": ${patternPropertyMarshalCode},\`;`;
      return `if(this.${prop} !== undefined) { 
  for (const [key, value] of this.${prop}.entries()) {
    //Only unwrap dictionary entries which are not already part of the properties
    if(Object.keys(this).includes(String(key))) continue;
    ${marshalCode}
  }
}`;
    } 
    const modelInstanceVariable = `this.${formattedPropertyName}`;
    const propMarshalCode = renderMarshalProperty(modelInstanceVariable, propModel.propertyModel);
    const marshalCode = `json += \`"${prop}": ${propMarshalCode},\`;`;
    return \`if(${modelInstanceVariable} !== undefined) {
    ${marshalCode} 
  }`;
  });
  return marshalProperties.join('\n');
}
```

</p>
</details>

<details>
<summary>Referenced model's type</summary>
<p>

[Old implementation can be found here](https://github.com/asyncapi/modelina/blob/d888adc86d89b989e4f2c2df5282b384a0b72cfe/src/generators/typescript/presets/CommonPreset.ts#L15-L25). 

New implementation makes it easier to access the `referencedModel` directly through the new **ConstrainedReferencedModel**. 
```ts
function renderMarshalProperty(modelInstanceVariable: string, model: ConstrainedMetaModel) {
  if (model instanceof RefModel) {
    const resolvedModel = model.referencedModel;
    if (resolvedModel instanceof ConstrainedObjectModel) {
      return `$\{${modelInstanceVariable}.marshal()}`;
    }
  }
  return realizePropertyFactory(modelInstanceVariable);
}
```

</p>
</details>  

<details>
<summary>Figuring out a property's type</summary>
<p>

[Old implementation can be found here](https://github.com/asyncapi/modelina/blob/d888adc86d89b989e4f2c2df5282b384a0b72cfe/src/generators/typescript/TypeScriptRenderer.ts#L51-L93). 

Because we are dealing with a constrained data model, we can get the type as is:
```ts
  renderType(model: ConstrainedMetaModel): string {
    return model.type;
  }
```

And we will remove this function and use the type as is. 

</p>
</details>

<details>
<summary>Accessing the model type externally</summary>
<p>
Because Modelina is expected to work with AsyncAPI generator templates we need a way to access the payload type. Similar to `Figuring out a property's type`. 

Before, we had to use a bunch of internal functions to access it, and in some cases it was impossible. Now it is as easy as accessing the `type` property for the model.

```ts
const typescriptGenerator = new TypeScriptGenerator(...);
const generatedModels = await typescriptGenerator.generate(message.payload());
const messagePayloadModel = generatedModels[0];
const messagePayloadType = messagePayloadModel.type;

return `function publish(message: ${messagePayloadType}){ ... }`
```

Probably need to provide an extra external function to simplify this even more such as:
```ts
const typescriptGenerator = new TypeScriptGenerator(...);
const messagePayloadType = typescriptGenerator.getType(message.payload());

return `function publish(message: ${messagePayloadType}){ ... }`
```

But you get the idea :smile: 

</p>
</details>


docs/usage.md
```diff
@@ -13,6 +13,7 @@ For more specific integration options, please check out the [integration documen
 - [Generate models from AsyncAPI documents](#generate-models-from-asyncapi-documents)
 - [Generate models from JSON Schema documents](#generate-models-from-json-schema-documents)
 - [Generate models from Swagger 2.0 documents](#generate-models-from-swagger-20-documents)
+- [Generate models from DataModel](#generate-models-from-datamodel)
```
smoya, line 16: The name `DataModel` here seems to refer to a solution or a language. I feel it can confuse the reader. How do you feel about changing it to something like "Generate models from any other input type" or similar?

docs/README.md
```diff
@@ -18,6 +19,9 @@
 
 This document gives the overview of all the available documentation for Modelina.
 
+### [Internal language](./Internal%20language.md)
```
smoya, line 22: I see there are several names for the same thing and I think it can confuse the reader: Internal language, internal data model, custom data model. We should stick with one name.

docs/Internal language.md
```diff
@@ -0,0 +1,79 @@
+# The model language
+
+It is important to understand the domain-specific language that describes the data models. It is what the input processors converts their inputs to, and the generators use to generate the output.
```
smoya, line 3: Very personal opinion:
```suggestion
The <whatever name> is what the input processors converts their inputs to, and the generators use to generate the output.
```

I think the first sentence adds nothing useful but confuses the reader.

docs/Internal language.md
```diff
@@ -0,0 +1,79 @@
...
+
+It is also what developers can use to create their own models, since they can provide a **raw data model** as input. This way you can create your own custom data models that can be interpreted to any output language, with the full sweep of features the generators and presets support. See [Create your own models from the ground up, instead of a supported input](./advanced#create-your-own-models-from-the-ground-up-instead-of-a-supported-input)
```
smoya, line 5: It is very hard to me to read this whole sentence. I feel I need an earlier answer for: What is `raw data model` ?

docs/Internal language.md
```diff
@@ -0,0 +1,79 @@
...
+  <img src="./img/RenderingProcess.png" />
+</p>
+
+The transformation happens in three stages. 
+
+1. Process the input and transform it into the raw data model. See [The raw data model](#the-raw-data-model) for more information.
+2. Split the raw data model into separate models that are rendered separately. See [The splitting of data models](#The-splitting-of-data-models) for more information. 
+3. Constrain the data models to the output language. See [The constrained data model](#the-constrained-data-model) for more information.
+
+## The raw data model
```
smoya, line 27: I think this should be move to the very top. You mentioned "raw data model" several times before this.

smoya (commented):
Good job @jonaslagoni! I like the approach of having that abstract model so we can support for any kind of input 👍 
I dropped some comments. Might keep reviewing later!

jonaslagoni:
@derberg have you ever encountered a PR that can go around the restriction of one peer-review? It says I can merge, but @magicmatatjahu has not approved it anywhere 😅 Is the settings changed?

docs/usage.md
jonaslagoni: I like it!

jonaslagoni:
> @jonaslagoni but it is not against master but some `next` branch. So yeah, you need to enable branch protection for this branch too

Right 🤦 I don't have those permissions though 😛

docs/usage.md
voxpelli: That or "Generate models from our `DataModel`", to qualify it

jonaslagoni: Definitely also an alternative 🤔 

Guess it depends on who reads the document. Those who are looking for how to generate models using the DataModel, or those who wish to generate models from anything else and don't know what `DataModel` means.

docs/README.md
jonaslagoni: You are right, changing!

docs/usage.md
smoya: > those who wish to generate models from anything else and would not know what `DataModel` means.

This would be, IMHO, the general use case. In fact, I'm not so sure those are the "data models" people may know about. In my case, the data model concept is abstract, and I don't see a "code" representation of it but just a diagram. I might be wrong, but I'm representing as well a user of this tool so I hope my insights are useful as well and perhaps also shared with other users that may come to this doc.

jonaslagoni:
I decided to rename `data model` to `meta model` as it's a lot of the same jab jab jab and hard to understand 😅 

Hopefully, this makes much more sense 🤞 Please have a look @smoya @voxpelli @magicmatatjahu

docs/Internal language.md
jonaslagoni: Please see https://github.com/asyncapi/modelina/pull/530#issuecomment-1042000476

docs/Internal language.md
jonaslagoni: Please see https://github.com/asyncapi/modelina/pull/530#issuecomment-1042000476

docs/usage.md
jonaslagoni: Please see https://github.com/asyncapi/modelina/pull/530#issuecomment-1042000476

docs/usage.md
```diff
@@ -58,6 +59,10 @@ The Swagger input processor expects that the property `swagger` is defined in or
 
 The response payload and `body` parameters, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
 
+## Generate models from MetaModels
```
smoya, line 62: Nitpick: Just to ensure people don't get the wrong idea, `MetaModels` can sound like a custom solution out there. However, if we don't treat it as a "own name", it gives the idea that you can build from something abstract or behind (meta).
Feel free to take it or leave it 👍 
```suggestion
## Generate models from meta models
```

examples/meta-model/README.md
```diff
@@ -0,0 +1,17 @@
+# Meta Model
```
smoya, line 1: Same as https://github.com/asyncapi/modelina/pull/530/files#r808839189

In fact, you are referring to it as `meta model` in the line below.

docs/usage.md
jonaslagoni: I changed the namings, I like it 👍

examples/meta-model/README.md
jonaslagoni: Changed 😄

smoya (approved):
Nice job @jonaslagoni 🎉 👏 ! It LGTM, however i'm not an owner 👐

docs/processing.md
```diff
@@ -0,0 +1,71 @@
...
+- **FloatModel** represent floating-point numbers. 
+- **StringModel** represent string values.
+- **AnyModel** represent generic values that cannot otherwise be represented by one of the other models.
+
+<p align="center">
+  <img src="./img/MetaModel.png" />
+</p>
+
+
+## The Constrained Meta Model
```
magicmatatjahu, line 27: I understand idea about that but maybe we should call it as `Language Meta Model` and then call `JavaEnumModel` etc... I think that it will produce some confusion because in some languages it will be exactly constraint model, because some languages don't support enums/unions etc but some allow more freedom than others. What do you think?

magicmatatjahu: Or maybe just `Language Model`?

magicmatatjahu (commented):
One comment to the docs.

I have also another question related to the development of new approach. Do you wanna make it in main branch but in separate folder to have current logic and new one and in some time switch to new or do you prefer new branch (like `next`)?

docs/processing.md
jonaslagoni: The idea is that all outputs should by some means support all MetaModels in one capacity or another.

For example for unions, if they are not supported like in TS `string | number` we can always render them as `any` or create wrapper classes `class unionT { stringProp: string, numberProp: string }`. It all comes down to the output it's possibilities, as well as what the user defines through options.

The constrained variant only refers to the values within the models, i.e. property name/type, model name and type, enum values, and keys, that ensure they are properly formatted in correct syntax.

jonaslagoni:
> I have also another question related to the development of new approach. Do you wanna make it in main branch but in separate folder to have current logic and new one and in some time switch to new or do you prefer new branch (like `next`)?

My initial idea was to use `next` and iterate with small PR's being merged. Of course that means it will take a few PR's before any release on the `next` branch will happen, because tests and code will be in a balance between the two. But thats fine by me.

Mostly it will be the same code just with MetaModel and ConstrainedMetaModel instead.

magicmatatjahu:
> My initial idea was to use next and iterate with small PR's being merged. Of course that means it will take a few PR's before any release on the next branch will happen, because tests and code will be in a balance between the two. But thats fine by me.
>
> Mostly it will be the same code just with MetaModel and ConstrainedMetaModel instead.

Ok 👌🏼

docs/processing.md
jonaslagoni: Constrained, refers to the process the meta models go through, i.e. the contrainer (`TypeScriptConstrained`, etc).

What would you change the naming of the process and model name to, to make it consistant? 🤔 

Can you try elaborate a bit more why you think `constrain` does not encapsulate correctly what it does?

magicmatatjahu: The name may be but I am a man who thinks differently. To me constraint means that something is limited but in a validation/parsing way (probably too much JSON Scheme, hah 😄 ) and not in a language constraint way. I'm also not a fan of Java jargon to call it like `LanguageConstrainedMetaModel`, so `ConstrainedMetaModel` is ok but I'd prefer without Meta -> `ConstrainedModel`.

jonaslagoni: Haha, okay yea, I see 😄 Is there a way we need to highlight this difference in the docs?

Regarding `ConstrainedModel`, the "problem" is that the constrained model, is of type meta model as it inherits the class 😅 Think that will create more confusion, even though it is a long name 🤔 

What if we keept it for now, and once we are ready to merge from `next` to `master` we can evaluate it then if it makes sense or not?

PR #539, opened 2021-12-21
feat: complete model rendering and file output for JavaScript generator

jonaslagoni:
**Description**
This PR adds both complete and file output for the JS generator.

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/400

### DIFF
--- a/src/generators/javascript/JavaScriptGenerator.ts
+++ b/src/generators/javascript/JavaScriptGenerator.ts
@@ -12,8 +12,8 @@ export interface JavaScriptOptions extends CommonGeneratorOptions<JavaScriptPres
   namingConvention?: CommonNamingConvention
 }
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface JavaScriptRenderCompleteModelOptions {
+  moduleSystem?: 'ESM' | 'CJS';
 }
 
 /**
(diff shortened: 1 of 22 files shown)


src/generators/javascript/JavaScriptGenerator.ts
```diff
@@ -27,9 +31,29 @@ export class JavaScriptGenerator extends AbstractGenerator<JavaScriptOptions> {
...
+   * @param model
+   * @param inputModel
+   * @param options
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async renderCompleteModel(model: CommonModel, inputModel: CommonInputModel, options: JavaScriptRenderCompleteModelOptions): Promise<RenderOutput> {
+    const outputModel = await this.render(model, inputModel);
+    const modelDependencies = model.getNearestDependencies().map((dependencyModelName) => {
+      const formattedDependencyModelName = this.options.namingConvention?.type ? this.options.namingConvention.type(dependencyModelName, { inputModel, model: inputModel.models[String(dependencyModelName)] }) : dependencyModelName;
+      return `const ${formattedDependencyModelName} = require('./${formattedDependencyModelName}');`;
```
magicmatatjahu, line 47: Have in mind that someone will want to render models with ESM not CJS module system, so you `JavaScriptRenderCompleteModelOptions` should have option to choose it - by default we can use ESM.

magicmatatjahu (changes requested):

src/generators/javascript/JavaScriptGenerator.ts
jonaslagoni: Good catch! Adapting.

jonaslagoni:
@magicmatatjahu I adapted the code to now support different module systems as well as adapted tests.

I also added two examples (why the total changes increased 😅) that show both systems.

src/generators/javascript/JavaScriptGenerator.ts
```diff
@@ -27,9 +31,33 @@ export class JavaScriptGenerator extends AbstractGenerator<JavaScriptOptions> {
...
+      if (options.moduleSystem === 'CJS') {
+        return `const ${formattedDependencyModelName} = require('./${formattedDependencyModelName}');`;
+      }
+      return `import ${formattedDependencyModelName} from './${formattedDependencyModelName}';`;
+    });
+    let modelCode = `export default ${outputModel.result}`;
+    if (options.moduleSystem === 'CJS') {
+      modelCode = `${outputModel.result}
+module.exports = ${outputModel.renderedName};`;
+    }
```
magicmatatjahu, line 56: LGTM, but `switch` should be better option here :)

magicmatatjahu (approved):

src/generators/javascript/JavaScriptGenerator.ts
jonaslagoni: Well, I know Sonarcloud don't like small switch statements, so better not 😆

PR #540, opened 2021-12-21
feat: complete model rendering and file output for Go generator

jonaslagoni:
**Description**
This PR adds both complete model rendering and file output for the Go generator.

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/398


jonaslagoni:
@smoya maybe you want to have a look as well? 🙂

jonaslagoni:
And I have no idea why this is failing...

test/generators/go/GoGenerator.spec.ts
```diff
@@ -155,4 +155,31 @@ type CustomEnum string`;
     expect(enumModel.result).toEqual(expected);
     expect(enumModel.dependencies).toEqual([]);
   });
+  test('should render models and their dependencies', async () => {
```
smoya, line 158: I don't see dependencies being set in this test.

src/generators/go/GoGenerator.ts
```diff
@@ -69,8 +73,26 @@ export class GoGenerator extends AbstractGenerator<GoOptions> {
...
+   * @param model
+   * @param inputModel
+   * @param options
+   */
+  async renderCompleteModel(model: CommonModel, inputModel: CommonInputModel, options: GoRenderCompleteModelOptions): Promise<RenderOutput> {
+    const outputModel = await this.render(model, inputModel);
+    let importCode = '';
+    if (outputModel.dependencies.length > 0) {
+      importCode = `import (  
+  ${outputModel.dependencies.join(',\n')}
```
smoya, line 88: Package names should be double-quoted, for example: `"io"`. I don't think this code does but not sure.

test/generators/go/GoFileGenerator.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        }
+      }
+    };
+    test('should throw accurate error if file cannot be written', async () => {
+      const generator = new GoFileGenerator();
+      const expectedError = new Error('write error');
+      jest.spyOn(FileHelpers, 'writerToFileSystem').mockRejectedValue(expectedError);
+      jest.spyOn(generator, 'generateCompleteModels').mockResolvedValue([new OutputModel('content', new CommonModel(), '', new CommonInputModel(), [])]);
+    
+      await expect(generator.generateToFiles(doc, '/test/', {packageName: 'SomePackage'})).rejects.toEqual(expectedError);
```
smoya, line 29: Just to be a bit consistent with Go standards, I would recommend you a name that follows:

```
Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:

time (provides functionality for measuring and displaying time)
list (implements a doubly linked list)
http (provides HTTP client and server implementations)
The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:

computeServiceClient
priority_queue
```

Source: https://go.dev/blog/package-names

smoya, line 29: Just to be a bit consistent with Go standards, I would recommend you a name that follows:

```
Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:

time (provides functionality for measuring and displaying time)
list (implements a doubly linked list)
http (provides HTTP client and server implementations)
The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:

computeServiceClient
priority_queue
```

Source: https://go.dev/blog/package-names

test/generators/go/GoGenerator.spec.ts
jonaslagoni: Changed 🙂

src/generators/go/GoGenerator.ts
jonaslagoni: Fixed!

test/generators/go/__snapshots__/GoGenerator.spec.ts.snap
```diff
@@ -0,0 +1,65 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`GoGenerator generateCompleteModels() should render dependencies 1`] = `
+"
+package some_package
+import (
```
smoya, line 6: I'm confused why there is no new line between `package` declaration and before/after `import` one, but there is here: https://github.com/asyncapi/modelina/pull/540/files#diff-b5e5cf8b3a4b39ae9bd7669e4a81ccb851cdc9802b1615e7162dc568f4c35381R39

smoya (approved):
Only one minor. LGTM otherwise

test/generators/go/__snapshots__/GoGenerator.spec.ts.snap
jonaslagoni: The reason it renders the new line is that a string as set here: https://github.com/asyncapi/modelina/blob/c58c71c8c4749dd06e8b81dab784c88a5f28480f/src/generators/go/GoGenerator.ts#L92

Renders the actual new lines contained with the ```

smoya: Intersting. So then it means all those lines should be separated by an empty line: https://github.com/asyncapi/modelina/blob/c58c71c8c4749dd06e8b81dab784c88a5f28480f/src/generators/go/GoGenerator.ts#L93-L95

jonaslagoni: If you want to have an empty line between them then yes. At the moment, the empty line is present because there was no imports 🙂

smoya: I mean, it is not strictly necessary because the program will run anyway but it looks unaesthetic if there is no empty line between package and import, import and rest. But nothing else so maybe we can just skip it for this time.

smoya (approved):

magicmatatjahu (approved):
LGTM! 🚀

jonaslagoni:
Thanks for the reviews @smoya @magicmatatjahu  🙇

PR #541, opened 2021-12-21
docs: added missing documentation for indentation example

jonaslagoni:
**Description**
This PR adds the missing documentation for the indentation example.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -67,7 +67,9 @@ The library uses 4 different logging levels:
 Check out this [example out for a live demonstration](../examples/custom-logging).
 
 ## Change the generated indentation type and size
-TODO 
+In some scenarios, depending on how you stitch them together, you might need to change the indentation type or size both of these cases are fully supported.
+
+Check out this [example out for a live demonstration](../examples/indentation-type-and-size).
 
 ## Change the naming format for properties
 TODO 


docs/advanced.md
```diff
@@ -67,7 +67,9 @@ The library uses 4 different logging levels:
 Check out this [example out for a live demonstration](../examples/custom-logging).
 
 ## Change the generated indentation type and size
-TODO 
+In some scenarios, depending on how you stitch them together, you might need to change the indentation type or size both of these cases are fully supported.
```
magicmatatjahu, line 70: I'm not a expert of English but maybe that `and` sounds better:

```suggestion
In some scenarios, depending on how you stitch them together, you might need to change the indentation type or size and both of these cases are fully supported.
```

@alequetzalli What do you think? 🤔

docs/advanced.md
```diff
@@ -67,7 +67,9 @@ The library uses 4 different logging levels:
...
+
+Check out this [example out for a live demonstration](../examples/indentation-type-and-size).
```
magicmatatjahu, line 72: I have problem with that `live demonstration`. If I read live I think that there are some playground, but we have only source code to check, so please remove `live`. Please also remove that `live` in other parts of this doc.

magicmatatjahu (changes requested):

docs/advanced.md
jonaslagoni: Please create another issue for this, regardless of the outcome, this won't be changed in this PR 🙂

magicmatatjahu (approved):
LGTM!

alequetzalli:
sorry! I was on vacation and didn't see this in time 🙏🏽❤️

PR #545, opened 2021-12-30
fix: wrong property name in JavaCommonPresetOptions interface (#544)

### LINKED ISSUE #544: Wrong property name in JavaCommonPresetOptions
#### Describe the bug
`JavaCommonPresetOptions` interface in `src/generators/java/presets/CommonPreset.ts` defines:
```typescript
export interface JavaCommonPresetOptions {
  equal: boolean;
  hash: boolean;
  classToString: boolean;
}
```
see `hash`

but the actual code expects: `options.hashCode`
```typescript
const shouldContainHashCode = options.hashCode === undefined || options.hashCode === true;
```
https://github.com/asyncapi/modelina/blame/master/src/generators/java/presets/CommonPreset.ts#L7
https://github.com/asyncapi/modelina/blame/master/src/generators/java/presets/CommonPreset.ts#L111

#### How to Reproduce
I found this when trying to generalize creation of JavaGenerator by creating a factory method for it. When I tried to use the example with this interface it obviously failed as (working) example states:
```typescript
    {
      preset: JAVA_COMMON_PRESET,
      options: {
        hashCode: true,
        classToString: true,
      }
    }
```

#### Expected behavior
The options interface should match the actual options of the preset or it shouldn't be exported at all. From simple github search the interface is not used anywhere.

Extra note: maybe it would be nicer to also change `equal` -> `equals` as the generated method is `equals` not `equal` but that would be breaking change.

---

If the fix will be just changing the name of the property I could provide the simple PR. 
/my first issue for this project so I didn't start with a PR and from what I see it probably won't be last but im happy to help :)/

kamko:
**Description**
Changed wrong `hash` property in `JavaCommonPresetOptions` to `hashCode` as that the value that is actually used.

**Related issue(s)**
Fixes #544

### DIFF
--- a/test/generators/java/presets/CommonPreset.spec.ts
+++ b/test/generators/java/presets/CommonPreset.spec.ts
@@ -29,27 +29,6 @@ describe('JAVA_COMMON_PRESET', () => {
     const classModel = await generator.renderClass(model, inputModel);
     expect(classModel.result).toMatchSnapshot();
   });
-  test('should not render any functions when all 3 options are disabled', async () => {
-    const options: JavaCommonPresetOptions = {
-      equal: false,
-      hashCode: false,
-      classToString: false,
-    };
-
-    const generator = new JavaGenerator(
-      {
-        presets: [{
-          preset: JAVA_COMMON_PRESET,
-          options
-        }]
-      }
-    );
-    const inputModel = await generator.process(doc);
-    const model = inputModel.models['Clazz'];
-
-    const classModel = await generator.renderClass(model, inputModel);
-    expect(classModel.result).toMatchSnapshot();
-  });
   describe('with option', () => {
     test('should render all functions', async () => {
       const generator = new JavaGenerator(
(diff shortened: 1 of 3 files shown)


test/generators/java/presets/CommonPreset.spec.ts
```diff
@@ -29,6 +29,27 @@ describe('JAVA_COMMON_PRESET', () => {
...
+          options
+        }]
+      }
+    );
+    const inputModel = await generator.process(doc);
+    const model = inputModel.models['Clazz'];
+
+    const classModel = await generator.renderClass(model, inputModel);
+    expect(classModel.result).toMatchSnapshot();
+  });
```
jonaslagoni, line 52: @kamko do you mind moving this test into the [related group of tests for options](https://github.com/asyncapi/modelina/blob/cf3658dd55901042f12fbc3ddc91e2d483b93a9d/test/generators/java/presets/CommonPreset.spec.ts#L53)? 🙂

jonaslagoni (changes requested):
Just a small change request, otherwise looks good 😄

test/generators/java/presets/CommonPreset.spec.ts
kamko: I had it there first but then I moved it to the top level as I'm only testing that it will compile and not the options themselves.

Although a second look tells me that there wasn't the test for the all-false case so it fits there 😅

jonaslagoni (approved):
LGTM 💯

jonaslagoni:
@all-contributors please add @kamko for test, bug, code

PR #553, opened 2022-01-03
fix: non-required properties are required by constructor

### LINKED ISSUE #538: Non-required properties are required by constructor
#### Describe the bug
The JavaScript generator creates constructors that requires non-required properties.

#### How to Reproduce
Given the following input: 

```js
{
      $id: 'TestObj',
      type: 'object',
      properties: {
        requiredProp: { type: 'string' },
        optionalProp: { type: 'string' },
      },
      required: ['requiredProp'],
}
```

It generates:

```js
class TestObj {
  streetName;
  city;

  constructor(input) {
    this.requiredProp = input.requiredProp;
    this.optionalProp = input.optionalProp;
  }

  get requiredProp() { return this.requiredProp; }
  set requiredProp(requiredProp) { this.requiredProp = requiredProp; }

  get optionalProp() { return this.optionalProp; }
  set optionalProp(optionalProp) { this.optionalProp = optionalProp; }
}
```

#### Expected behavior
Only expected the constructor to be:

```js
  constructor(input) {
    this.requiredProp = input.requiredProp;
  }
```

This is because the [JavaScript class render](https://github.com/asyncapi/modelina/blob/593a6e83740827444d7e72746a8feec8a82f3a78/src/generators/javascript/renderers/ClassRenderer.ts) creates the parameters based on all properties instead of only the required properties, see:

https://github.com/asyncapi/modelina/blob/593a6e83740827444d7e72746a8feec8a82f3a78/src/generators/javascript/renderers/ClassRenderer.ts#L74

To solve this, filter out any properties which are not required `model.isRequired(propertytName)`.

mahakporwal02:
**Description**
The JavaScript generator created constructors that requires non-required properties.
Added a filter function that only filters required properties.

**Related issue(s)**
Fixes #538

### DIFF
--- a/examples/generate-javascript-models/index.ts
+++ b/examples/generate-javascript-models/index.ts
@@ -10,8 +10,7 @@ const jsonSchemaDraft7 = {
       type: 'string',
       format: 'email'
     }
-  },
-  required: ['email'],
+  }
 };
 
 export async function generate() : Promise<void> {
--- a/examples/generate-javascript-models/index.ts
+++ b/examples/generate-javascript-models/index.ts
@@ -10,7 +10,8 @@ const jsonSchemaDraft7 = {
       type: 'string',
       format: 'email'
     }
-  }
+  },
+  required: ['email'],
 };
 
 export async function generate() : Promise<void> {
--- a/examples/javascript-use-cjs/index.ts
+++ b/examples/javascript-use-cjs/index.ts
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      }
+      },
+      required: ['email']
     }
-  }
+  },
+  required: ['person']
 };
 
 export async function generate() : Promise<void> {
--- a/examples/javascript-use-cjs/index.ts
+++ b/examples/javascript-use-cjs/index.ts
@@ -14,11 +14,9 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      },
-      required: ['email']
+      }
     }
-  },
-  required: ['person']
+  }
 };
 
 export async function generate() : Promise<void> {
--- a/examples/javascript-use-esm/index.ts
+++ b/examples/javascript-use-esm/index.ts
@@ -14,11 +14,9 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      },
-      required: ['email'],
+      }
     }
-  },
-  required: ['person'],
+  }
 };
 
 export async function generate() : Promise<void> {
--- a/examples/javascript-use-esm/index.ts
+++ b/examples/javascript-use-esm/index.ts
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      }
+      },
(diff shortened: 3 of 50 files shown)


mahakporwal02:
@jonaslagoni I just realised one of the tests is failing. I'll update the test later today.

jonaslagoni:
You can update the changed snapshots with `npm run test -- -u` 🙂

mahakporwal02:
@jonaslagoni Sorry I missed your previous comment and changed the tests by updating the test cases instead of the snapshots.
Kindly have a look if this works or should I anyway update the snapshots only?

examples/javascript-use-cjs/index.ts
```diff
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      }
+      },
+      required: ['email']
```
jonaslagoni, line 18: To stay consistent I don't think we should add anything unnecessary to the inputs, as you already have a test in the JavaScriptGenerator 🙂

examples/generate-javascript-models/index.ts
```diff
@@ -10,7 +10,8 @@ const jsonSchemaDraft7 = {
       type: 'string',
       format: 'email'
     }
-  }
+  },
+  required: ['email'],
```
jonaslagoni, line 14: To stay consistent I don't think we should add anything unnecessary to the inputs, as you already have a test in the JavaScriptGenerator 🙂

examples/javascript-use-esm/index.ts
```diff
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
           type: 'string',
           format: 'email'
         }
-      }
+      },
+      required: ['email'],
```
jonaslagoni, line 18: To stay consistent I don't think we should add anything unnecessary to the inputs, as you already have a test in the JavaScriptGenerator 🙂

examples/javascript-use-esm/index.ts
```diff
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
...
     }
-  }
+  },
+  required: ['person'],
```
jonaslagoni, line 21: To stay consistent I don't think we should add anything unnecessary to the inputs, as you already have a test in the JavaScriptGenerator 🙂

examples/javascript-use-cjs/index.ts
```diff
@@ -14,9 +14,11 @@ const jsonSchemaDraft7 = {
...
     }
-  }
+  },
+  required: ['person']
```
jonaslagoni, line 21: To stay consistent I don't think we should add anything unnecessary to the inputs, as you already have a test in the JavaScriptGenerator 🙂

jonaslagoni (changes requested):
Just have a small suggestion, otherwise it looks good 👍 🙂

jonaslagoni (approved):
Perfect 🎉

jonaslagoni:
!rtm

jonaslagoni:
👀 Well, it should not complain about that... Gonna need to figure out why it complains

jonaslagoni:
@all-contributors please add @mahakporwal02 for code

PR #554, opened 2022-01-03
fix: correctly store already converted schemas while parsing AsyncApi schema (#548)

### LINKED ISSUE #548: AsyncAPIInputProcessor incorrectly stores encountered schemas
#### Describe the bug
When you have a schema that references some other schema more than once `AsyncAPIInputProcessor` fails to correctly convert all but the first property.

#### How to Reproduce
Create AsyncAPI spec with simple schemas:
```yaml
components:
  schemas:
    RootSchema:
      type: object
      properties:
        firstRef:
          $ref: '#/components/schemas/ReferencedSchema'
        secondRef:
          $ref: '#/components/schemas/ReferencedSchema'
    ReferencedSchema:
      type: object
      properties:
        str:
          type: string
```
For simplicity, I'll show the result from JavaGenerator. The resulting java classes contain these properties:
```java
  private ReferencedSchema firstRef;
  private Object secondRef;
```

#### Expected behavior
The correct result would be:
```java
  private ReferencedSchema firstRef;
  private ReferencedSchema secondRef;
```

In the background what is happening is that during processing the schema for `secondRef` is set to `{}` (actually it's `new AsyncapiV2Schema()`).

#### Cause
The issue comes from these lines:
https://github.com/asyncapi/modelina/blob/4ab00523213ae5b01992270cbd05ccc3d0c6ebd6/src/processors/AsyncAPIInputProcessor.ts#L58-L61

As the first thing, we store empty schema object which we want to hydrate in the rest of the method. The issue is that right after we store the object in the `alreadyIteratedSchemas`, we discard it and instead copy json properties to a fresh new object. This means that in the first iteration we get correctly converted schema but all other method calls are simply returning the existing schema which in our case stays empty.

The fix is easy as we only need to change:
```typescript
    convertedSchema = Object.assign({}, schema.json());
```
to
```typescript
    convertedSchema = Object.assign(convertedSchema, schema.json());
```

Without this fix, I'm not able to use modelina. I can provide the PR with a simple test case later on.

kamko:
**Description**
Simple change - correctly save converted schemas in `alreadyIteratedSchemas`

**Related issue(s)**
Fixes #548

### DIFF
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -55,10 +55,10 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
     if (alreadyIteratedSchemas.has(schemaUid)) {
       return alreadyIteratedSchemas.get(schemaUid) as AsyncapiV2Schema; 
     }
-    let convertedSchema = new AsyncapiV2Schema();
-    alreadyIteratedSchemas.set(schemaUid, convertedSchema);
-    convertedSchema = Object.assign({}, schema.json());
+
+    const convertedSchema = Object.assign(new AsyncapiV2Schema(), schema.json());
     convertedSchema[this.MODELGEN_INFFERED_NAME] = schemaUid;
+    alreadyIteratedSchemas.set(schemaUid, convertedSchema);
 
     if (schema.allOf() !== null) {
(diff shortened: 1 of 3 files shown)


jonaslagoni (approved):
LGTM, thanks @kamko!

PR #559, opened 2022-01-04
fix: special cases in enum values for C# generator

### LINKED ISSUE #503: Special cases in enum values fails to generate accurate enum keys for C# generator

ron-debajyoti:
**Description**
I used the helper function `FormatHelpers` similar to #497  to resolve for special characters in the enum renderer for C#. Also wrote down an additional test and updated the jest snapshot.

**Related issue(s)**
Fixes #503

### DIFF
--- a/test/generators/csharp/CSharpGenerator.spec.ts
+++ b/test/generators/csharp/CSharpGenerator.spec.ts
@@ -129,16 +129,6 @@ describe('CSharpGenerator', () => {
       enum: ['Texas', 'Alabama', 'California'],
     };
 
-    generator = new CSharpGenerator({ presets: [
-      {
-        enum: {
-          self({ content }) {
-            return content;
-          },
-        }
-      }
-    ] });
-
     const inputModel = await generator.process(doc);
     const model = inputModel.models['CustomEnum'];
     
--- a/test/generators/csharp/CSharpGenerator.spec.ts
+++ b/test/generators/csharp/CSharpGenerator.spec.ts
@@ -150,6 +150,36 @@ describe('CSharpGenerator', () => {
     expect(enumModel.result).toMatchSnapshot();
     expect(enumModel.dependencies).toEqual([]);
   });
+
+  test('should render enums with translated special characters', async () => {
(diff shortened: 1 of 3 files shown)


test/generators/csharp/CSharpGenerator.spec.ts
```diff
@@ -150,6 +150,36 @@ describe('CSharpGenerator', () => {
...
+    
+    generator = new CSharpGenerator({ presets: [
+      {
+        enum: {
+          self({ content }) {
+            return content;
+          },
+        }
+      }
+    ]});
```
jonaslagoni, line 169:
```suggestion

```

You don't need any custom generator here, so this is redundant 🙂 It will just [use the default generator](https://github.com/asyncapi/modelina/blob/7e278f9c6a628b8fe23125f22f3ea81bcaf7029a/test/generators/csharp/CSharpGenerator.spec.ts#L6).

jonaslagoni (changes requested):
Only one comment 🙂 👍

test/generators/csharp/CSharpGenerator.spec.ts
ron-debajyoti: Ohh right. I forgot to check the `beforeEach` function

jonaslagoni (approved):
LGTM 👍

jonaslagoni:
@all-contributors please add @ron-debajyoti for code, test

PR #564, opened 2022-01-05
fix: model dependencies not needed for C# generator

jonaslagoni:
**Description**
There is no need for model dependencies as if they are within the same package, they automatically have access to them.

### DIFF
--- a/src/generators/csharp/CSharpGenerator.ts
+++ b/src/generators/csharp/CSharpGenerator.ts
@@ -50,13 +50,8 @@ export class CSharpGenerator extends AbstractGenerator<CSharpOptions, CSharpRend
     }
 
     const outputModel = await this.render(model, inputModel);
-    const modelDependencies = model.getNearestDependencies().map((dependencyModelName) => { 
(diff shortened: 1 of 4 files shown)


magicmatatjahu:
@jonaslagoni There will be option to have models in seperate packages? Is there a point in creating an issue for this? Also I see that tests are failing on `macos`.

jonaslagoni:
> @jonaslagoni There will be option to have models in seperate packages? Is there a point in creating an issue for this? Also I see that tests are failing on `macos`.

No, for complex model rendering (having models in separate packages), you need to use the normal `generate`  function and build the desired output yourself. There you still have access to all the dependant models, so you can create your own mapping of what models go where 🙂 At least until we see a strong reason to support this natively. 

The `generateCompleteModels` are assuming that all models are within the same package for simplicity 🙂

magicmatatjahu (approved):

magicmatatjahu:
@jonaslagoni You should create issue for case when someone want render model not in this same namespace, but I'm not expert in C# so I don't know 😅

jonaslagoni:
> @jonaslagoni You should create issue for case when someone want render model not in this same namespace, but I'm not expert in C# so I don't know

Think it is better if we let use-case drive that change, as I dont see any benefits in having it, I cant argue why we would want it 😅

Of course, others that encounter is encouraged to create a feature request 😄

Issue #581, opened 2022-01-13
Assign optional/non required arguments in the constructor if provided


mahakporwal02:
#### Reason/Context
- We can add an enhancement to javascript generator to assign optional/non required arguments in the constructor during object formation.
- Constructor accept arguments and set it to required member variables. But currently it only set values for only required arguments.
- For example:
If email is non required property - 
```
const jsonSchemaDraft7 = {
  $schema: 'http://json-schema.org/draft-07/schema#',
  type: 'object',
  additionalProperties: false,
  properties: {
    email: {
      type: 'string',
      format: 'email'
    }
  }
}
```
The model generated would be - 
```
class Root {
  email;
  constructor(input) {
  }
  get email() { return this.email; }
  set email(email) { this.email = email; }
}
```
After adding the enhancement the model generated would be -
```
class Root {
  email;
  constructor(input) {
  	if (input.hasOwnProperty(email)) {
  		this.email = input.email;
  	}
  }
  get email() { return this.email; }
  set email(email) { this.email = email; }
}
```

ritik307:
@derberg @fmvilas @mahakporwal02 I would like to work on this issue

mahakporwal02:
@ritik307 Apologies, didn't see your comment until now.

PR #586, opened 2022-01-14
feat: assign optional/non required arguments in the constructor if pr…

mahakporwal02:
Enhancement to javascript generator to assign optional/non required arguments in the constructor during object formation.

**Related issue(s)**
#581


mahakporwal02:
@jonaslagoni Kindly review the PR.

examples/generate-javascript-models/__snapshots__/index.spec.ts.snap
```diff
@@ -6,7 +6,9 @@ Array [
   email;
 
   constructor(input) {
-  
+    if (input.hasOwnProperty(email)) {
```
jonaslagoni, line 9: This will throw an error when used, as there is no variable called `email` 🙂

jonaslagoni (changes requested):

examples/generate-javascript-models/__snapshots__/index.spec.ts.snap
jonaslagoni: Please do yes 😄

jonaslagoni (approved):
LGTM! 🎉

PR #594, opened 2022-01-21
feat: enable Typescript Input

### LINKED ISSUE #187: Enable TS input
#### Reason/Context
It would be awesome to take TS code and be able to output models in any other language.

#### Description
There already exist libraries we can leverage to minimize the effort to integrate this where we could use the already existing JSON Schema process or create a custom one.

Link shared by Kevin Swiber: https://github.com/YousefED/typescript-json-schema

ron-debajyoti:
**Description**

- Added TypescriptInputProcessor along with options to allow processing and generation of json schemas based on options like `uniqueNames`, `compilerOptions`
- Added tests for TypescriptInputProcessor
- Added 2 usage examples `java-from-typescript-type` and `java-from-typescript-type-with-options`.

**Related issue(s)**

Fixes #187


ron-debajyoti:
Hey @jonaslagoni ! 

I made this draft PR for issue #187 . Please check it out and provide your feedback on whether I'm heading in the right direction. There's still a lot of work left like writing proper tests and adding more implementation features based on the [library](https://github.com/YousefED/typescript-json-schema) I'm using.

jonaslagoni (commented):
Nice progress! 🎉 

> I made this draft PR for issue #187 . Please check it out and provide your feedback on whether I'm heading in the right direction. There's still a lot of work left like writing proper tests and adding more implementation features based on the library I'm using.

I just realized that we run into a bit of a problem here 😄 

Currently, the input type has always been one file, or rather one JSON object which of course wont fit here. So we need to adapt the input processing to accommodate these new input types. 

Would you like to start the issue? 🙂

src/processors/TypeScriptInputProcessor.ts
```diff
@@ -0,0 +1,53 @@
...
+      const generator = buildGenerator(program, TypeScriptInputProcessor.settings);
+      if (generator) {
+        return generateSchema(program, Type, TypeScriptInputProcessor.settings, [], generator);
+      }
+    }
+  
+    return generateSchema(program, Type, TypeScriptInputProcessor.settings);
+  }
+
+  process(basePath: string, listofTypes: Array<string>, multiple = false): CommonInputModel {
```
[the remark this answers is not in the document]
ron-debajyoti: Yeah, I was wondering that as well. Do you think we should just pass the Type as a parameter by the user?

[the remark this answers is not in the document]
jonaslagoni: I think the most optimal way forward would be to support a single string input, for now 🙂 

```
Single string input -> convert to JSON Schema (or multiple, dont know how it works) -> Use JSON Schema processor -> Return common models
```

And then in the new issue, track improved input types. How does that sound to you?

ron-debajyoti:
Sorry @jonaslagoni for the late reply,  I was kinda busy with my university coursework. I'm also going to reply to the comments above here.

> Currently, the input type has always been one file, or rather one JSON object which of course wont fit here. So we need to adapt the input processing to accommodate these new input types.

Yeah so is this the case for handling the whole Typescript project? For that, we need to use the `tsconfig.json` of the project and then pass the required TYPE, as described [here](https://github.com/YousefED/typescript-json-schema).  Currently, to just start working I tested it using a type declaration example.

Additionally, I'm trying to understand what more input types are we going to accept here? Since we either have to get the individual `type.ts` files or the `tsconfig.json` file, it was easier to just consider a base path as a `string`. If I misunderstood something please do mention.

jonaslagoni:
> Yeah so is this the case for handling the whole Typescript project? For that, we need to use the `tsconfig.json` of the project and then pass the required TYPE, as described [here](https://github.com/YousefED/typescript-json-schema). Currently, to just start working I tested it using a type declaration example.
> 
> Additionally, I'm trying to understand what more input types are we going to accept here? Since we either have to get the individual `type.ts` files or the `tsconfig.json` file, it was easier to just consider a base path as a `string`. If I misunderstood something please do mention.

My recommendation is to start small, what is the absolute minimum we need to support in order to say we have this feature. So for now don't worry about projects, but a single file 🙂 (unless the library supports providing the content of a file that would be better 😄)

I would suggest you took a look at how you can change the current processor line to follow this format:

```ts
process(input: any): CommonInputModel {
   if(typeof input !== 'string') throw error;
   const pathToFile = input;
```

As I dont know the library in and out, try to see what is possible and feel free to ping me with what worked and didnt 🙂

src/processors/TypeScriptInputProcessor.ts
```diff
@@ -0,0 +1,63 @@
...
+  
+    return TJS.generateSchema(program, typeRequired, TypeScriptInputProcessor.settings);
+  }
+
+  shouldProcess(input: Record<string, any>): boolean {
+    if (typeof input !== 'object' || typeof input.basePath !== 'string') {return false;}
+    return true;
+  }
+
+  process(input: Record<string, any>): Promise<CommonInputModel> {
```
jonaslagoni, line 42: You can use `input: any` here 🙂 We will just have to change the abstract processor as well 🙂

ron-debajyoti:
@jonaslagoni  So I guess this is the absolute base case of the implementation where I pass a `.ts` file as `{ baseFile : ourFile }` as input `Record`, and it generates and adds all the types present in the file to our common model. 

There's a lot more features to add to this, like : 
- generate selected types 
- generate types from a whole project using `tsconfig.json`.....
and many more to explore. Let me know your views and if possible we can have a discussion. 

Also let me know if I need to improve on the documentation of the current implementation.

test/processors/TypeScriptInputProcessor.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    test('should throw error for empty input', async () => {
+      const processsor = new TypeScriptInputProcessor();
+      await expect(processsor.process({})).rejects.toThrow('Input is not of the valid file format');
+    });
+
+    test('should process input', async () => {
+      const processsor = new TypeScriptInputProcessor();
+      const commonModel = await processsor.process({ baseFile });
+      expect(commonModel).toMatchSnapshot();
+    });
```
jonaslagoni, line 39: This has nothing to do with should process 🤔

test/processors/__snapshots__/TypeScriptInputProcessor.spec.ts.snap
```diff
@@ -0,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`TypeScriptInputProcessor process() should be able to process input 1`] = `
+CommonInputModel {
+  "models": Object {},
```
jonaslagoni, line 5: I dont see it generating any models 🤔

jonaslagoni (changes requested):

test/processors/TypeScriptInputProcessor.spec.ts
ron-debajyoti: Yeah i should have used `shouldProcess()` method

test/processors/__snapshots__/TypeScriptInputProcessor.spec.ts.snap
ron-debajyoti: The snapshot file doesn't save objects :confused:

jonaslagoni: Ahhh right, cause we are mocking it 😅

jonaslagoni (changes requested):
Looking good @ron-debajyoti 💪 

A couple of things are missing until we can merge:
- We need to add it as a input to the input processor: https://github.com/asyncapi/modelina/blob/8a79fc09c97ec195586946fa85fc1ee1b24f4046/src/processors/InputProcessor.ts#L19
- We need an example showing that it works similar to how we do for AsyncAPI inputs: https://github.com/asyncapi/modelina/tree/master/examples/asyncapi-from-object
  - Check out https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#adding-examples how to add examples 🙂 
-  We need to adapt the following documentation files:
  - https://github.com/asyncapi/modelina#features where we add TypeScript files as supported input
  - Add a new [usage section](https://github.com/asyncapi/modelina/blob/master/docs/usage.md#usages) that we can link the example to 🙂

examples/java-from-typescript-type/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    processorOptions: {
+      typescript: {
+        uniqueNames: true,
+        compilerOptions: {
+          strictNullChecks: true
+        }
+      }
+    }
+  }
+);
```
jonaslagoni, line 17: Not sure I truly like having two generations within the same example 🤔 

Would it not make more sense to have two instead?

docs/usage.md
```diff
@@ -70,6 +70,16 @@ The OpenAPI input processor expects that the property `openapi` is defined in or
 
 The response and request payloads, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
 
+## Generate models from TypeScript type files
+
+Currently, we support generating models from a TypeScript type file.
+
+- [Generate Java model from a TypeScript file](../examples/java-from-typescript-type/)
```
jonaslagoni, line 77: Maybe it makes sense to split them? 🤔 
```suggestion
- [Generate Java model from a TypeScript file](../examples/java-from-typescript-type/)
- [Generate Java model from a TypeScript file with options](../examples/java-from-typescript-type-with-options/)
```

jonaslagoni (changes requested):
Got a few comments, but overall it looks great 👍 

@magicmatatjahu think it's time you take a look if you want to, as we have reached the final state before merging.

examples/java-from-typescript-type/index.ts
ron-debajyoti: I was confused on this tbh. The process and final outcome of both of them are the same, they only differ by the input `options` parameter.

I can split into two use cases if it's easier for people to understand.

docs/usage.md
```diff
@@ -70,6 +70,16 @@ The OpenAPI input processor expects that the property `openapi` is defined in or
...
+
+The TypeScript input processor expects that the typescript file and base directory where it's present, is passed as input, in order to process the types accurately.
+
+The response and request payloads, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
```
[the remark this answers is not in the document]
ron-debajyoti: Sorry that's a mistake from my part.

docs/usage.md
ron-debajyoti: Splitting it into two examples.

jonaslagoni:
> @magicmatatjahu this has been pending for review for many days

Life of OpenSource unfortunately 😄 

@magicmatatjahu do you want to review it, or can we go ahead?

magicmatatjahu (commented):
Only one suggestion :) Rest lgtm. Awesome work!

src/processors/TypeScriptInputProcessor.ts
```diff
@@ -0,0 +1,91 @@
...
+    // obtain generated schema
+    const generatedSchemas = this.generateJSONSchema(baseFile, '*', options?.typescript);
+    if (generatedSchemas) {
+      for (const schema of generatedSchemas) {
+        const commonModels = JsonSchemaInputProcessor.convertSchemaToCommonModel(schema as Record<string, any>);
+        common.models = {...common.models, ...commonModels };
+      }
+    }
+    return Promise.resolve(common);
+  }
```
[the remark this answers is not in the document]
ron-debajyoti: We don't `await` for any of the internal functions called inside `process()`, so specifying `async` at the beginning is going to throw errors.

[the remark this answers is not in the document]
jonaslagoni: @ron-debajyoti just remove `async` keyword, it is not needed 🙂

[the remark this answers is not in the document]
magicmatatjahu: but you return a Promise anyway. `async` doesn't necessarily mean you have to `await`, `async` automatically creates a Promise for you, that's the point.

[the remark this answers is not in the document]
ron-debajyoti: > just remove async keyword, it is not needed slightly_smiling_face This definitely simplifies the code and avoid unnecessary promises.

We have to return a promise because `AbstractInputProcessor.render()` specifies to return a promise.



> but you return a Promise anyway. async doesn't necessarily mean you have to await, async automatically creates a Promise for you, that's the point.

Perhaps I worded it wrongly. What I meant was specifying the method as `async` without any `await` expressions would throw this error `` Async method 'process' has no 'await' expression. `` by eslint. A quick fix would be to just call one of the underlying function expressions with `await`.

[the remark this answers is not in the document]
magicmatatjahu: > Perhaps I worded it wrongly. What I meant was specifying the method as async without any await expressions would throw this error Async method 'process' has no 'await' expression. by eslint. A quick fix would be to just call one of the underlying function expressions with await.

I forgot about eslint 😆  Ok, so we can leave it is.

magicmatatjahu (approved):

jonaslagoni (approved):
LGTM 👍

PR #596, opened 2022-01-23
feat: add un/marshal JSON preset for Java

### LINKED ISSUE #488: Add helper functions to un/marshal data models for Java
#### Reason/Context
We need some helper functions to be able to marshal the data model instances to JSON and unmarshal JSON to instances of the data model.

#### Description
The reason we need to write this our self is that we add properties to data models which are technically not properties directly, an example here is `additionalProperties`, as it is rendered as a property in the data model, but when you want to convert an instance of the data model to bytes or JSON for a payload, it should not be used as is but unwrapped.

ritik307:
**Description**
- Added un/marshalling in Java inside `src/generators/java/presets/CommonPreset.ts`

**Related issue(s)**

Fixed #488


ritik307:
hi @jonaslagoni, Kindly look into my PR and let me know if I am heading in the right direction or not. 🙂

ritik307:
@jonaslagoni I am currently facing errors while writing a test for un/marshalling in Java and not able to figure out what the error kindly look into them

test/generators/java/presets/__snapshots__/CommonPreset.spec.ts.snap
```diff
@@ -66,6 +66,28 @@ exports[`JAVA_COMMON_PRESET should render accurately when there is no additional
     }
     return o.toString().replace(\\"\\\\n\\", \\"\\\\n    \\");
   }
+
+  @Override
+  public Map<String, Object> marshal() {
```
jonaslagoni, line 71: The marshaling method should return the JSON representation in string format 🙂 It should function similar to what `JSON.stringify` does in JS 🙂

test/generators/java/presets/__snapshots__/CommonPreset.spec.ts.snap
```diff
@@ -66,6 +66,28 @@ exports[`JAVA_COMMON_PRESET should render accurately when there is no additional
...
+    Map<String, Object> map = new HashMap<>();
+      \\"requiredProp\\": requiredProp,
+      \\"stringProp\\": stringProp,
+      \\"numberProp\\": numberProp,
+      \\"booleanProp\\": booleanProp,
+      \\"arrayProp\\": arrayProp
```
jonaslagoni, line 77: This does not look like proper Java syntax 🙂

src/generators/java/presets/CommonPreset.ts
```diff
@@ -110,6 +160,8 @@ export const JAVA_COMMON_PRESET: JavaPreset = {
       const shouldContainEqual = options.equal === undefined || options.equal === true;
       const shouldContainHashCode = options.hashCode === undefined || options.hashCode === true;
       const shouldContainToString = options.classToString === undefined || options.classToString === true;
+      const shouldContainMarshal = options.marshal === undefined || options.marshal === true;
+      const shouldContainUnmarshal = options.unmarshal === undefined || options.unmarshal === true;
```
jonaslagoni, line 164: Here you split it up into two options, but later you only use one? 🤔

jonaslagoni:
Definitely in the right direction 👍 Just keep iterating 🙂

ritik307:
@jonaslagoni I have implemented the changes you have mentioned. Kindly look at them and let me know if any more changes are required. :)

src/generators/java/presets/CommonPreset.ts
```diff
@@ -97,6 +98,94 @@ private String toIndentedString(Object o) {
...
+  if (model.additionalProperties !== undefined) {
+    propertyKeys.push(getUniquePropertyName(model, DefaultPropertyNames.additionalProperties));
+  }
+  const unmarshalProperties = propertyKeys.map(prop => {
+    const renderedPropertyName = renderer.nameProperty(prop);
+    const renderPropertyType = renderer.renderType(model);
+    return `${renderPropertyType} ${renderedPropertyName}`;
+  });
+
+  return `${renderer.renderAnnotation('Override')}
```
jonaslagoni, line 180: What internal method are you overriding? 🤔

src/generators/java/presets/CommonPreset.ts
```diff
@@ -97,6 +98,94 @@ private String toIndentedString(Object o) {
...
+    // const marshalCode = `json += \`"${prop}": ${propMarshalCode}\`;`;
+    if(properties[prop].type?.toString() === "array"){
+      return `if(jsonObject.has("${formattedPropertyName}")) {
+        JSONArray jsonArray = jsonObject.getJSONArray("${formattedPropertyName}");
+        String[] ${formattedPropertyName} = new String[jsonArray.length()];
+        for(int i = 0; i < jsonArray.length(); i++) {
+          ${formattedPropertyName}[i] = jsonArray.getString(i);
+        }      
+        result.${setterFunction}(${formattedPropertyName});
+        `;
```
jonaslagoni, line 151: You are missing an end `}` character to finish the if statement.

src/generators/java/presets/CommonPreset.ts
```diff
@@ -97,6 +98,94 @@ private String toIndentedString(Object o) {
...
+/**
+ * Render `marshal` function based on model's properties
+ */
+function renderMarshalling({ renderer, model }: {
+  renderer: JavaRenderer,
+  model: CommonModel,
+}): string {
+  const properties = model.properties || {};
+  const propertyKeys = [...Object.keys(properties)];
+  return `${renderer.renderAnnotation('Override')}
```
jonaslagoni, line 124: What internal method are you overriding? 🤔

src/generators/java/presets/CommonPreset.ts
```diff
@@ -110,16 +199,24 @@ export const JAVA_COMMON_PRESET: JavaPreset = {
       const shouldContainEqual = options.equal === undefined || options.equal === true;
       const shouldContainHashCode = options.hashCode === undefined || options.hashCode === true;
       const shouldContainToString = options.classToString === undefined || options.classToString === true;
+      const shouldContainMarshal = options.marshal === undefined || options.marshal === true;
 
-      if (shouldContainEqual === true || shouldContainHashCode === true) {
+      if (shouldContainEqual === true || shouldContainHashCode === true || shouldContainMarshal === true) {
         renderer.addDependency('import java.util.Objects;');
+        renderer.addDependency('import java.util.stream;');
```
jonaslagoni, line 206: You are missing a `Map` import 🙂

jonaslagoni: Also please make a new if statement for adding dependencies for `shouldContainMarshal`. As you do not want these imports present if `shouldContainMarshal == false`.

jonaslagoni (changes requested):
Have a couple of remarks 🙂 

Remember your code smells from sonarcloud 🙂

ritik307:
@jonaslagoni everything is working now 😊. I ran `npm run lint:fix` and it was getting warning `security/detect-object-injection`. So as of now, I have added a command to skip this warning line as adding `parseInt(prop,10)` isn't helping and generating more errors.

src/generators/java/presets/CommonPreset.ts
```diff
@@ -97,6 +98,74 @@ private String toIndentedString(Object o) {
...
+}
+
+function renderUnmarshalProperties(renderer: JavaRenderer, model:CommonModel) {
+  const properties = model.properties || {};
+  const propertyKeys = [...Object.keys(properties)];
+  const unmarshalProperties = propertyKeys.map(prop => {
+    const formattedPropertyName = renderer.nameProperty(prop,model);
+    const setterFunction = `set${formattedPropertyName.charAt(0).toUpperCase()}${formattedPropertyName.slice(1)}`;
+    // eslint-disable-next-line
+    const getType = `jsonObject.get${properties[prop].type?.toString().charAt(0).toUpperCase()}${properties[prop].type?.toString().slice(1)}`;
```
jonaslagoni, line 134: I am assuming these lines are the problems to eslint complaining?

```suggestion
    const getType = `jsonObject.get${FormatHelpers.upperFirst(properties[prop].type?.toString())}`
```

Was it the optional chaining of `properties[prop].type?`? 

In case it was, at the beginning of the map callback, you can add the following and then change all calls of ` properties[prop].type?` to `properties[prop].type`.

```ts
const propModel = properties[prop];
if(propModel.type === 'undefined') {
  Logger.error(`Could not render unmarshal for property ${prop}`);
  return; 
}
```

src/generators/java/presets/CommonPreset.ts
```diff
@@ -110,16 +179,28 @@ export const JAVA_COMMON_PRESET: JavaPreset = {
       const shouldContainEqual = options.equal === undefined || options.equal === true;
       const shouldContainHashCode = options.hashCode === undefined || options.hashCode === true;
       const shouldContainToString = options.classToString === undefined || options.classToString === true;
+      const shouldContainMarshal = options.marshal === undefined || options.marshal === true;
```
jonaslagoni, line 182: Since marshaling is not as common as overwriting Equal and hashcode methods I suggest we only use the following:

```suggestion
      const shouldContainMarshal = options.marshal === true;
```

This will also automatically fix the blackbox tests 🙂

jonaslagoni (changes requested):
Small changes requested, looking good! 💪

We need to enlighten the user of the new preset, so they can start using it!

We need to adapt our Java language documentation: https://github.com/asyncapi/modelina/blob/master/docs/languages/Java.md

Here you could add something along the lines of:

```
## Include JSON marshaling and unmarshaling methods
Sometimes you just want to convert your class to JSON without the use of annotations such as Jackson.

Check out this [example for a live demonstration](https://github.com/asyncapi/modelina/blob/master/examples/java-generate-marshal).

External dependencies
- Requires [org.json package](https://search.maven.org/artifact/org.json/json/20211205/bundle) to work

```

Next, we need to create an example to show exactly how a user can enable this option (as well as always ensuring it works). We do that by adding the following example: https://github.com/asyncapi/modelina/blob/master/examples/java-generate-marshal

You can copy this template: https://github.com/asyncapi/modelina/tree/master/examples/TEMPLATE

And then adapt the index.ts file to contain the generator you have here: https://github.com/asyncapi/modelina/blob/981fd9af3b058c2c7ed136a0f92eec9f96aef40d/test/generators/java/presets/CommonPreset.spec.ts#L139

Then there are a few mentions in the documentation that needs to be changed 🙂 

Let me know if you have any questions!

jonaslagoni:
> hi @jonaslagoni , I have made changes but I am facing an error kindly look at them. ss1

That is indeed weird. Have you tried printing the content or debugging the test to see what is actually inside  `classModel.dependencies`? 


Is it just some hidden character not matching or?

ritik307:
@jonaslagoni This error is actually occurring because I changed the code from 

> const shouldContainMarshal = options.marshal === undefined || options.marshal === true;

to 

> const shouldContainMarshal = options.marshal === true;

But if I remove the following imports the code works fine I think it is not able to process multiple imports.

> expect(classModel.dependencies.includes('import java.util.stream;')).toEqual(true);
> expect(classModel.dependencies.includes('import org.json.JSONObject;')).toEqual(true);

jonaslagoni:
> But if I remove the following imports the code works fine I think it is not able to process multiple imports.

It definitely should be 😄 

Do you mind trying to print out what is inside the array? 🙂

jonaslagoni:
Here we are testing two dependencies: https://github.com/asyncapi/modelina/blob/7b2a33f65c06795c36c1b995334665a327068698/test/generators/java/presets/JacksonPreset.spec.ts#L39

ritik307:
> 

ss1

ritik307:
> Do you mind trying to print out what is inside the array `console.log(classModel.dependencies)`?
@jonaslagoni  Here is the screenshot
ss1

src/generators/java/presets/CommonPreset.ts
```diff
@@ -110,15 +182,26 @@ export const JAVA_COMMON_PRESET: JavaPreset = {
       const shouldContainEqual = options.equal === undefined || options.equal === true;
       const shouldContainHashCode = options.hashCode === undefined || options.hashCode === true;
       const shouldContainToString = options.classToString === undefined || options.classToString === true;
+      const shouldContainMarshal = options.marshal === true;
```
jonaslagoni, line 185: @ritik307 try double-checking what options you pass in your test and see if it matches this one 😄

ritik307: @jonaslagoni  Sorry to bother you again but I am not able to figure out where to check for the options :(

jonaslagoni: Don't worry about it, what I mean by options, is the `JavaCommonPresetOptions`, and what was the property you added there vs what are you using here 🙂

jonaslagoni: Hint, look at https://github.com/asyncapi/modelina/pull/596/files#diff-a1e439279cc6c3345bcb667a54b404bc408869de67235c3133af054088cfe854R12

test/generators/java/presets/CommonPreset.spec.ts
```diff
@@ -130,5 +135,27 @@ describe('JAVA_COMMON_PRESET', () => {
...
+              marshalling: true,
+            }
+          }] 
+        }
+      );
+      const inputModel = await generator.process(doc);
+      const model = inputModel.models['Clazz'];
+      
+      const classModel = await generator.renderClass(model, inputModel);
+      expect(classModel.dependencies.includes('import java.util.Map;')).toEqual(true);
```
jonaslagoni, line 156: Why are you no longer testing snapshot here? 🤔

docs/languages/Java.md
```diff
@@ -57,3 +57,12 @@ Check out this [example for a live demonstration](../../examples/java-generate-j
 To generate Java data models with Jackson annotation using `JAVA_JACKSON_PRESET` option.
 
 Check out this [example for a live demonstration](../../examples/java-generate-jackson-annotation).
+
+## Include JSON marshaling and unmarshaling methods
+
+Sometimes you just want to convert your class to JSON without the use of annotations such as Jackson.
+
+Check out this [example for a live demonstration](https://github.com/asyncapi/modelina/blob/master/examples/java-generate-marshal).
```
jonaslagoni, line 65: Don't see the actual example anywhere? Please re-read https://github.com/asyncapi/modelina/pull/596#pullrequestreview-874221410

jonaslagoni (changes requested):
Small comments, nearly there, looking great 💪

ritik307:
@jonaslagoni Added example for marshalling. Kindly review :)

jonaslagoni (approved):
LGTM, awesome @ritik307 🎉

jonaslagoni:
@all-contributors please add @ritik307 for code, test, example

PR #621, opened 2022-02-08
feat: add JSDoc style comment support to Typescript generator

gabormagyar:
Might need tests; tests for comment generation are already included just not for them being included. Don't think it needs docs as the feature is automatic and self explanatory.

Make sure to filter whitespace changes when viewing diff.

PR #622, opened 2022-02-08
feat: add option to generate named exports from Typescript generator

gabormagyar:
This probably needs tests, I'm not sure when I will have time to add them. Feel free to take over if someone wants to.

It also needs ~~(minor)~~ documentation changes about the added generator option.
Correction: `TypeScriptRenderCompleteModelOptions` is wholly undocumented if I'm not mistaken, should probably be added at some point


jonaslagoni (changes requested):
This is great!

Yes, we definitely need some tests, examples, and documentation to ensure this works as intended before it can be merged.

It seems you might have auto-formatting enabled in your development environment making unnecessary changes that make it harder to review, do you mind disabling it and re-commit the intended changes? 🙂

You can try to reach out on slack to see if you can find someone to help you finish it 🙂

gabormagyar:
Auto formatting removed

gabormagyar:
@jonaslagoni Please advide if you think there's anything else that needs to be done. I personlally think this doesn't need a separate example.

jonaslagoni (changes requested):
Looks good tbh 👍 

We do need to update the documentation though: https://github.com/asyncapi/modelina/blob/master/docs/languages/TypeScript.md#rendering-complete-models-to-a-specific-module-system

What's the reason you think an example is unnecessary? 🙂

gabormagyar:
@jonaslagoni Added a note to the TypeScript docs.

I don't think this is a big enough thing to add separate examples for. If we add examples for every combination of every possible option it will be a lot of examples. What do you think?

gabormagyar:
Also interestingly the named exports don't really play nice with the generated description comments from the other PR (#621). It still generates valid code, it just looks ugly: the comments are between the export keyword and the node.

I think it's fine for now, but in a future PR I will fix it so that the export keyword is generated through the preset system with priority over any comments. I think this will work but I'm open to suggestions, this is a rather tricky interaction.

gabormagyar:
Fixed, it's not pretty but it works and I do believe it's robust

src/generators/typescript/TypeScriptGenerator.ts
```diff
@@ -1,10 +1,11 @@
...
-  AbstractGenerator, 
+import {
+  AbstractGenerator,
   CommonGeneratorOptions,
   defaultGeneratorOptions
 } from '../AbstractGenerator';
-import { CommonModel, CommonInputModel, RenderOutput } from '../../models';
+import { CommonModel, CommonInputModel, RenderOutput, PresetWithOptions } from '../../models';
 import { TypeHelpers, ModelKind, CommonNamingConvention, CommonNamingConventionImplementation } from '../../helpers';
+import { TS_EXPORT_KEYWORD_PRESET } from './presets/ExportKeywordPreset'
```
Samridhi-98, line 8: Semicolon missing please correct 🙂
Do run `npm run lint:fix` before pushing changes 🙂

Samridhi-98 (changes requested):
Please fix all the linting errors🙂

src/generators/typescript/TypeScriptGenerator.ts
```diff
@@ -39,15 +41,29 @@ export class TypeScriptGenerator extends AbstractGenerator<TypeScriptOptions,Typ
...
+    // Add preset that adds the `export` keyword if needed
+    if (
+      moduleSystem === 'ESM' &&
+      exportType === 'named' &&
+      (originalPresets[0] !== TS_EXPORT_KEYWORD_PRESET ||
+        (Object.prototype.hasOwnProperty.call(originalPresets[0], 'preset') &&
+          (originalPresets[0] as PresetWithOptions).preset !== TS_EXPORT_KEYWORD_PRESET))
+    ) {
+      this.options.presets = [TS_EXPORT_KEYWORD_PRESET, ...originalPresets];
+    }
```
jonaslagoni, line 65: I see your problem yea 😅 I guess we are hitting the limits of what string presets can handle here.

Interesting workaround adding it as a preset and just ensuring it is the first one called 🤔 

I wonder if we cant make the check a bit easier to understand 🤔 
```
      (originalPresets[0] !== TS_EXPORT_KEYWORD_PRESET ||
        (Object.prototype.hasOwnProperty.call(originalPresets[0], 'preset') &&
          (originalPresets[0] as PresetWithOptions).preset !== TS_EXPORT_KEYWORD_PRESET))
```
Gonna think about that one for a bit.

cc @magicmatatjahu

gabormagyar: This bit detects the `presets: [TS_EXPORT_KEYWORD_PRESET]` syntax
 ```typescript
      (originalPresets[0] !== TS_EXPORT_KEYWORD_PRESET ||
```
This bit detects the `presets: [ { preset: TS_EXPORT_KEYWORD_PRESET, options?: { ... } } ]` syntax, that we don't add but the user might. I thought for completeness' sake it's better to include this and make it look a bit more complicated, but I'd be interested to hear what you think.
```typescript
        (Object.prototype.hasOwnProperty.call(originalPresets[0], 'preset') &&
          (originalPresets[0] as PresetWithOptions).preset !== TS_EXPORT_KEYWORD_PRESET))
 ```

Comment (gabormagyar) [reply src/generators/typescript/TypeScriptGenerator.ts:65]: Might be just a matter of exporting this logic as a `hasPreset` or `isPreset` util function or something.

**Comment by gabormagyar** [comment]
@jonaslagoni What are next steps for this PR?

File: src/generators/typescript/TypeScriptGenerator.ts (hunk 2 above)
Comment (jonaslagoni) [reply src/generators/typescript/TypeScriptGenerator.ts:65]: I feel like this is just a bandage to the gun wound, cause eventually, we will end up in a situation where two presets cannot be used with each other because of how presets work 😅

That said, until that is fixed, I think this makes sense to introduce regardless 🤔

@magicmatatjahu @Samridhi-98 @gabormagyar what do you folks think?

Comment (Samridhi-98) [reply src/generators/typescript/TypeScriptGenerator.ts:65]: Yeah, I agree on this, it's just a temporary fix.

Comment (gabormagyar) [reply src/generators/typescript/TypeScriptGenerator.ts:65]: So can we merge it?

File: src/generators/typescript/TypeScriptGenerator.ts [hunk 3]
```diff
@@ -39,15 +41,29 @@ export class TypeScriptGenerator extends AbstractGenerator<TypeScriptOptions,Typ
...
+    const originalPresets = [...(this.options.presets ? this.options.presets : [])];
+
+    // Add preset that adds the `export` keyword if needed
+    if (
+      moduleSystem === 'ESM' &&
+      exportType === 'named' &&
+      (originalPresets[0] !== TS_EXPORT_KEYWORD_PRESET ||
+        (Object.prototype.hasOwnProperty.call(originalPresets[0], 'preset') &&
+          (originalPresets[0] as PresetWithOptions).preset !== TS_EXPORT_KEYWORD_PRESET))
+    ) {
```
jonaslagoni, line 63: @gabormagyar do you mind moving this to a helper function such as you suggested `hasPreset`, that way we don't clutter up this function, and in the future, we probably need this in other generators 🙂

src/generators/typescript/TypeScriptGenerator.ts
jonaslagoni: Left a comment 🙂

jonaslagoni (approved):
LGTM 👍 

@Samridhi-98 anything you want to add?

Samridhi-98 (approved):
LGTM 🎉🎉

PR #623, opened 2022-02-08
fix: change package.json `prepublishOnly` script to `prepare` script

gabormagyar:
Proposal: Change build script to run on `prepare` rather than `prepublishOnly`. The difference:

`prepublishOnly`:
> Run BEFORE the package is prepared and packed, ONLY on npm publish.

`prepare`:
> Run both BEFORE the package is packed and published, and on local npm install without any arguments. This is run AFTER prepublish, but BEFORE prepublishOnly.

The latter makes it easier to add the dependency from github (say from a fork).

sources:
https://docs.npmjs.com/cli/v8/using-npm/scripts#prepare-and-prepublish
https://stackoverflow.com/questions/40528053/npm-install-and-build-of-forked-github-repo/57829251#57829251


jonaslagoni:
Thanks for taking the time to provide the PR!

I struggle to see why you would need the build source files at all times during your development process (automatically generated when you use install dependencies), can you elaborate why you think that? 🙂 To me it sounds like an annoying extension of time that it will take before you get started making changes (because you need to wait for the build to finish).

gabormagyar:
I think you misunderstand, this does not run on `npm install` in the modelina package, but rather when `npm install` is run in a parent package that includes modelina.

If modelina is added via an npm dependency (the usual way) this does not matter (and doesn't run), but rather when it is added some other way (for example via a [github dependency](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#github-urls)) then this ensures that the build happens.

gabormagyar:
Closing as I suppose this is a matter of preference

PR #624, opened 2022-02-09
ci: only run workflows when necessary

jonaslagoni:
**Description**
Adds checks to blackbox tests and coverall to ensure they only run when necessary.

**Related issue(s)**
See https://github.com/asyncapi/.github/issues/116#issuecomment-1032521012

### DIFF
--- a/.github/workflows/blackbox-testing.yml
+++ b/.github/workflows/blackbox-testing.yml
@@ -12,28 +12,38 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
-      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/blackbox-testing.yml
```diff
@@ -12,28 +12,38 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
-      - name: Checkout repository
+      - if: "github.event.pull_request.draft == false &&!((github.actor == 'asyncapi-bot' && startsWith(github.event.pull_request.title, 'ci: update global workflows')) || (github.actor == 'asyncapi-bot' && startsWith(github.event.pull_request.title, 'chore(release):')) || (github.actor == 'allcontributors' && startsWith(github.event.pull_request.title, 'docs: add')))"
```
magicmatatjahu, line 15: Is there something like global condition on job? I don't think so that we need to check  `steps.should_run.outputs.shouldRun == 'true'` on every step

jonaslagoni: Good call 🤔 Let me check

magicmatatjahu (approved):
LGTM! :)

PR #634, opened 2022-02-14
feat: add date and date-time format support to Typescript generator

gabormagyar:


jonaslagoni (changes requested):
Looks good to me, however, we need to update the test here: https://github.com/asyncapi/modelina/blob/master/test/generators/typescript/TypeScriptRenderer.spec.ts

gabormagyar:
@jonaslagoni Added tests to test the new feature. Are any changes needed to marshalling code generation?

gabormagyar:
On a second thought this might not be desirable in all cases. This would only work right when used in tandem with unmarshalling/deserializing code that supports this same mapping based on the format field.

A better solution would be to add a config option that allows arbitrary format to TS type mapping. That would allow users to use a different date library as well for example.

If you agree I will close this PR and work on that instead. I need the mapping option for something else anyway.

jonaslagoni:
> On a second thought this might not be desirable in all cases. This would only work right when used in tandem with unmarshalling/deserializing code that supports this same mapping based on the format field.

It's a good catch yea 🤔

It's an interesting scenario, and not quite sure what the right approach is, the looser we are with what types are supported in the generator, the more difficult it is for preset to support them 🤔 

Do you have any approach in mind? 🙂

gabormagyar:
I think this is better solved on user side with presets. Closing.

Issue #628, opened 2022-02-10
Model representation in the form of Node in AST


magicmatatjahu:
As I'm the "author" of the presets, I see that I made a big mistake of treating them as middlewares which are supposed to return the changed content (as a string) and not as data (in meaning of the structure) that is later changed to the correct model in the given language (class/interface/structure etc).

It seems to me that we should move to representing a given model as a node in a typical AST. What does that mean? During the processing of presets, the user/developer would have access to the model representation as a structure (not a string!) and then, after each preset (btw. we should change the name to `hook`), this structure should be transformed into a valid model in the given language. Here is an example to better understand the idea:
 
```ts
interface JavaClassNode {
  name: string;
  modifiers: Array<'public', 'private', 'static' ...>;
  annotations: Record<string, {
    ...
  }>;
  interfaces: Array<Type>; // interfaces which implements given class
  extends: Type; // we can extend only by one class
  properties: Record<string, {
    modifiers: Array<'public', 'private', 'final', 'protected', ...>;
    type: Type,
    initialValue: string,
    ...
  }>;
  methods: Record<string, {
    modifiers: Array<'public', 'private', 'final', 'protected', ...>;
    arguments: Array<{ name: string, type: Type }>;
    returnType: Type;
    body: string; // I think that body of method still be of string type, but we can discuss about it
  }>;
  ... // other things like comments etc
}
```

then we will have hooks only for given type kind - in Java class, interface, enum (maybe record?) - and we can remove all presets for `properties`, `additionalProperties` etc, because it's bullshit and they are difficult to use. Inside hooks we will be able to change that internal data in structure (and what is important, add/remove/update new properties, methods, annotations etc), e.g.:

```ts
// java hooks
const hook1 = {
  class: ({ model }) => {
    model.modifiers.push('private');
    model.interfaces.push({ name: 'ISomeInterface' });
    model.annotations.push({ name: 'SomeAnnotation', value: ... });
    model.extends = { name: 'AnotherJavaClass' };
  },
  interface: () => {...},
  enum: () => {...},
}
const hook2 = {
  class: ({ model }) => {
    model.interfaces.push({ name: 'IAnotherInterface' });
    model.properties['someProperty'] = { modifiers: ['private'], type: { name: 'SomeType', initialValue = 'new SomeType()' } };
[...]

PR #639, opened 2022-02-14
feat: add Go JSON tag preset

jonaslagoni:
**Description**
This PR adds a basic preset for adding JSON tags to go models.

Fixes https://github.com/asyncapi/modelina/issues/299


examples/README.md
```diff
@@ -33,3 +33,7 @@ This directory contains a series of self-contained examples that you can use as
 - [generate-csharp-models](./generate-csharp-models) - A basic example to generate C# data models
 - [java-change-collection-type](./java-change-collection-type) - An example to render collections as List in Java.
 - [java-generate-hashcode](./java-generate-hashcode) - A basic example that shows how to generate models that overwrite the `hashCode` method
+- [go-generate-json-tags](./go-generate-json-tags) - A basic example on how to generate models with JSON Tags
+
+
+
```
smoya, line 39: nitpick: 

remove empty lines
```suggestion
```

src/generators/go/presets/JsonPreset.ts
```diff
@@ -0,0 +1,17 @@
...
+ * 
+ * @implements {GoPreset}
+ */
+export const GO_JSON_PRESET: GoPreset = {
+  struct: {
+    field({fieldName, content, type }) {
+      if (type === FieldType.field) {
+        return `${content} \`json:"${fieldName},omitempty"\``;
+      }
+      return `${content} \`json:"-"\``;
```
smoya, line 14: I would love to see this part being tested. What happens when the type is either `additionalProperty` or `patternProperties` ? You won't add the tag?

smoya (changes requested):
Nice addition @jonaslagoni !👏 👏 

I wrote a couple of comments on my review. I hope they make sense to you.

src/generators/go/presets/JsonPreset.ts
jonaslagoni: Correct, and this is because both of those two properties need to be unwrapped. And since the JSON tag don't support this, see https://github.com/golang/go/issues/6213, we need to customize methods instead. And this can be added later.

smoya: But the test can be added right now, isn't?

jonaslagoni: Yep, added!

smoya (approved):
LGTM 🚀 🌔

src/generators/go/presets/JsonPreset.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Preset which adds basic JSON Tags for fields
+ * 
+ * @implements {GoPreset}
+ */
+export const GO_JSON_PRESET: GoPreset = {
+  struct: {
+    field({fieldName, content, type }) {
+      if (type === FieldType.field) {
+        return `${content} \`json:"${fieldName},omitempty"\``;
```
magicmatatjahu, line 12: I have problem with that approach, because people should have possibility to add new tags to the existing one, so then they need to regex tags and retrieve existing one and merge with new... That preset does not work here and causes problems. Maybe we should have new one?

jonaslagoni: That... is a great question and something I did not consider. I think that might make sense yes 🤔 

Unless we want to wait until we face the use-case that a user wants to do it of course?

smoya: It is right to assume people will want to add more tags in some cases. Not only about JSON or other serialization mechanisms, but others like database-related, etc. So 👍 to what @magicmatatjahu says.

However, I would say that if it's hard to do at this PR, maybe it makes sense to create a feature request and work after? Just a suggestion.

magicmatatjahu: I just noticed this problem :) You can do as you see fit, however I know it @felixjung has a problem with the preset for description, so that can be "fixed" in this PR.

src/generators/go/renderers/StructRenderer.ts
```diff
@@ -21,18 +22,26 @@ type ${formattedName} struct {
...
-  field({ fieldName, field, renderer, type }) {
-    fieldName = renderer.nameField(fieldName, field);
+  async field({ fieldName, field, renderer, type }) {
+    const formattedFieldName = renderer.nameField(fieldName, field);
     let fieldType = renderer.renderType(field);
     if (type === FieldType.additionalProperty || type === FieldType.patternProperties) {
       fieldType = `map[string]${fieldType}`; 
     }
-    return `${ fieldName } ${ fieldType }`;
+    const fieldTags = await renderer.runFieldTagPreset(fieldName, field, type);
```
magicmatatjahu, line 39: I have problem with that line, why? If someone want to write `field` preset have to remember about running also `fieldTag`. I think that we should concatenate field and tag content inside `StructRenderer` rather than in preset. I hope you understand it :)

jonaslagoni: On the other hand, nothing is stopping someone from using the field preset to also define tags 🤔 For example if you add the field preset and return:
    
```ts
return `${ fieldName } ${ fieldType } \`example:"name"\``;
```

Using any of the presets that add tags, wont work 🤷

magicmatatjahu: But as we discussed then you have the problem of overwriting previously overwritten tags. Only concatenation of these tags would somehow make sense if we were doing it inside the renderer. Then people would write the preset for the field and tags separately and then concatenate them inside the source code, here https://github.com/asyncapi/modelina/blob/10aca3b10d985596241761f8976697a72dc360e1/src/generators/go/GoRenderer.ts#L75

magicmatatjahu: More and more cases where AST would be better 😄

jonaslagoni:
Use-case where AST would be better: https://github.com/asyncapi/modelina/issues/628

smoya:
What is the state of this PR @jonaslagoni ?

jonaslagoni:
Closed in favour of https://github.com/asyncapi/modelina/pull/867

Issue #648, opened 2022-02-16
Generate descriptions as comments in Go


felixjung:
#### Reason/Context

In Go it's standard practice to write godoc comments for types and struct fields. These get picked up by tooling etc.

Currently, Modelina only renders comments for enums and structs. The comments always follow the same template of `// ${formattedName} represents a ${formattedName} model.` In my personal opinion these do not add a lot of value.

Instead, it would be nice if Modelina rendered the `description` field from the AsyncAPI spec as comments for types and fields. This would provide developers with a 1:1 match between the generated code and the generated docs.

The `JAVA_DESCRIPTION_PRESET` implements a similar approach but for javadoc.

#### Description

- Add a new Go preset to render comments.
- Remove the old comments from the default Go renderer as they would conflict with the new renderers comments. This would generally make the default renderer more composable with other renderers.

- I don't think this would be a breaking change, as we are only talking about dropping the current comments and would provide everything else as opt-in.

PR #649, opened 2022-02-16
feat: render descriptions in Go comments via preset

felixjung:
**Description**

- This PR adds a new `GO_DESCRIPTION_PRESET` similar to the `JAVA_DESCRIPTION_PRESET` with the ability to render descriptions from specs as comments on types (structs and enums) and struct fields.
- The PR also removes the default preset's comments on structs and enums as these were conflicting and didn't seem to add much value.
- I also fixed a file name, which seemed had a typo.

**Related issue(s)**

- #648


src/generators/go/presets/DescriptionPreset.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (
+        type === FieldType.additionalProperty ||
+        type === FieldType.patternProperties
+      ) {
+        fieldType = `map[string]${fieldType}`;
+      }
+
+      const content = `${fieldName} ${fieldType}`;
+
+      return renderDescription({ renderer, content, item: field });
```
jonaslagoni, line 42: This can be simplified to:

```suggestion
      return renderDescription({ renderer, content, item: field });
```

jonaslagoni (changes requested):
One specific comment to what you added other then that it looks great 👍 

Next we need too add some documentation and examples 😄 

- We need to add documentation for how to add descriptions through the preset here: https://github.com/asyncapi/modelina/blob/master/docs/languages/Go.md which links to the new example, see below.
- We need to add an example, see: https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#adding-examples

felixjung:
Thanks for the review. I'll get to it tomorrow, I hope.

src/generators/go/presets/DescriptionPreset.ts
felixjung: Can you explain to me how you got to this simplification? Does `renderDescription` already fix the content for these specific fields? Haven't looked at the code since making the PR. 🙃

jonaslagoni:

PR #653, opened 2022-02-20
feat: added reserved keyword in go

### LINKED ISSUE #460: Go generates reserved keyword properties

Samridhi-98:
**Description**

- Added `constants.ts` in go
- Updated `gogenerator.ts`
- Updated `gorenderer.ts`

**Related issue(s)**

Fixes #460

### DIFF
--- a/src/generators/go/Constants.ts
+++ b/src/generators/go/Constants.ts
@@ -1,0 +1,31 @@
+export const RESERVED_GO_KEYWORDS = [
+  'break',
+  'case',
+  'chan',
+  'const',
+  'continue',
+  'default',
+  'defer',
+  'else',
+  'fallthrough',
+  'for',
+  'func',
+  'go',
+  'goto',
+  'if',
+  'import',
+  'interface',
+  'map',
+  'package',
+  'range',
+  'return',
+  'select',
+  'struct',
+  'switch',
+  'type',
+  'var'
+];
+
+export function isReservedGoKeyword(word: string): boolean {
+  return RESERVED_GO_KEYWORDS.includes(word);
+}
--- a/src/generators/go/GoGenerator.ts
+++ b/src/generators/go/GoGenerator.ts
@@ -15,21 +15,30 @@ import { Logger } from '../../utils/LoggingInterface';
  * The Go naming convention type
  */
 export type GoNamingConvention = {
-  type?: (name: string | undefined, ctx: { model: CommonModel, inputModel: CommonInputModel }) => string;
-  field?: (name: string | undefined, ctx: { model: CommonModel, inputModel: CommonInputModel, field?: CommonModel }) => string;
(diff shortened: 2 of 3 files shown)


Samridhi-98:
@jonaslagoni please review 🙂

jonaslagoni:
@Samridhi-98 you have a little code smell 🙂

src/generators/go/Constants.ts
```diff
@@ -0,0 +1,32 @@
+export const RESERVED_GO_KEYWORDS = [
+  'break',
+  'case',
+  'chan',
+  'const',
+  'continue',
+  'default',
+  'defer',
+  'else',
+  'enum',
```
smoya, line 10: `enum` is not a keyword in Go. See https://go.dev/ref/spec#Keywords.

test/generators/go/GoGenerator.spec.ts
```diff
@@ -5,6 +5,28 @@ describe('GoGenerator', () => {
   beforeEach(() => {
     generator = new GoGenerator();
   });
+
+  test('should not render reserved keyword', async () => {
+    const doc = {
+      $id: 'Address',
+      type: 'object',
+      properties: {
+        enum: { type: 'string' },
```
smoya, line 14: I'm a bit confused. Even though `enum` is not a real keyword in go, you added it to the list here: https://github.com/asyncapi/modelina/pull/653/files#r813939730 so I would expect this test to fail since `Enum` would never be possible as field name. 

Anyway, as mentioned earlier, `enum` is not a keyword. What do you think about generating a test that takes all the keywords defined at `RESERVED_GO_KEYWORDS` constant testing them one by one?

smoya (changes requested):
Nice addition @Samridhi-98 👏 👏 !

I left a couple of comments!

smoya:
@jonaslagoni would you mind reviewing the JS part of this? I did check the Go one (I mean that it makes sense from Go perspective and we test what we should). Dropped a couple of request changes about it.

src/generators/go/Constants.ts
Samridhi-98: Noted 🙂

test/generators/go/GoGenerator.spec.ts
Samridhi-98: Do you mean testing all 25 reserved keywords in a single test or creating a test for each reserved keyword?

smoya: I think one single test that tests all of them at once iterating them should do the job here. WDYT?

Samridhi-98: I agree, I have added the single iteration for all please review and let me know If I implemented it correctly or not 🙂

jonaslagoni (approved):
LGTM 👍 

@smoya anything to add? 🙂

PR #654, opened 2022-02-22
refactor: next starting point from master

jonaslagoni:
**Description**
This PR merges master into next.

PR #655, opened 2022-02-22
refactor: introduce new core models

jonaslagoni:
**Description**
This PR adds the two basic models discussed in #530 

**Related issue(s)**
Blocked by #530, #654


jonaslagoni:
> * we should use inline modifiers in the constructors to initialize all properties in the models (I wonder if that type should be also readonly using `public readonly` modifiers).

Why?

> * we should use generic for `originalInput` to define type of input (where possible):

Why? What the benefits of that? `StringModel` does not mean that the input was a string, it could be `{type: string}`

> we should avoid long name for props (and also for class names), for example ReferencedModel we can name as RefModel with ref property or for UnionModel the union prop is descriptive:

I would probably rather have them completely descriptive, as oftentimes short names lose it's meaning to different users. And when open source, that should be prefered in my opinion 🙂 

> what is the prop type responsible for in ConstrainedMetaModel?

It is the usage `type` of a model, i.e. if you want to use a model as a type of a parameter, property, etc, you can simply define it as `model.type`.

magicmatatjahu:
> Why? 🙂

Because it's a sugar syntax for manual definitions of properties:

```ts
export class MetaModel<T = any> {
  constructor(
    public name: string,
    public originalInput: T,
  ) {}
}

// vs

export class MetaModel<T = any> {
  constructor(
    name: string,
    originalInput: T,
  ) {
    this.name = name;
    this.originalInput = originalInput;
  }
}
``` 

> StringModel does not mean that the input was a string, it could be {type: string}, so I dont quite understand that 😅

You're right 👍🏼 

> I would probably rather have them completely descriptive, as oftentimes short names lose it's meaning to different users. And when open source, that should be prefered in my opinion 🙂
>
> But I dont think it causes problems for now 🤔 We can change them later with a simple search-replace before releasing.

The same arguments were used for version `0.1.0` when we decided to have only one `CommonModel` :) Really, it makes no sense to have such names. If we have TS then we still have a hint of what type it is then we don't need to add the suffix `Model` because TS tells us.

> It is the usage type of a model, i.e. if you want to use a model as a type of a parameter, property, etc, you can simply define it as model.type.

So then it should be union type not as simple string. We can extend that union, but we should start with some simple type for that.

src/models/MetaModel.ts
```diff
@@ -0,0 +1,82 @@
...
+  constructor(
+    public name: string,
+    public originalInput: any) {
+  }
+}
+
+export class ReferenceModel extends MetaModel {
+  constructor(
+    public name: string,
+    public originalInput: any,
```
magicmatatjahu, line 11: You need to define modifiers for this same properties:

```suggestion
    name: string,
    originalInput: any,
```

src/models/ConstrainedMetaModel.ts
```diff
@@ -0,0 +1,93 @@
...
+    public type: string) {
+    super(name, originalInput);
+  }
+}
+
+export class ConstrainedReferenceModel extends ConstrainedMetaModel {
+  constructor(
+    public name: string,
+    public originalInput: any, 
+    public type: string,
```
magicmatatjahu, line 16: As I wrote in another comment, you don't need to redefine properties:

```suggestion
    name: string,
    originalInput: any, 
    type: string, 
```

magicmatatjahu (approved):
👌🏼  I added some suggestion. Decide what to do with them :)

src/models/MetaModel.ts
jonaslagoni: Ahh 👌 Changed!

magicmatatjahu (approved):

Issue #647, opened 2022-02-16
Render Go enum values


felixjung:
#### Reason/Context

Currently, Modeling seems to not render the values an enum type can take in Go.

For example

```yaml
properties:
  status:
    $id: Status
    type: string
    enum:
      - foo
      - bar
```

will only result in

```go
type Status string
```

While there are different approaches on how to deal with enums in Go, modelina could provide a simple standard approach, where the possible values are defined as constant strings. This would look as follows.

```go
type Status string

const (
  StatusFoo Status = "foo"
  StatusBar = "bar"
)
```

#### Description

- Update the Enum Renderer to render the values.
- This shouldn't be a breaking change as we are only generating additional code.

ibernabeudev:
Hi Felix, i just open a PR for this issue.

https://github.com/asyncapi/modelina/pull/662

PR #662, opened 2022-03-02
feat: add method to render as const enums values for go generator.

ibernabeudev:
**Description**

- Add a new func to render enum values as const for Go generator
- Works for integer (using iota) or strings

**Related issue(s)**
See also #647


jonaslagoni:
Thanks for taking the time to add this @ibernabeudev!

@smoya would you be able to take a look at the Go output? 🙂

examples/generate-go-enums/__snapshots__/index.spec.ts.snap
```diff
@@ -0,0 +1,38 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Should be able to render Go Enums and should log expected output to console 1`] = `
+Array [
+  "// Cities represents an enum of string.
+type Cities string
+
+const (
```
smoya, line 8: Nitpick: Would it make sense to add a comment here such as: 
```suggestion
// These are all the valid values for Cities enum.
const (
```

In that way, Godoc documentation will add an extra context, useful IMHO:

```bash 
❯ go doc -all main.go


CONSTANTS

const (
        London    Cities = "London"
        Rome             = "Rome"
        Brusseles        = "Brusseles"
)
    These are all the valid values for Cities enum.


TYPES

type Cities string
    Cities represents an enum of string.
```

examples/generate-go-enums/README.md
```diff
@@ -0,0 +1,17 @@
+# Go Data Models
+
+A basic example of how to Modelina render enums values in Go.
```
smoya, line 3:
```suggestion
A basic example of how Modelina renders enum values in Go.
```

smoya (commented):
Thanks for your work here @ibernabeudev ! 💯 🚀 

Go output code seems ok to me 👍  Just wrote a nitpick.
JS part review is pending by other folks.

ibernabeudev:
Thanks @smoya for your feedback I commit your suggestions. 

I just add a minor change to generate the const names with the type as first word (camel case) to avoid possible collisions when you have different name type enums but with an equal shared value, also I think this is how it was expected to works in the initial issue.

jonaslagoni (commented):
Do you mind adding the example to https://github.com/asyncapi/modelina/tree/master/examples ? 🙂 

You also have some linting errors 🙂 

Otherwise, it looks good to me 👍

jonaslagoni (changes requested):
You have some linting errors 🙂

ibernabeudev:
> You have some linting errors 🙂

Fixed, I rewrite the render method because I was having a  **Generic Object Injection Sink (security/detect-object-injection).** annoying linter error.

Also, I added notation **eslint-disable-next-line sonarjs/no-duplicate-string** on line 15 because I saw we used it in another file to avoid have to extract the literal "interface{}" (I think we want to be explicit for these cases).

jonaslagoni:
> Fixed, I rewrite the render method because I was having a **Generic Object Injection Sink (security/detect-object-injection).** annoying linter error.

Don't disable them 🙂 You can use `String(xxx)` to fix it.

jonaslagoni (approved):
LGTM 👍

jonaslagoni (changes requested):
I see the blackbox tests caught a problem 🙂

ibernabeudev:
> I see the blackbox tests caught a problem 🙂
Hi, sorry I was a bit busy these weeks. I fixed the blackbox test error. 

Also I saw a problem with the dockerfile for build the image that runs this test but you can easily solve adding two lines before copy the package-lock.json from hosts to container.

```
....
# Setup library 
WORKDIR /usr/app
COPY package.json .
COPY package-lock.json .
RUN npm install
COPY . .
```

In fact, you could move the last **COPY** command to the next line after **WORKDIR** under the comment **Setup library** and delete the others COPY lines saving two docker layers I think. 

I didn't add this fix to the MR because I don't know if this is correct and I suppose this must be place in another PR.

jonaslagoni (approved):

jonaslagoni:
@all-contributors please add @ibernabeudev for code, test, example, docs

PR #671, opened 2022-03-05
feat: shortened the way c# properties are generated

### LINKED ISSUE #670: Utilising shorthands for accessors in CSHARP_DEFAULT_CLASS_PRESET
#### Reason/Context
Currently, each property is generated like this
```
private Type field;

public Type Field
{
    get { return field; }
    set { field = value; }
}
```
which generates 6 times more code than it's necessary. C# allows for such a shorthand:
```
public Type Field { get; set; }
```
Way less code, the result the same

#### Description
I don't think it's a breaking change. Generated data models will be able to be used the same way they currently are.
The resolution for it would include modifying the `src/generators/csharp/renderers/ClassRenderer.ts` file and more specifically the implementation of CSHARP_DEFAULT_CLASS_PRESET. If anybody would ever want to use the more complex way of declaring properties, they would still be able to do it by implementing their own CsharpClassPreset, but I don't see a reason to make it default.

slowikowskiarkadiusz:
**Description**

- Incorporates using shorthand for property & accessors generation for C#
- Some formatting in awkward places

Code generation output before my changes
```
public class ExampleObject {
  private float? numberField;
  private dynamic[] collectionField;
  private Dictionary<string, dynamic> additionalProperties;

  public float? NumberField 
  {
    get { return numberField; }
    set { numberField = value; }
  }

  public dynamic[] CollectionField 
  {
    get { return collectionField; }
    set { collectionField = value; }
  }

  public Dictionary<string, dynamic> AdditionalProperties 
  {
    get { return additionalProperties; }
    set { additionalProperties = value; }
  }
}
```

Code generation output after my changes
```
public class ExampleObject {
  public float? NumberField { get; set; }
  public dynamic[] CollectionField { get; set; }
  public Dictionary<string, dynamic> AdditionalProperties { get; set; }
}
```

**Related issue(s)**

Resolved #670


jonaslagoni:
While I do think this is great to have, I don't think we should overwrite the default behavior, because it's only a matter of preference.

Instead, I suggest you add it to options such as `useShortHandAccessors` or something similar. And then based on that you either include either one of them as needed.

Also make sure to add examples, tests and documentation 🙂

jonaslagoni (changes requested):
While I do think this is great to have, I don't think we should overwrite the default behavior, because it's only a matter of preference.

Instead, I suggest you add it to options such as useShortHandAccessors or something similar. And then based on that you either include either one of them as needed.

slowikowskiarkadiusz:
Alright, so I've made a possibility to alter the default behaviour of CSHARP_COMMON_PRESET by setting an 'autoImplementedProperties' option to true, but I'm not sure whether I like it. It required me to pass options through a couple of functions and it basically completely changes the way the preset behaves. I wonder whether it would make more sense to just implement it as a separate preset instead. What do you think, @jonaslagoni?

test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
```diff
@@ -163,3 +163,69 @@ exports[`CSHARP_COMMON_PRESET should render GetHashCode support function 2`] = `
...
+  private string stringProp;
+  private float? numberProp;
+  private NestedTest objectProp;
+  private Dictionary<string, dynamic> additionalProperties;
+  private Dictionary<string, string> sTestPatternProperties;
+
+  public string StringProp 
+  {
+    get { return stringProp; }
+    set { stringProp = value; }
```
jonaslagoni, line 178: Should this not be short accessors? 🤔

test/generators/csharp/presets/CommonPreset.spec.ts
```diff
@@ -48,4 +48,18 @@ describe('CSHARP_COMMON_PRESET', () => {
...
+          preset: CSHARP_COMMON_PRESET,
+          options: {
+            autoImplementedProperties: true,
+          }
+        }
+      ]
+    });
+    const inputModel = await generator.generate(doc);
+    expect(inputModel[0].result).toMatchSnapshot();
+  });
```
jonaslagoni, line 64: Don't think we really need this one, as the example is plenty IMO 🙂

jonaslagoni (changes requested):
You have a few linting errors 🙂 

> Alright, so I've made a possibility to alter the default behaviour of CSHARP_COMMON_PRESET by setting the 'autoImplementedProperties' option to true, but I'm not sure whether I like it. It required me to pass options through a couple of functions and it basically completely changes the way the preset behaves. I wonder whether it would make more sense to just implement the whole thing as a separate preset instead. What do you think, @jonaslagoni?

Well, as presets are still in the early days, it's hard to say if they should be in one or the other. So up to you. I am happy with where they are at the moment 🙂

test/generators/csharp/presets/__snapshots__/CommonPreset.spec.ts.snap
slowikowskiarkadiusz: Should be, but now it's removed anyways :)

test/generators/csharp/presets/CommonPreset.spec.ts
slowikowskiarkadiusz: Done :)

jonaslagoni (changes requested):
@slowikowskiarkadiusz looks good to me 👍 You need to resolve some conflicts from master though, i.e. try merging the upstream master into your branch 🙂

jonaslagoni:
@ron-debajyoti anything you want to add?

ron-debajyoti:
I resolved the merge conflict, but looks like a test is failing. Mind taking a look ?

jonaslagoni:
@slowikowskiarkadiusz do you mind updating the test snapshots so we can merge this? 🙂

slowikowskiarkadiusz:
> @slowikowskiarkadiusz do you mind updating the test snapshots so we can merge this? 🙂

Sorry for the delay :) Done. All of the tests pass locally for me

slowikowskiarkadiusz:
> Looks like linter is having some problems as well 🙂

Should be in place now

jonaslagoni (approved):

jonaslagoni:
@ron-debajyoti as codeowner you have final review 🙂

ron-debajyoti (approved):

jonaslagoni:
@all-contributors please add @slowikowskiarkadiusz for code, test, example, doc

PR #676, opened 2022-03-08
refactor: add splitter 

jonaslagoni:
**Description**
This PR adds the splitter functionality that generators can use to split the MetaModel input based on what they support.


src/helpers/Splitter.ts
```diff
@@ -0,0 +1,82 @@
...
+  const shouldSplit = options.splitEnum === true && model instanceof EnumModel ||
+    options.splitUnion === true && model instanceof UnionModel ||
+    options.splitArray === true && model instanceof ArrayModel ||
+    options.splitTuple === true && model instanceof TupleModel ||
+    options.splitString === true && model instanceof StringModel ||
+    options.splitInteger === true && model instanceof IntegerModel ||
+    options.splitFloat === true && model instanceof FloatModel ||
+    options.splitBoolean === true && model instanceof BooleanModel ||
+    options.splitObject === true && model instanceof ObjectModel ||
+    options.splitDictionary === true && model instanceof DictionaryModel;
```
magicmatatjahu, line 34: You can create helper for that:

```ts
function splitGuard(model: MetaModel, condition: boolean, clazz: Object): boolean {
  return condition === true && model instanceof clazz;
}
```

and then in code:

```ts
const shouldSplit = splitGuard(model, options.splitEnum, EnumModel) || ...
```

magicmatatjahu (approved):
LGTM, but I added one suggestion.

src/helpers/Splitter.ts
jonaslagoni: True 😆 

I like the explicit style I think 🤔

jonaslagoni: We can always change it later if we find it weird.

PR #724, opened 2022-04-20
chore: improved formatting for C#

slowikowskiarkadiusz:
**Description**
In this PR I resolve most issues with formatting of the generated C# code, accordingly to conventions I've observed as a C# programmer. The changes include:
- Opening bracket at classes/enums definition were at the same line as the class name. Now it's going to be generated in a line below
- A newline between dependencies (usings) and class definition
- Provided indentation in appropriate places (i.e. class members were being generated with the same indent as the class declaration) 
- Each member of an enum is now in a separate line (this might seem a bit controversial, since basically neither of C# formatters is going to try to fix enum members that are all in a single line)
- Inserted/Removed newlines between particular parts of code (members of classes, enums definitions etc)

There might be also other issues than those that remain unresolved, but I think the generated code is going to look much more up to the standard.

No docs changes are required IMO.

**Related issue(s)**
Just a chore, I don't think it's worth making an issue


src/generators/csharp/CSharpGenerator.ts
```diff
@@ -52,25 +52,29 @@ export class CSharpGenerator extends AbstractGenerator<CSharpOptions, CSharpRend
     }
 
     const outputModel = await this.render(model, inputModel);
+
+    const outputDependencies = outputModel.dependencies.length === 0 ? '' : `${outputModel.dependencies.join('\n')}
+
+`;
```
jonaslagoni, line 58:
```suggestion
    const outputDependencies = outputModel.dependencies.length === 0 ? '' : `${outputModel.dependencies.join('\n')} \n \n`
```
Would this make it better for the eye? 🤔

jonaslagoni (changes requested):
This is great 🔥 

Had a few comments 🙂

src/generators/csharp/presets/CommonPreset.ts
```diff
@@ -64,16 +66,15 @@ function renderHashCode({ renderer, model }: {
...
+}`;
 }
 
 /**
  * Preset which adds `Equals`, `GetHashCode` functions to class. 
  * 
  * @implements {CSharpPreset}
  */
-export const CSHARP_COMMON_PRESET: CSharpPreset<CSharpCommonPresetOptions> = {
+export const CSHARP_COMMON_PRESET: CSharpPreset = {
```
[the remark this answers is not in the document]
slowikowskiarkadiusz: Oh yeah, that's absolutely right

src/generators/csharp/CSharpGenerator.ts
slowikowskiarkadiusz: I suppose 🙂

ron-debajyoti:
There are a few failing snapshots @slowikowskiarkadiusz. Would you mind updating them ?

slowikowskiarkadiusz:
> There are a few failing snapshots @slowikowskiarkadiusz. Would you mind updating them ?

Should be in place now :)

ron-debajyoti (approved):

jonaslagoni (approved):
LGTM 👍 

/rtm

jonaslagoni:
Gonna use `fix: ` instead as we are changing the formatting that is rendered, and it need a new release 🙂

PR #725, opened 2022-04-22
refactor: simplified constraints and type mapping

jonaslagoni:
**Description**
This PR finalizes the constraint and type mapping behavior.


src/helpers/ConstrainHelpers.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+  constrainedModel.type = getTypeFromMapping(typeMapping, {constrainedModel, renderer: context.renderer, propertyKey: context.propertyKey});
+  return constrainedModel;
+}
+
+export function constrainMetaModel<R extends AbstractRenderer>(typeMapping: TypeMapping<R>, constrainRules: Constraints, context: ConstrainContext<R, MetaModel>): ConstrainedMetaModel {
+  const constrainedName = constrainRules.modelName({modelName: context.metaModel.name});
+  const newContext = {...context, constrainedName};
+  if (newContext.metaModel instanceof ObjectModel) {
+    return constrainObjectModel(typeMapping, constrainRules, {...newContext, metaModel: newContext.metaModel});
```
magicmatatjahu, line 182: You can simplify it:

```ts
const newContext = {...context, constrainedName, metaModel: context.metaModel};

....

return constrainReferenceModel(typeMapping, constrainRules, newContext);
```

magicmatatjahu (approved):
Awesome, one suggestion, but you don't have to propagate it :)

src/helpers/ConstrainHelpers.ts
jonaslagoni: I tried, TS is throwing errors that the types for metaModel, when done like that, would be `MetaModel`, which does not match with `ReferenceModel` (for example).

Alternatively, I could do:
```
const newContext = {...context, constrainedName};
....
return constrainReferenceModel(typeMapping, constrainRules, newContext as any);
```

But that seems forced as well, and kinda shady 🤔 What you think?

magicmatatjahu: Ok, so please don't propagate my suggestion :) LGTM!

jonaslagoni: Alright 👍

magicmatatjahu (approved):

PR #728, opened 2022-04-25
fix: support net 6.0 for serialization and pass serialization options

### LINKED ISSUE #701: [C#] Generated JsonSerializer does not compile with .NET 6
#### Describe the bug
Code does not compile when generated using the C# code generation with `CSHARP_JSON_SERIALIZER_PRESET`

```
The call is ambiguous between the following methods or properties: 'JsonSerializer.Serialize<TValue>(TValue, JsonSerializerOptions?)' and 'JsonSerializer.Serialize<TValue>(TValue, JsonTypeInfo<TValue>)' [HAModels-DotNet]csharp(CS0121)
```

#### How to Reproduce
The issue is a known breaking change in .NET 6 further explained here: https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/6.0/jsonserializer-source-generator-overloads

#### Expected behavior
Generated code can compile. Now I'm manually adding `, (JsonSerializerOptions)null` to the following line: https://github.com/asyncapi/modelina/blob/51062c1d74e6bf0514619510e0e6f413d73bf4e3/src/generators/csharp/presets/JsonSerializerPreset.ts#L119

Polygens:
**Description**
I decided to **not** go for the solution proposed earlier in #701. I believe it makes more sense to pass the options block trough to the `Serialize` function. 

This fixes two bugs:
1. Removes the ambiguity introduced in .NET6.0 and will now compile. The function signature used in this fix is supported since the initial release of `JsonSerializer` in .NET Core 3.0 so it should be compatible with all versions.
2. Before during serialization it used the default options to serialize the fields. After this fix the options given during the call to `JsonSerializer.Serialize()` are passed trough. The new output is more inline of what would be expected.

For bug 2 let's take the example: https://github.com/u-packs/modelina-example/blob/example/Models/Auth/AuthMsg.cs
``` c#
AuthMsg msg = new() { Type = 5};

var options = new JsonSerializerOptions()
{
     NumberHandling = System.Text.Json.Serialization.JsonNumberHandling.WriteAsString,
};

JsonSerializer.SerializeToUtf8Bytes(msg, options);
```

Would result in `{"type": 5}` while using this fix it would correctly return `{"type": "5"}`

**Related issue(s)**
Fixes #701

### DIFF
--- a/examples/csharp-generate-serializer/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-generate-serializer/__snapshots__/index.spec.ts.snap
@@ -62,7 +62,7 @@ internal class RootConverter : JsonConverter<Root>
   {
     if (value == null)
     {
-      JsonSerializer.Serialize(writer, null);
+      JsonSerializer.Serialize(writer, null, options);
       return;
     }
     var properties = value.GetType().GetProperties();
@@ -72,7 +72,7 @@ internal class RootConverter : JsonConverter<Root>
     if(value.Email != null) { 
       // write property name and let the serializer serialize the value itself
       writer.WritePropertyName(\\"email\\");
-      JsonSerializer.Serialize(writer, value.Email);
+      JsonSerializer.Serialize(writer, value.Email, options);
     }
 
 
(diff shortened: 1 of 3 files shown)


jonaslagoni (approved):
That sounds like a great fix! 🔥

jonaslagoni:
@all-contributors please add @Polygens for code, test, bug

PR #737, opened 2022-04-27
fix: exclude _ from Java key enum rendering

### LINKED ISSUE #687: Exclude _ from Java key enum rendering
#### Reason/Context
Given an enum key such as `something_something` the enum renderer creates the following key `SOMETHING_UNDERSCORE_SOMETHING` even when Java fully supports `SOMETHING_SOMETHING`

#### Description
The same issue was already fixed for TS with this ticket https://github.com/asyncapi/modelina/issues/609. 
Maybe it's possible to fix it in a more generic way, so that it does not have to be fixed for each language generator individually, as I would guess that `_` is supported by most/all languages as part of an enum key.

Samridhi-98:
**Description**

-  Modified `java/EnumRenderer.ts`
-  Added `java/renderer/EnumRenderer.spec.ts`

**Related issue(s)**

Fixes #687

### DIFF
--- a/src/generators/java/renderers/EnumRenderer.ts
+++ b/src/generators/java/renderers/EnumRenderer.ts
@@ -1,5 +1,5 @@
 import { JavaRenderer } from '../JavaRenderer';
-import { EnumPreset} from '../../../models';
+import { EnumPreset } from '../../../models';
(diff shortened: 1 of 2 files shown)


jonaslagoni (approved):
Awesome 👍

PR #746, opened 2022-05-07
fix(c#): generated enum extension methods lead to invalid references

### LINKED ISSUE #700: [C#] Generated enum extension methods lead to invalid references
#### Describe the bug
Invalid enum code is generated when using the C# code generation with `CSHARP_JSON_SERIALIZER_PRESET`
https://github.com/asyncapi/modelina/blob/ef4c0b739f31893cd1320bc1e55c4f9dbc746eea/src/generators/csharp/renderers/EnumRenderer.ts#L20 seems to be conflicting with 
https://github.com/asyncapi/modelina/blob/51062c1d74e6bf0514619510e0e6f413d73bf4e3/src/generators/csharp/presets/JsonSerializerPreset.ts#L142

There is a discrepancy between how the enum class name and it's external references are generated. Mostly just plural vs singular but I wonder if `formattedName` vs `type` might also lead to issues in the future :thinking: 

#### How to Reproduce
Created a test repo to reproduce: https://github.com/u-packs/modelina-example
When running `npm i && npm start` the invalid code is generated in the ./models folder

Same issue is shown in the test files: https://github.com/asyncapi/modelina/blob/c94c67d2320a02b4bb472b37cd87bd2bbd07c649/test/generators/csharp/presets/__snapshots__/JsonSerializerPreset.spec.ts.snap

#### Expected behavior
Both the enum class name and references to the enum class should be consistent.

Polygens:
**Description**
As mentioned in #700 the generated serializer code using enums did not match with the enum generated code.
This led to invalid code that required manual fixing. This tiny PR aligns both generators.

**Related issue(s)**

Fixes #700

### DIFF
--- a/src/generators/csharp/presets/JsonSerializerPreset.ts
+++ b/src/generators/csharp/presets/JsonSerializerPreset.ts
@@ -139,7 +139,7 @@ function renderDeserializeProperty(type: string, model: CommonModel, inputModel:
     const propertyModelKind = TypeHelpers.extractKind(resolvedModel);
     //Referenced enums is the only one who need custom serialization
     if (propertyModelKind === ModelKind.ENUM) {
-      return `${type}Extension.To${type}(JsonSerializer.Deserialize<dynamic>(ref reader, options))`;
+      return `${type}Extensions.To${type}(JsonSerializer.Deserialize<dynamic>(ref reader, options))`;
     }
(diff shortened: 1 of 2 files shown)


jonaslagoni:
@ron-debajyoti need your review as well 🙂

PR #752, opened 2022-05-13
fix: #749 Added nullable checking.

### LINKED ISSUE #749: Java Template: json schema union is generated as object
#### Describe the bug
In our schema, we have multiple types, that are nullable types, like a string.
Eg.
```
          username:
            type: [ "string", "null" ]
            description: The username of the user that sent this message
```
What we expected would happen:
  private String username;

What we got
  private Object username;

#### How to Reproduce
We generate our file like this:

```java
const generator = new JavaFileGenerator({
  collectionType: 'List',
  presets: [
    {
      preset: JAVA_COMMON_PRESET,
      options: {
        equals: true,
        hashCode: true,
        classToString: true,
      },
    },
    JAVA_CONSTRAINTS_PRESET,
    JAVA_JACKSON_PRESET,
    JAVA_DESCRIPTION_PRESET,
  ],
});

(async () => {
  const asyncDoc = readFileSync(schemaFile, 'utf8');
    const parsedDoc = await parser.parse(asyncDoc);
  await generator.generateToFiles(parsedDoc, outDirectory, {
    packageName: rootPackage,
  });


  //Temp fix:
  /*
   * We replace the union of nulls, with just the object, because currently the code generation turns it into objects, instead of string or boolean etc.
   */
  //const parsedDoc = await parser.parse(
  //  asyncDoc.replace(/\[(?: ?null" ?,)? ?("\w+")(?: ?, ?"null")? ?]/g, '$1')
  //);
  //await generator.generateToFiles(parsedDoc, outDirectory, {
  //  packageName: rootPackage,
  //});
})();
```

#### Expected behavior
Making a string nullable, should leave it as a java String, instead of turning it into an object.

RobinGoussey:
**Description**
Nullables, as Type: ['string','null'] should now return string instead of Object[].

**Related issue(s)**
Fixes #749

### DIFF
--- a/src/generators/java/JavaRenderer.ts
+++ b/src/generators/java/JavaRenderer.ts
@@ -1,12 +1,12 @@
-import { AbstractRenderer } from '../AbstractRenderer';
-import { JavaGenerator, JavaOptions } from './JavaGenerator';
-import { CommonModel, CommonInputModel, Preset } from '../../models';
-import { FormatHelpers, ModelKind, TypeHelpers } from '../../helpers';
-import { isReservedJavaKeyword } from './Constants';
+import {AbstractRenderer} from '../AbstractRenderer';
+import {JavaGenerator, JavaOptions} from './JavaGenerator';
+import {CommonInputModel, CommonModel, Preset} from '../../models';
+import {FormatHelpers, ModelKind, TypeHelpers} from '../../helpers';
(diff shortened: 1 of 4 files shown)


jonaslagoni (changes requested):
Seems like there is an issue with some references in the blackbox tests that make it uncompilable 🤔

PR #755, opened 2022-05-18
feat: add nullable object in ts generator

### LINKED ISSUE #754: [TS] Nullable objects lose their object type
#### Describe the bug
When an object is nullable (i.e. having a `oneOf` with a null item and an object item), the TypeScript generator will _lose_ the object type by outputting `type MyObject = null | object`.

#### How to Reproduce
Considering the following AsyncAPI spec:

```yaml
asyncapi: 2.4.0
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
          address:
            oneOf:
              - type: 'null'
              - $ref: '#/components/schemas/Address'
  schemas:
    Address:
      type: object
      properties:
        line1:
          type: string
        line2:
          type: string
        postCode:
          type: string
        city:
          type: string
````

The previous snippet is the example spec AsyncAPI provides, with a `Address` schema added. The `UserSignedUp` payload has a `address` field pointing to the `Address` schema. One can put either a valid `Address` object or a `null` value.

Here is the generated code:

```ts
  displayName?: string;
  email?: string;
  address?: AnonymousSchema_4;
}
export type AnonymousSchema_4 = null | object;
```

As you can see, the `Address` schema is _lost_ and the union type has _simplified_ the type to `object`.

#### Expected behavior

The generator should have kept the `Address` type and used it in the `address` field's union type:

```ts
interface AnonymousSchema_1 {
  displayName?: string;
  email?: string;
  address?: AnonymousSchema_4;
}
type AnonymousSchema_4 = null | Address;
interface Address {
  line1?: string;
  line2?: string;
  postCode?: string;
  city?: string;
}
```

IamBlueSlime:
**Description**

When a model with a composed type of `['object', 'null']`, the TS generator will generate an _Inner_ type for the object and create a nullable type alias.

Example (from the tests):

The following model,

```js
{
  $id: 'NullableObject',
  type: ['object', 'null'],
  properties: {
    foo: {
      type: 'string'
    },
  },
  additionalProperties: false,
}
```

Will generate the following code:

```ts
interface NullableObjectInner {
  foo?: string;
}
type NullableObject = NullableObjectInner | null;
```

or

```ts
class NullableObjectInner {
  private _foo?: string;

  constructor(input: {
    foo?: string,
  }) {
    this._foo = input.foo;
  }

  get foo(): string | undefined { return this._foo; }
  set foo(foo: string | undefined) { this._foo = foo; }
}
type NullableObject = NullableObjectInner | null;
```

**Related issue(s)**

Fixes #754

### DIFF
--- a/src/generators/typescript/TypeScriptGenerator.ts
+++ b/src/generators/typescript/TypeScriptGenerator.ts
@@ -167,13 +167,14 @@ ${modelCode}`;
     modelObject.type = 'object';
     model.$ref = modelObject.$id;
 
-    const [modelOutput, typeOutput] = await Promise.all([
+    const [modelObjectOutput, modelOutput] = await Promise.all([
       this.renderModelType(modelObject, inputModel),
       this.renderType(model, inputModel),
     ]);
     return RenderOutput.toRenderOutput({
       ...modelOutput,
-      result: `${modelOutput.result}\n${typeOutput.result}`,
+      result: `${modelObjectOutput.result}\n${modelOutput.result}`,
+      dependencies: [...modelObjectOutput.dependencies, ...modelOutput.dependencies],
     });
   }
 
--- a/src/generators/typescript/TypeScriptGenerator.ts
+++ b/src/generators/typescript/TypeScriptGenerator.ts
@@ -119,6 +119,12 @@ ${modelCode}`;
       }
       return this.renderEnum(model, inputModel);
     }
+    case ModelKind.UNION: {
+      if (TypeHelpers.isNullableObject(model)) {
+        return this.renderNullableObject(model, inputModel);
+      }
+      return this.renderType(model, inputModel);
+    }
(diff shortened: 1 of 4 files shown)


jonaslagoni (commented):
I don't see anything wrong with the implementation. My only concern is that the implementation focuses on the `type: ['object', 'null']` input, and it can quickly become a big mess considering https://github.com/asyncapi/modelina/pull/752.

As `next` is "soon" going to take over, I dont mind merging this, as it solves the specific use-case.

@magicmatatjahu @Samridhi-98 what are your thoughts?

PR #764, opened 2022-06-07
docs: add example to use custom enum value name

jonaslagoni:
**Description**
This example is to showcase and answer https://stackoverflow.com/questions/72530361/asyncapi-enum-with-specific-values

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ This package is still under development and has not reached version 1.0.0 yet. T
 
 <!-- toc -->
 
-- [:loudspeaker: ATTENTION:](#loudspeaker-attention)
 - [Requirements](#requirements)
 - [Installation](#installation)
 - [Features](#features)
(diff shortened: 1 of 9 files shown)


magicmatatjahu (dismissed):
LGTM! However, we I think that we should change `chore` to the `docs`, even if you added new example.

jonaslagoni:
> LGTM! However, we I think that we should change `chore` to the `docs`, even if you added new example.

True, changing :)

magicmatatjahu:
@jonaslagoni Yeah, I see that also on other PRs, but as I see they try to fix it https://github.com/actions/setup-node/issues/516

jonaslagoni:
@magicmatatjahu lets just wait for the fix then, no rush and I dont want to start changing the CI back and furth.

magicmatatjahu:
@jonaslagoni You can merge it :)

magicmatatjahu (approved):

PR #765, opened 2022-06-09
chore: refactored TypeScript and generator implementation

jonaslagoni:
**This PR breaks compilation and jobs will fail which is expected until all the generators have been refactored. So failing jobs are from now on to be ignored.**

**Description**
This PR refactors the TypeScript and abstract generator/renderer to support the new core model.

This PR will be used as a guideline for the next generator rewrites.

**Related issue(s)**
Blocked by #

### DIFF
--- a/src/generators/AbstractGenerator.ts
+++ b/src/generators/AbstractGenerator.ts
@@ -1,4 +1,4 @@
-import { CommonInputModel, CommonModel, OutputModel, Preset, Presets, RenderOutput, ProcessorOptions } from '../models';
+import { InputMetaModel, OutputModel, Preset, Presets, RenderOutput, ProcessorOptions, MetaModel, ConstrainedMetaModel } from '../models';
 import { InputProcessor } from '../processors';
 import { IndentationTypes } from '../helpers';
 import { isPresetWithOptions } from '../utils';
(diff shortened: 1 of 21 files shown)


src/generators/AbstractGenerator.ts
```diff
@@ -9,35 +9,34 @@ export interface CommonGeneratorOptions<P extends Preset = Preset> {
...
-  public abstract renderCompleteModel(model: CommonModel, inputModel: CommonInputModel, options: RenderCompleteModelOptions): Promise<RenderOutput>;
+  public abstract render(model: MetaModel, inputModel: InputMetaModel): Promise<RenderOutput>;
+  public abstract renderCompleteModel(model: MetaModel, inputModel: InputMetaModel, options: RenderCompleteModelOptions): Promise<RenderOutput>;
+  public abstract constrainToMetaModel(model: MetaModel): ConstrainedMetaModel;
+  public abstract splitMetaModel(model: MetaModel): MetaModel[];
 
-  public process(input: Record<string, unknown>): Promise<CommonInputModel> {
-    return InputProcessor.processor.process(input, this.options.processorOptions);
+  public async process(input: Record<string, unknown>): Promise<InputMetaModel> {
+    return await InputProcessor.processor.process(input, this.options.processorOptions);
```
magicmatatjahu, line 39: You don't need to wrap async with await, if return type of simple function (like above) is promise.

magicmatatjahu (approved):
I have nothing to complain about, just one comment :) Good work! 👏🏼

src/generators/AbstractGenerator.ts
jonaslagoni: True, removed 👍

PR #766, opened 2022-06-09
chore: refactored model setup to support generators

jonaslagoni:
**This PR breaks compilation and jobs will fail which is expected until all the generators have been refactored. So failing jobs are from now on to be ignored.**

**Description**
This PR introduces a couple of changes:
1. Constrainers no longer have access to the renderer, as they have no reason to and it overcomplicates the entire setup with recursive dependencies. The only reason it was added, was because they needed access to the underlying options, that are now passed along instead.
2. Core models have gotten a new function called `getNearestDependencies` same as the last core model which is to be used to find the nearest dependencies. I am not 100% sure about the implementation and still need to add tests, but with a broken source code, this will be changed later.
3. Introduces `InputMetaModel` to replace `CommonInputModel`.

### DIFF
--- a/src/helpers/TypeHelpers.ts
+++ b/src/helpers/TypeHelpers.ts
@@ -1,5 +1,3 @@
-/* eslint-disable no-unused-vars */
-import { AbstractRenderer } from '../generators';
 import { CommonModel } from '../models';
 import { ConstrainedAnyModel, ConstrainedBooleanModel, ConstrainedFloatModel, ConstrainedIntegerModel, ConstrainedMetaModel, ConstrainedObjectModel, ConstrainedReferenceModel, ConstrainedStringModel, ConstrainedTupleModel, ConstrainedArrayModel, ConstrainedUnionModel, ConstrainedEnumModel, ConstrainedDictionaryModel } from '../models/ConstrainedMetaModel';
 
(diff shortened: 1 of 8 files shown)


src/helpers/TypeHelpers.ts
```diff
@@ -26,30 +24,30 @@ export class TypeHelpers {
...
-  Float: TypeMappingFunction<ConstrainedFloatModel, R>,
-  Integer: TypeMappingFunction<ConstrainedIntegerModel, R>,
-  String: TypeMappingFunction<ConstrainedStringModel, R>,
-  Boolean: TypeMappingFunction<ConstrainedBooleanModel, R>,
-  Tuple: TypeMappingFunction<ConstrainedTupleModel, R>,
-  Array: TypeMappingFunction<ConstrainedArrayModel, R>,
-  Enum: TypeMappingFunction<ConstrainedEnumModel, R>,
-  Union: TypeMappingFunction<ConstrainedUnionModel, R>,
-  Dictionary: TypeMappingFunction<ConstrainedDictionaryModel, R>
+export type TypeMapping<Options> = {
```
magicmatatjahu, line 35: In `ConstrainContext` we have `Options` as first generic type, maybe we should also standarize it, and change here order?

```ts
export type TypeContext<Options, T extends ConstrainedMetaModel>
export type TypeMappingFunction<Options, T extends ConstrainedMetaModel>
```

? I'm just thinking out loud 😄

src/helpers/TypeHelpers.ts
```diff
@@ -26,30 +24,30 @@ export class TypeHelpers {
...
+  Boolean: TypeMappingFunction<ConstrainedBooleanModel, Options>,
+  Tuple: TypeMappingFunction<ConstrainedTupleModel, Options>,
+  Array: TypeMappingFunction<ConstrainedArrayModel, Options>,
+  Enum: TypeMappingFunction<ConstrainedEnumModel, Options>,
+  Union: TypeMappingFunction<ConstrainedUnionModel, Options>,
+  Dictionary: TypeMappingFunction<ConstrainedDictionaryModel, Options>
 };
 
-export function getTypeFromMapping<T extends ConstrainedMetaModel, R extends AbstractRenderer>(typeMapping: TypeMapping<R>, context: TypeContext<T, R>): string {
+export function getTypeFromMapping<T extends ConstrainedMetaModel, Options>(typeMapping: TypeMapping<Options>, context: TypeContext<T, Options>): string {
```
magicmatatjahu, line 50: Also here.

src/models/ConstrainedMetaModel.ts
```diff
@@ -1,12 +1,18 @@
 import { MetaModel } from './MetaModel';
 
-export class ConstrainedMetaModel extends MetaModel {
+export abstract class ConstrainedMetaModel extends MetaModel {
```
magicmatatjahu, line 3: 👌🏼

magicmatatjahu (changes requested):
Awesome! However, I added some comments, please refer to them :)

src/helpers/TypeHelpers.ts
jonaslagoni: I mean, aren't they in the same order currently? Just switched from what you suggested, of course `TypeMapping<Options>` is not, but I think that's okay.

magicmatatjahu: Yeah, but I refer to the `ConstrainContext` where `Options` is first argument, rest types have `Options` as last argument :P

src/models/ConstrainedMetaModel.ts
```diff
@@ -49,12 +81,33 @@ export class ConstrainedObjectModel extends ConstrainedMetaModel {
...
+        return modelProperty.property instanceof ConstrainedReferenceModel;
+      }
+    ).map((modelProperty) => {
+      return modelProperty.property as ConstrainedReferenceModel;
+    });
+    //Ensure no self references
+    dependencyModels = dependencyModels.filter((referenceModel) => {
+      return referenceModel.name !== this.name;
+    });
+    return dependencyModels;
```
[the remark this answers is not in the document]
magicmatatjahu: hard to debug, why? 😄 C'mon :P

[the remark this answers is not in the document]
jonaslagoni: Cause you cannot explicitly see the intermediary results, right?

src/models/Preset.ts
```diff
@@ -22,31 +22,29 @@ export enum PropertyType {
...
-  ctor?: (args: PresetArgs<R, O>) => Promise<string> | string;
-  property?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
-  getter?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
-  setter?: (args: PresetArgs<R, O> & PropertyArgs) => Promise<string> | string;
+export interface ClassPreset<R extends AbstractRenderer, O> extends CommonPreset<R, O, ConstrainedObjectModel> {
+  ctor?: (args: PresetArgs<R, O, ConstrainedObjectModel>) => Promise<string> | string;
+  property?: (args: PresetArgs<R, O, ConstrainedObjectModel> & PropertyArgs) => Promise<string> | string;
+  getter?: (args: PresetArgs<R, O, ConstrainedObjectModel> & PropertyArgs) => Promise<string> | string;
+  setter?: (args: PresetArgs<R, O, ConstrainedObjectModel> & PropertyArgs) => Promise<string> | string;
 }
```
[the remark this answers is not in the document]
jonaslagoni: I dont follow. The third generic argument is for defining it's a hook for the model `ConstrainedObjectModel` that is being rendered 🤔

src/models/ConstrainedMetaModel.ts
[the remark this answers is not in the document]
magicmatatjahu: I agree with that, but these functions are so easy that I don't see the point that breaking them down into several steps would give us more benefit, but as you prefer, I'm not forcing you to do anything.

src/models/Preset.ts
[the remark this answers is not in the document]
magicmatatjahu: 😅 Hah, yes, you're right. So many similar names, I didn't notice this `...ObjectModel`. It's ok, forgive the comment.

src/helpers/TypeHelpers.ts
jonaslagoni: I see what you mean. Hmm, but would that not also be the case the other way around if you had a function that took a generic model type but a specific option? I.e. flipped case then here.

So does not really matter about changing it as it would be like potato potato 😆

magicmatatjahu: Yeah, right. Model has much "power" in this case, so it can stay as it is.

magicmatatjahu (approved):

PR #772, opened 2022-06-16
docs: add inheritance example

jonaslagoni:
**Description**
This PR adds an inheritance example, where we overwrite the entire rendering process.

**Related issue(s)**
Related to https://stackoverflow.com/questions/72638195/modelina-csharp-generator-add-inheritance


magicmatatjahu (approved):
I know we don't support it at the moment but it's worth remembering to correct such an example when we'll support inheritance as first-citizen.

jonaslagoni:
> I know we don't support it at the moment but it's worth remembering to correct such an example when we'll support inheritance as first-citizen.

Absolutely 👍

PR #788, opened 2022-07-03
fix: make sure Java switch and assert words are treated as keywords

### LINKED ISSUE #786: Java Generator: "switch" and "assert" keyword are not considered reserved Java keywords
#### Describe the bug
By inspecting the `Constants.ts` file from Java Generator we can see, that we have `'switch  assert',` as an entry in the keywords array. While we should have `'switch', 'assert'` as separate entries.

#### How to Reproduce
To reproduce the issue we should have a schema that looks like this:
```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "title": "Test",
  "type": "object",
  "properties": {
    "switch": {
      "type": "string"
    },
    "assert": {
      "type": "string"
    }
  } 
}
```
Once this schema is run through Java generator we get back a class declaration like this:
```java
public class Test {
  private String switch;
  private String assert;

  public String getSwitch() { return this.switch; }
  public void setSwitch(String switch) { this.switch = switch; }
  public String getAssert() { return this.assert; }
  public void setAssert(String assert) { this.assert = assert; }
}
```
Which won't compile.

#### Expected behavior
If we have a schema that has `switch` or `assert` as a schema name or property it should be rendered as `reservedSwitch` and `reservedAssert` since `switch` and `assert` are Java keywords.

artur-ciocanu:
**Description**
In the Java generator folder we have a file named `Constants.ts`. This should contain all the Java reserved keywords. In the file we have an entry `'switch  assert'`. Apparently it was a typo and the file should have `'switch'` and `'assert'` as separate entries.

**Related issue(s)**
Fixes #786


jonaslagoni (approved):
LGTM 👍

jonaslagoni:
@all-contributors please add @artur-ciocanu for code, bug

jonaslagoni:
> @jonaslagoni as far as I saw the same issue is present in the `next` branch, will you take care of merging this fix or I need to open a PR against `next`?

I will merged it 👍

PR #791, opened 2022-07-04
refactor: switch interpretation of pattern properties

jonaslagoni:
**Description**
This PR changes the functionality for how patternProperties are interpreted, by merging the structures into `additionalProperties` to only have a single property that contains all "extra" properties. 

This change is because Modelina does not have runtime constraints (and might never have, or at least far out in the future), which means you cannot really have multiple dictionaries with unwrapping serialization options. This makes sure JSON Schema never creates that scenario.


magicmatatjahu (approved):
LGTM!

PR #799, opened 2022-07-07
chore: give constrained properties access to the raw property

jonaslagoni:
**Description**
This PR gives constrained property models access to the raw properties. The reason for adding this is because some presets want to convert the models to JSON, there they need access to the raw, unconstrained property name


src/models/ConstrainedMetaModel.ts
```diff
@@ -159,16 +152,33 @@ export class ConstrainedDictionaryModel extends ConstrainedMetaModel {
...
+        return model instanceof ConstrainedReferenceModel;
+      }
+    ).map((model) => {
+      return model as ConstrainedReferenceModel;
+    });
+    //Ensure no self references
+    dependencyModels = dependencyModels.filter((referenceModel) => {
+      return referenceModel.name !== this.name;
+    });
+    return dependencyModels;
```
magicmatatjahu, line 169:
```suggestion
    //Ensure no self references
    return dependencyModels.filter((referenceModel) => {
      return referenceModel.name !== this.name;
    });
```

magicmatatjahu (approved):
One thing that we can improve, but it's not mandatory.

PR #813, opened 2022-07-14
fix: generate Jackson annotations for Java enums only with Jackson Preset

### LINKED ISSUE #812: Java generator always adds @Json annotations for enums
@micro-jumbo commented on [Thu Jul 14 2022](https://github.com/asyncapi/spec/issues/822)

### Describe the bug
Java generator always adds @Json annotations for enums

### To Reproduce
For document:
```json
{
  "asyncapi": "2.3.0",
  "id": "urn:my:service",
  "defaultContentType": "application/json",
  "info": {
    "title": "My services",
    "version": "1.0.0"
  },
  "channels": {
    "accepted": {
      "subscribe": {
        "message": {
          "name": "SuccessMessage",
          "title": "SuccessMessage",
          "payload": {
            "type": "object",
            "title": "SuccessMessage",
            "properties": {
              "result": {
                "title": "SuccessMessageResult",
                "type": "string",
                "const": "SUCCESS"
              },
              "message": {
                "type": "string",
                "title": "SuccessMessageMessage"
              }
            },
            "required": [
              "result"
            ],
            "additionalProperties": false
          }
        }
      }
    }
  }
}
```

and generator:
```javascript
const generator = new JavaGenerator({
  collectionType: "List",
});
```

generated `SuccessMessageMessage` `enum` contains Jackson annotations (`@JsonValue` and `@JsonCreator`), but it shouldn't:
```java
import com.fasterxml.jackson.annotation.*;

public enum SuccessMessageResult {
  SUCCESS("SUCCESS");

  private String value;

  SuccessMessageResult(String value) {
    this.value = value;
  }
    
  @JsonValue
  public String getValue() {
    return value;
  }

  @Override
  public String toString() {
    return String.valueOf(value);
  }

  @JsonCreator
  public static SuccessMessageResult fromValue(String value) {
    for (SuccessMessageResult e : SuccessMessageResult.values()) {
      if (e.value.equals(value)) {
        return e;
      }
    }
    throw new IllegalArgumentException("Unexpected value '" + value + "'");
  }
}
```

These should be generated only when Jackson annotations are enabled.

---

@github-actions[bot] commented on [Thu Jul 14 2022](https://github.com/asyncapi/spec/issues/822#issuecomment-1184728142)

Welcome to AsyncAPI. Thanks a lot for reporting your first issue. Please check out our [contributors guide](https://github.com/asyncapi/community/blob/master/CONTRIBUTING.md) and the instructions about a [basic recommended setup](https://github.com/asyncapi/.github/blob/master/git-workflow.md) useful for opening a pull request.<br />Keep in mind there are also other channels you can use to interact with AsyncAPI community. For more details check out [this issue](https://github.com/asyncapi/asyncapi/issues/115).

micro-jumbo:
**Description**

- generate Jackson annotations for Java enums only with Jackson Preset
- update all relevant tests

**Related issue(s)**
Fixes #812

### DIFF
--- a/src/generators/java/presets/JacksonPreset.ts
+++ b/src/generators/java/presets/JacksonPreset.ts
@@ -19,5 +19,43 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
       }
       return renderer.renderBlock([content]);
     },
+  },
+  enum: {
+    self({renderer, content}) {
+      renderer.addDependency('import com.fasterxml.jackson.annotation.*;');
+      return content;
+    },
+    additionalContent({ renderer, model }) {
+      const enumName = renderer.nameType(model.$id);
+      const type = Array.isArray(model.type) ? 'Object' : model.type;
+      const classType = renderer.toClassType(renderer.toJavaType(type, model));
+
+      return `private ${classType} value;
+    
+${enumName}(${classType} value) {
+  this.value = value;
+}
+    
+@JsonValue
+public ${classType} getValue() {
+  return value;
+}
+
+@Override
+public String toString() {
+  return String.valueOf(value);
+}
+
+@JsonCreator
+public static ${enumName} fromValue(${classType} value) {
+  for (${enumName} e : ${enumName}.values()) {
+    if (e.value.equals(value)) {
+      return e;
+    }
+  }
+  throw new IllegalArgumentException("Unexpected value '" + value + "'");
+}`;
+    },
+
   }
 };
--- a/src/generators/java/presets/JacksonPreset.ts
+++ b/src/generators/java/presets/JacksonPreset.ts
@@ -3,7 +3,7 @@ import { JavaPreset } from '../JavaPreset';
 
 /**
  * Preset which adds `com.fasterxml.jackson` related annotations to class's property getters.
- * 
+ *
  * @implements {JavaPreset}
  */
 export const JAVA_JACKSON_PRESET: JavaPreset = {
@@ -12,7 +12,7 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
       renderer.addDependency('import com.fasterxml.jackson.annotation.*;');
       return content;
     },
-    getter({ renderer, propertyName, content, type }) {
+    getter({renderer, propertyName, content, type}) {
       if (type === PropertyType.property) {
         const annotation = renderer.renderAnnotation('JsonProperty', `"${propertyName}"`);
         return renderer.renderBlock([annotation, content]);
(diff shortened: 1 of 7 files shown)


src/generators/java/presets/JacksonPreset.ts
```diff
@@ -19,5 +19,43 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
...
+@JsonCreator
+public static ${enumName} fromValue(${classType} value) {
+  for (${enumName} e : ${enumName}.values()) {
+    if (e.value.equals(value)) {
+      return e;
+    }
+  }
+  throw new IllegalArgumentException("Unexpected value '" + value + "'");
+}`;
+    },
```
jonaslagoni, line 58: I see we are kinda missing two preset hooks, one that renders the `fromValue` and `getValue` methods. 

Because this way you can do something like:

```ts
    getValue({content}) {
      return `@JsonValue
${content}`
    },
    fromValue({content}) {
      return `@JsonValue
${content}`;
    },
```

Cause the problem with the current implementation is that it completely overwrites whatever the enum renderer generates which should rarely be the case for presets.

Do you want to take a crack at adding those preset hooks @micro-jumbo ?

jonaslagoni (changes requested):

src/generators/java/presets/JacksonPreset.ts
micro-jumbo: @jonaslagoni added new presets: `getValue`, `fromValue`. Additionally added `ctor` preset as having constructor in `additionalContent` looked a bit ackward

jonaslagoni: Awesome 👍

src/models/Preset.ts
```diff
@@ -44,6 +44,9 @@ export interface EnumArgs {
 
 export interface EnumPreset<R extends AbstractRenderer, O extends object = any> extends CommonPreset<R, O> {
   item?: (args: PresetArgs<R, O> & EnumArgs) => string;
+  ctor?: (args: PresetArgs<R, O> & EnumArgs) => string;
+  getValue?: (args: PresetArgs<R, O> & EnumArgs) => string;
+  fromValue?: (args: PresetArgs<R, O> & EnumArgs) => string;
```
jonaslagoni, line 49: As these presets are used for all generators, we need to move this addition to the [JavaPreset](https://github.com/asyncapi/modelina/blob/master/src/generators/java/JavaPreset.ts) file. I think you should be able to use the following:

```ts
export interface JavaEnumPreset<O> extends EnumPreset<EnumRenderer, O> {
  ctor?: (args: PresetArgs<EnumRenderer, O> & EnumArgs) => string;
  getValue?: (args: PresetArgs<EnumRenderer, O> & EnumArgs) => string;
  fromValue?: (args: PresetArgs<EnumRenderer, O> & EnumArgs) => string;
}

export type JavaPreset<O extends object = any> = Preset<{
  class: ClassPreset<ClassRenderer, O>;
  enum: JavaEnumPreset<O>;
}>;
```

jonaslagoni (changes requested):
I only have one small addition, otherwise, it looks good from my side 🔥 

@magicmatatjahu anything from your side?

jonaslagoni (approved):
This looks good to me 👍

jonaslagoni:
@all-contributors please add @micro-jumbo for code, test, bug

PR #814, opened 2022-07-15
fix: oneof implementation

### LINKED ISSUE #367: Typescript: Using oneOf leads to creation and use of an anonymous schema rather than use of a union type
#### Describe the bug
Typescript: Using oneOf leads to creation and use of an anonymous schema rather than use of a union type.

#### How to Reproduce

api.yaml
```
asyncapi: 2.1.0
info:
  title: oneOf bug
  version: "0.1.0"
channels: {}
components:
  messages:
    a:
      payload:
        $ref: "#/components/schemas/a"

  schemas:
    a:
      type: object
      properties:
        prop:
          oneOf:
            - $ref: "#/components/schemas/b"
            - $ref: "#/components/schemas/c"

    b:
      type: object
      properties:
        someProp:
          type: integer
          format: int32

    c:
      type: object
      properties:
        anotherProp:
          type: integer
          format: int32

```
Example code to consume the api:
```
import { TypeScriptGenerator } from "@asyncapi/modelina";
import fs from "fs";
import parser from "@asyncapi/parser";

const generator = new TypeScriptGenerator({ modelType: "interface" });
const contents = fs.readFileSync("api.yaml", "utf8");
const doc = await parser.parse(contents);

async function generate() {
  const models = await generator.generate(doc);
  models.forEach((model) => console.log(model.result));
}

generate();
```
Generated outcome:
```
export interface A {
  prop?: AnonymousSchema_1;
}
  someProp?: number;
  anotherProp?: number;
}
```

#### Expected behavior
According to my understanding, this should be generated instead:
```
export interface A {
  prop?: B | C
}
export interface B {
  someProp?: number;
}
export interface C {
  anotherProp?: number;
}
```

jacopomaroli:
**Description**
Currently oneOf and anyOf are treated equally https://github.com/asyncapi/modelina/blob/7a2394956f8f82670e10859efd6b1c67c4487033/src/interpreter/Interpreter.ts#L85-L86

This PR aims to handle correctly oneOf in accordance to the OpenAPI specs https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof-vs-oneof

I needed to update one of the tests because there's no reason for oneOf to generate an array type implicitly in the datamodel definition

```
Test Suites: 115 passed, 115 total
Tests:       699 passed, 699 total
Snapshots:   123 passed, 123 total
Time:        40.322 s
Ran all test suites.
```

**Related issue(s)**

Fixes #367

### DIFF
--- a/src/generators/typescript/TypeScriptRenderer.ts
+++ b/src/generators/typescript/TypeScriptRenderer.ts
@@ -61,6 +61,9 @@ export abstract class TypeScriptRenderer extends AbstractRenderer<TypeScriptOpti
     if (Array.isArray(model.type)) {
       return [... new Set(model.type.map(t => this.toTsType(t, model)))].join(' | ');
     }
+    if (Array.isArray(model.items) && model.originalInput.oneOf !== undefined) {
+      return model.items.map(t => this.renderType(t)).join(' | ');
+    }
     return this.toTsType(model.type, model);
   }
 
--- a/src/interpreter/InterpretOneOf.ts
+++ b/src/interpreter/InterpretOneOf.ts
@@ -1,0 +1,21 @@
+import { CommonModel } from '../models/CommonModel';
+import { Interpreter, InterpreterOptions, InterpreterSchemaType } from './Interpreter';
+
+/**
+ * Interpreter function for oneOf keyword.
+ * 
+ * It puts the schema reference into the items field.
+ * 
+ * @param schema 
+ * @param model 
+ * @param interpreter 
+ * @param interpreterOptions to control the interpret process
+ */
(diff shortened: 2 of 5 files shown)


jonaslagoni (changes requested):
Hey @jacopomaroli, thanks for providing this!

I really dont want to change what's currently on the master branch, because this change is really difficult to do right in that environment.

Do you have the possibility to recreate it and target `next`? 

I can guide you through the change it require 🙂

jonaslagoni:
Fixed in `next` here: https://github.com/asyncapi/modelina/pull/899

PR #839, opened 2022-08-04
feat: generate test along side models

jonaslagoni:
**Description**
This PR is currently is a proof of concept for how to generate test along side the model. Eventually, if we want to pursue further this can be converted to a proper PR.

Requirements for the setup:
1. The test code MUST be extendable by each preset, so each can add their own test code for the specific functionality that they add.
2. It MUST be tested at runtime, so we can spin the test code against the model code to ensure maximum assurance that we generate the "correct" thing.

**Todo** 
These are the todo items that needs to be solved before this can become anything other then a draft.
- [ ] Figure out what to do with test dependencies, as currently only model have it.
- [ ] Fix the reference "example()" value generation, as it does not work.
- [ ] Fix the constructor parameters. Figure out what to do about required properties.
- [ ] Figure out how to better setup tests and checks when we have to consider multiple libraries and not just one.


jonaslagoni:

Issue #825, opened 2022-07-28
Duplicate dependencies are being rendered


jonaslagoni:
#### Describe the bug
Related to `next` branch.

If a model has duplicate dependencies i.e. two properties that both reference another model. Two dependencies will be rendered which is unintended.

Think something like this should work:

```ts

    //Filter out any duplicate dependencies
    dependencyModels = dependencyModels.filter((referenceModel) => {
      return dependencyModels.indexOf(referenceModel) === dependencyModels.lastIndexOf(referenceModel);
    });

    test('should not return duplicate dependencies', () => {
      const stringModel = new StringModel('', undefined);
      const referenceModel = new ReferenceModel('', undefined, stringModel);
      const referenceObjectPropertyModel = new ObjectPropertyModel('reference', false, referenceModel);
      const stringObjectPropertyModel = new ObjectPropertyModel('string', false, stringModel);
      const rawModel = new ObjectModel('test', undefined, {
        reference: referenceObjectPropertyModel,
        reference2: referenceObjectPropertyModel,
        string: stringObjectPropertyModel
      });
      
      const model = constrainMetaModel(mockedTypeMapping, mockedConstraints, {
        metaModel: rawModel,
        constrainedName: '',
        options: undefined
      }) as ConstrainedObjectModel;
      const dependencies = model.getNearestDependencies();
      expect(dependencies).toHaveLength(1);
      expect(dependencies[0]).toEqual(model.properties['reference'].property);
    });
```

PR #842, opened 2022-08-04
fix: duplicate dependencies being rendered

jonaslagoni:
**Description**
This PR fixes the duplicate dependency issue.

**Related issue(s)**
fixes https://github.com/asyncapi/modelina/issues/825


src/models/ConstrainedMetaModel.ts
```diff
@@ -55,6 +55,9 @@ export class ConstrainedTupleModel extends ConstrainedMetaModel {
     ).map((tupleModel) => {
       return tupleModel.value = tupleModel.value as ConstrainedReferenceModel;
     });
+    //Ensure no duplicate references
+    dependencyModels = [...new Set(dependencyModels)];
```
magicmatatjahu, line 59: You can create some helper function to do this, but it's not mandatory.

magicmatatjahu (approved):

src/models/ConstrainedMetaModel.ts
jonaslagoni: That's true but yea, just going to keep it as is 🙂

PR #849, opened 2022-08-08
chore: remove unused functions

jonaslagoni:
**Description**
This PR removes some unused functions.


src/helpers/TypeHelpers.ts
```diff
@@ -1,29 +1,6 @@
 import { CommonModel } from '../models';
```
Samridhi-98, line 1: All the unused import statements can be removed 😊

Samridhi-98 (changes requested):

src/helpers/TypeHelpers.ts
magicmatatjahu: Yeah, you have even lint error with that in testing workflow.

jonaslagoni, line 1:
```suggestion
```

PR #666, opened 2022-03-03
feat: add JSON Schema support for generators

ron-debajyoti:
**Description**

- Added base files
- ...
- ...

**Related issue(s)**

Fixes #532

Issue #612, opened 2022-02-03
Testing generated code in-depth


jonaslagoni:
### What we are trying to solve?
Modelina is aiming to be the defacto standard for generating data models, whether it is to [generate them in templates](https://github.com/asyncapi/generator#overview) or somewhere else. 

Imagine Modelina has to support +50 different inputs, +50 different outputs, with +100 different configurations in each output language. How do we ensure that we always produce the expected data models and they work at runtime?

Currently, we only test that we generate expected output through [snapshot matching](https://github.com/asyncapi/modelina/blob/7b2a33f65c06795c36c1b995334665a327068698/test/generators/typescript/preset/MarshallingPreset.spec.ts#L46) and that they can [compile/transpile](https://github.com/asyncapi/modelina/blob/master/test/blackbox/blackbox.spec.ts). None of which ensures that the generated output is semantically correct.

For example, say we generate Java data models that [overwrite the `hashcode` method](https://github.com/asyncapi/modelina/tree/master/examples/java-generate-hashcode), we want to ensure that it does exactly what it is supposed to do at runtime. And that is that one instance of a model should match against another instance when they are created with the same values. This means we need to be able to write tests such as the following:

```java
@Test
public void testHashcode_Symmetric() {
    Person x = new Person("Foo Bar");  // equals and hashCode check name field value
    Person y = new Person("Foo Bar");
    Assert.assertTrue(x.hashCode() == y.hashCode());
}
@Test
public void testHashcode_NonSymmetric() {
    Person x = new Person("Foo Bar");
    Person y = new Person("Foo");
    Assert.assertFalse(x.hashCode() == y.hashCode());
}
```

So we need to figure out how we from TS can run and potentially generate tests in all of the output languages 🤯 But don't worry, we start small and keep iterating.

### GSOC status

**Difficulty**: HARD (MEDIUM, if you are already well established in the output languages)

**What you will learn**:
- What Modelina is and how it works.
   - This includes the basics of AsyncAPI, JSON Schema, and OpenAPI. 
- GitHub CI and workflows used by Modelina.
- How to work with code generation.
- How you can test code generation in other projects such as code templates (same principles apply).
- How to create basic projects and tests in Java
- How to create basic projects and tests in C#
- How to create basic projects and tests in Go

**Requirements**:
- You should have basic knowledge of TypeScript as this is the language Modelina is written in.
- Hunger for working in multiple languages and with code generation.
[...]

PR #854, opened 2022-08-10
feat: add jsonbinpack preset

jonaslagoni:
**Description**
This PR adds jsonbinpack preset that enables you to convert your models into highly space-efficient binary. This is an alternative to just converting the models into JSON.

This preset does have contingencies:
1. Since the original input does not contain metadata about it, it is assuming draft 7 as default. But it does have support for draft 4 and 6. And because AsyncAPI schema object is a draft 7 variant, that is supported as well.
2. Once a JSON schema generator is supported - https://github.com/asyncapi/modelina/pull/666 - it might be worth within the preset to use that generator. That way any input would be able to still use this preset.

Blocked by https://github.com/asyncapi/modelina/pull/855


Samridhi-98 (dismissed):
LGTM🎉

docs/languages/TypeScript.md
```diff
@@ -45,6 +46,20 @@ This can be done by including the preset `TS_COMMON_PRESET` using the option `ma
...
+
+This functionality is for the library [jsonbinpack](https://github.com/sourcemeta/jsonbinpack).
+
+This preset can ONLY be used with AsyncAPI 2.x and JSON Schema draft 4 to 7 inputs.
+
+This functionality has two requirements:
+1. You MUST manually install the library `jsonbinpack`.
+2. You MUST also use the [Generate un/marshal functions for classes](#generate-unmarshal-functions-for-classes)
+
+This feature enables you to conver the models into a buffer that is highly space efficient, this is instead of sending pure JSON data across the wire.
```
magicmatatjahu, line 59:
```suggestion
This feature allows you to convert models to a buffer, which is highly space-efficient, instead of sending pure JSON data over the wire.
```

src/generators/typescript/presets/JsonBinPackPreset.ts
```diff
@@ -0,0 +1,55 @@
...
+ * 
+const instance = new Test({});
+instance.email = 'testing@test.com';
+const expectedJSON = JSON.parse(instance.marshal());
+const jsonpackbuffer = await instance.jsonbinSerialize();
+const test = await Test.jsonbinDeserialize(jsonpackbuffer);
+const actualJson = JSON.parse(test.marshal());
+expect(actualJson).toEqual(expectedJSON);
+ * 
+ */
```
magicmatatjahu, line 55: Do we need that here?

src/generators/typescript/presets/JsonBinPackPreset.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+    if(originalInput.$schema.includes('http://json-schema.org/draft-06/schema')) {
+      return 'draft6';
+    }
+  }
+  return 'draft7';
+}
+
+/**
+ * Preset which adds descriptions
```
magicmatatjahu, line 17:
```suggestion
 * Preset which adds jsonbinpack marshalling/unmarshalling methods
```

src/generators/typescript/presets/JsonBinPackPreset.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @implements {TypeScriptPreset}
+ */
+export const TS_JSONBINPACK_PRESET: TypeScriptPreset = {
+  class: {
+    async additionalContent({renderer, content, model}) {
+      renderer.addDependency('const jsonbinpack = require(\'jsonbinpack\');');
```
magicmatatjahu, line 24: Hmmm... Here you use CommonJS imports. I don't remember but we have a possibility to render dependencies in ESM or CJS, or am I wrong?

docs/languages/TypeScript.md
```diff
@@ -45,6 +46,20 @@ This can be done by including the preset `TS_COMMON_PRESET` using the option `ma
...
+
+Check out this [example out for a live demonstration](../../examples/typescript-generate-jsonbinpack/).
```
magicmatatjahu, line 61: Hmm, live demonstration sentence is usually used for browser app, so maybe

```suggestion
Check out this [example](../../examples/typescript-generate-jsonbinpack/), which uses a preset.
```

src/generators/typescript/presets/JsonBinPackPreset.ts
jonaslagoni: Not necessarily, I just included it because that was what I used to validate it worked, eventually, with #612 this is the code that should be run, so did not want to loose it 😄

src/generators/typescript/presets/JsonBinPackPreset.ts
jonaslagoni: Ahh, yea, you are right, good catch!

🤔 Need to create some support functions for this then!

src/generators/typescript/presets/JsonBinPackPreset.ts
magicmatatjahu: So if you wanna leave it, please move to the corresponding test file :)

docs/languages/TypeScript.md
jonaslagoni: This is what we use across all other docs, I am all up for changing it to something else, but let's stay consistent 🙂

jonaslagoni:
Leaving it here for later when implementing #612 

```

/**
 * Potential runtime test:
 * 
const instance = new Test({});
instance.email = 'testing@test.com';
const expectedJSON = JSON.parse(instance.marshal());
const jsonpackbuffer = await instance.jsonbinSerialize();
const test = await Test.jsonbinDeserialize(jsonpackbuffer);
const actualJson = JSON.parse(test.marshal());
expect(actualJson).toEqual(expectedJSON);
 * 
 */
```

src/generators/typescript/presets/JsonBinPackPreset.ts
jonaslagoni: Removed it 🙂

docs/migrations/version-0-to-1.md
```diff
@@ -36,5 +36,23 @@ These are all the preset hook changes:
 - C#, enum preset hooks had access to `model: CommonModel` which has now been changed to `model: ConstrainedEnumModel`
 
 General changes:
-- Hooks that gave access to properties/fields formally had the arguments `property`, `propertyName` and `type`, these are now wrapped within the [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) and can be accessed through the `property` argument.
-- Hooks that gave access to enum items, now has the type [`ConstrainedEnumValueModel`](./internal-model.md#the-constrained-meta-model).
\ No newline at end of file
+- Hooks that gave access to properties/fields formally had the arguments `property`, `propertyName` and `type`, these are now wrapped within the [`ConstrainedObjectPropertyModel`](../internal-model.md#the-constrained-meta-model) and can be accessed through the `property` argument.
+- Hooks that gave access to enum items, now has the type [`ConstrainedEnumValueModel`](../internal-model.md#the-constrained-meta-model).
+
+## TypeScript and JavaScript module system
```
jonaslagoni, line 42: OBS: Ended up having to move the module system option around for presets to access it 🙂 Which I had to ripple to the JavaScript generator as well.

magicmatatjahu (approved):
🚀

Samridhi-98 (approved):

PR #855, opened 2022-08-10
fix: unwrapping dictionaries not working in marshalling preset

jonaslagoni:
**Description**
This PR fixes an issue with dictionaries that should be unwrapped and are not in the TypeScript preset.


Samridhi-98 (approved):
LGTM 🎉

PR #857, opened 2022-08-11
docs: update usage documentation

jonaslagoni:
**Description**
This PR updates the usage documentation to add specifics about the limitations of inputs.


Samridhi-98 (dismissed):

docs/interpretation_of_JSON_Schema.md
```diff
@@ -22,7 +22,7 @@ The order of interpretation:
 - `dependencies` only apply to schema dependencies, since property dependencies adds nothing to the underlying model. Any schema dependencies are interpreted and then [merged](#Merging-models) together with the current interpreted model.
 - `enum` is interpreted as is, where each `enum`. Usage of `enum` infers the enumerator value type to the model, but only if the schema does not have `type` specified.
 - `const` interpretation overwrite already interpreted `enum`. Usage of `const` infers the constant value type to the model, but only if the schema does not have `type` specified.
-- [oneOf/anyOf/then/else](#Processing-sub-schemas)
+- [allOf/oneOf/anyOf/then/else](#Processing-sub-schemas)
```
magicmatatjahu, line 25: You missed:

```suggestion
- [allOf/oneOf/anyOf/if/then/else](#Processing-sub-schemas)
```

or am I wrong? 😄

docs/usage.md
```diff
@@ -30,74 +36,92 @@ For more specific integration options, please check out the [integration documen
 
 ## Understanding the output format
 
-TODO
+The output format is designed for you to use the generated models in further contexts. It might be part of a larger code generation such as AsyncAPI templates. This means that you can glue multiple models together into one large file, or split it out as you see fit.
 
 ## Generate models from AsyncAPI documents
 
-When providing an AsyncAPI document, the library iterates the entire document and generate models for all defined messages. If any other kind of iteration is wanted, feel free to create a [feature request](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md).
+When providing an AsyncAPI document, the library iterates the entire document and generate models for all defined message payloads. The message payloads are interpreted based on the schema format.
```
magicmatatjahu, line 43: We should clarify it that custom formats are converted to the JSON Schema, and then that JSON Schema follows the JSON Schema input processing.

docs/usage.md
```diff
@@ -30,74 +36,92 @@ For more specific integration options, please check out the [integration documen
...
+- [Generate from an AsyncAPI 2.x JS object](../examples/asyncapi-from-object)
 
 The library expects the `asyncapi` property for the document to be set in order to understand the input format.
 
-The message payloads, since it is a JSON Schema variant, is [interpreted as a such](./interpretation_of_JSON_Schema.md).
+### Limitations and Compatibility
+These are the current known limitation of the AsyncAPI input.
+
+#### Message Schema formats
+Currently, only a limited number of schema formats are supported and we currently rely on the [AsyncAPI parser](github.com/asyncapi/parser-js/) to handle those schema formats and convert them into [AsyncAPI Schema format](https://github.com/asyncapi/parser-js/#custom-message-parsers). At some point in the future, Modelina will support all native schema formats, so it does not matter which standard you use to define the message payloads, you will be able to generate models from it.
```
magicmatatjahu, line 58: If the parser will not be able to support the given formats then the Modelina itself should not either. It seems to me that this paragraph is badly formatted, because even if Modelina will support e.g `xml` the parser still has to convert it to JSON first, or do you have a plan to do it in Modelina side?

docs/usage.md
```diff
@@ -30,74 +36,92 @@ For more specific integration options, please check out the [integration documen
...
+Long term if this is something you wish was supported, voice your [opionion here](https://github.com/asyncapi/modelina/issues/108).
+
 ## Generate models from TypeScript type files
 
 Currently, we support generating models from a TypeScript type file.
 
 - [Generate Java model from a TypeScript file](../examples/java-from-typescript-type/)
 
-The TypeScript input processor expects that the typescript file and base directory where it's present, is passed as input, in order to process the types accurately.
+The TypeScript input processor expects that the typescript file and base directory where it's present, is passed as input, in order to process the types accurately. The input processor converts the TypeScript types into JSON Schema, which is [interpreted as a such](./interpretation_of_JSON_Schema.md).
```
magicmatatjahu, line 118: interpreted as a such? 😄

docs/interpretation_of_JSON_Schema.md
jonaslagoni: You are, because the `if` does not force any new structure for the schema, so there is no need to process it. `if` is only used to determine if `then` or `else` should be applied to the data 🙂

docs/usage.md
jonaslagoni: Technically, we should not convert the payload to JSON Schema, but instead determine based on the schemaFormat which other input processor we should use to process the message payload 🤔 I think I left it out on purpose for this change 🧐

docs/usage.md
jonaslagoni: Think https://github.com/asyncapi/modelina/pull/857/files#r968490085 answers is right? 🤔

docs/usage.md
magicmatatjahu: So then you will have to write your own logic to handle custom schemas and modelina will use parser-js only to check spec if is valid 🤔

docs/usage.md
magicmatatjahu: Yeah, please, it's hard to deduce.

docs/usage.md
jonaslagoni: Basically yes. 

We still yet have to see how we handle non-json payloads, cause it might happen we have to do some kind of handling there. But otherwise yes not much the AsyncAPI input processor needs to do 😄

docs/usage.md
jonaslagoni: Alright, I changed how we linked to the JSON Schema processor, let me know if it's sufficient :)

magicmatatjahu (approved):

Issue #49, opened 2021-02-11
Implement Python generator and renderers


magicmatatjahu:
Implement Python generator and renderers like in TS or Java.

Generator should have possibility to define presets for `class`, `enum` type.

Additional: create presets for serialisation/converters from json.

joe-narvaez:
If we could get dataclass and/or pydantic support, that would be ideal.

humbleengineer:
> If we could get dataclass and/or pydantic support, that would be ideal.

Double thumbs up to that!

derberg:
@damaru-inc hey, this one might be interesting for you. Since you created https://github.com/asyncapi/python-paho-template/ it means to me you are a good candidate to contribute here. Side effect is that later you can use Modelina in your template to generate some Python code

jonaslagoni:
Tasklist to implement it:

- [ ] Duplicate one of the existing generators, for example, the Java generator, (I find it easier to start with that) https://github.com/asyncapi/modelina/tree/master/src/generators
- [ ] Start with empty generator options
  - https://github.com/asyncapi/modelina/blob/57948b508f9f5472f6ee14b2ac130d5e5fee9c1d/src/generators/java/JavaGenerator.ts#L17
  - https://github.com/asyncapi/modelina/blob/57948b508f9f5472f6ee14b2ac130d5e5fee9c1d/src/generators/java/JavaGenerator.ts#L13
- [ ] Adapt reserved keyword list to match it for python - https://github.com/asyncapi/modelina/blob/master/src/generators/typescript/Constants.ts
- [ ] Adapt the core PythonRenderer to render python types - https://github.com/asyncapi/modelina/blob/57948b508f9f5472f6ee14b2ac130d5e5fee9c1d/src/generators/typescript/TypeScriptRenderer.ts#L14
- [ ] I find it easiest to start with [two renders](https://github.com/asyncapi/modelina/tree/master/src/generators/java/renderers) (those that call presets at specific times during the rendering process), one for `Class` and one of `Enum`.
- [ ] Adapt the code being generated for the Class and Enum to match Python.
- [ ] Start with no [presets](https://github.com/asyncapi/modelina/tree/master/src/generators/typescript/presets) as we just start with the bare minimum.
- [ ] Adapt all releated docs. 


For testing (yes we need tests 😄) I would again start by copying one of the existing test collections for a generator https://github.com/asyncapi/modelina/tree/master/test/generators/typescript and then remove anything that is not related to what you have in the Python generator.

[...]

PR #604, opened 2022-01-28
feat: implemented generator and renderer for python

mahakporwal02:
Fixes #49 
I have implemented `ClassRenderer.ts` right now.
I'll implement `EnumRenderer.ts` once my current implementation of `ClassRenderer.ts` has been approved.

PR #863, opened 2022-08-11
feat: add python generator

jonaslagoni:
**Description**
This PR extends the already existing work from @mahakporwal02 in https://github.com/asyncapi/modelina/pull/604 and adds a python generator for both class and enums.

cc @liquidautumn @niraveen as you showed interest in the other PR, I am just pinging you in case you wish to review it 🙂 

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/49


src/generators/python/renderers/ClassRenderer.ts
```diff
@@ -0,0 +1,97 @@
...
+${renderer.indent(body)}`;
+  },
+  getter({ property }) {
+    return `@property
+def ${property.propertyName}(self):\n\treturn self._${property.propertyName}`;
+  },
+  setter({ property }) {
+    return `@${property.propertyName}.setter
+def ${property.propertyName}(self, ${property.propertyName}):\n\tself._${property.propertyName} = ${property.propertyName}`;
+  },
```
magicmatatjahu, line 96: Should we use decorators as default behaviour? I don't know Python at all, so I only ask.

magicmatatjahu (commented):
One question related to code. It would be best if someone who knows Python does a review :)

src/generators/python/renderers/ClassRenderer.ts
jonaslagoni: This was what https://github.com/asyncapi/modelina/pull/604 used so I just copied it 🙂 

You will also likely not even reach a consensus as it's preference

magicmatatjahu (approved):
Accept but I don't know Python so please wait for other reviewers :)

jonaslagoni:
> Accept but I don't know Python so please wait for other reviewers :)

Why? 😄 There was no objections in the last pr and we have Black Box tests to ensure it works.

If something else is desired from the models we get those requests down the line. So i dont see any reason to wait 😄 Unless you know someone who knows python that wants to review it 🤔?

jonaslagoni:
Gonna take the decision to just merge this now, as we can always backtrack in next generation if something is completely wrong.

PR #867, opened 2022-08-16
feat: add Go tags support and JSON preset

jonaslagoni:
**Description**
This PR is a rewrite of https://github.com/asyncapi/modelina/pull/639 to target `next`. 

The last discussion that was left unanswered was https://github.com/asyncapi/modelina/pull/639#discussion_r815755537.

I honestly don't see the alternative as being better, i.e. rendering the tags within the renderer. And the reason is that it alters how presets usually works. The `fieldTag` should not be called for the specific property if you overwrite the entire field rendering. 

I can go into depths here if need be, but in short, this is completely fine to do:
```
    generator = new GoGenerator({
      presets: [
        {
          struct: {
            field({ content }) {
              return `//Something comment \n ${content}`;
            },
          }
        },
        {
          struct: {
            fieldTag({ content }) {
              return `${content} gorm:"primary_key"`;
            },
          }
        }
      ]
    });
```
Where the following, will result in the `fieldTag` in the second preset never to be called.
```
    generator = new GoGenerator({
      presets: [
        {
          struct: {
            field({ content }) {
              return `StringProp string`;
            },
          }
        },
        {
          struct: {
            fieldTag({ content }) {
              return `${content} gorm:"primary_key"`;
            },
          }
        }
      ]
    });
```

This is exactly how I expect it to work with the limitations of presets.

**Related issue(s)**
Extension of https://github.com/asyncapi/modelina/pull/639
Fixes https://github.com/asyncapi/modelina/issues/299


test/generators/go/presets/__snapshots__/JsonTagPreset.spec.ts.snap
```diff
@@ -0,0 +1,49 @@
...
+  EnumTestNumber_2
+)
+
+// Value returns the value of the enum.
+func (op EnumTest) Value() any {
+	if op >= EnumTest(len(EnumTestValues)) {
+		return nil
+	}
+	return EnumTestValues[op]
+}
```
magicmatatjahu, line 31: You can have even enums as string, so I think that we should somehow base logic on type of enum.

https://www.sohamkamani.com/golang/enums/

cc @smoya

jonaslagoni: Pretty sure, without fully remembering it, was this not the desired behaviour we sketched out Sergio?

magicmatatjahu: hmm, I don't remember, cc @smoya

jonaslagoni: Okay, the implementation is insufficient to support encoding and decoding enums and structs into and from JSON. Gonna make some alterations and re-request a review once done 🙂

smoya: @magicmatatjahu what about supporting multiple types inside those enum? Not all values will be `string`

magicmatatjahu: But it's a limitation of enums in languages, even in TS you can have string or number values. In this case you can define enum values as string or numbers. That's my point.

smoya: > But it's a limitation of enums in languages, even in TS you can have string or number values. In this case you can define enum values as string or numbers. That's my point.

There is one missing piece @jonaslagoni didn't add to this code but it was in the implementation we baked together:

```go
func (op *EnumTest) UnmarshalJSON(raw []byte) error {
	var v any
	if err := json.Unmarshal(raw, &v); err != nil {
		return err
	}
	*op = ValuesToEnumTest[v]
	return nil
}
```

With that, when unmarshaling, the value will be transformed to (and only to) the `uint` const.
There is a case that won't work, and its when the value is an object, because that `ValuesToEnumTest` map can't have a key that is a map. However I'm gonna give another though to it.

coreywagehoft:
Are there any updates when this will be fixed and merged?  Without the ability to generate tags on go structs we cannot adopt using asyncapi.

jonaslagoni:
> Are there any updates when this will be fixed and merged? Without the ability to generate tags on go structs we cannot adopt using asyncapi.

I did plan to finish this before releasing the `next` branch 🙂

jonaslagoni:
This is gonna need a bit more tinkering, but it's still a feature change, so it should not be a problem to introduce after the release of the next branch.

jonaslagoni:

PR #868, opened 2022-08-16
refactor: simplified example tests

### LINKED ISSUE #864: Simplify example tests
#### Reason/Context
The examples always run the [generate function](https://github.com/asyncapi/modelina/blob/527cc6a84f6ae5229d9b352fcbfc4b0ad8249036/examples/adapting-input-and-output/index.ts#L54), It is, however, a problem from a testing point of view the [models are generated twice](https://github.com/asyncapi/modelina/blob/527cc6a84f6ae5229d9b352fcbfc4b0ad8249036/examples/adapting-input-and-output/index.spec.ts#L11).

Instead, it should only be generated once.

To fix this we need to add the following wrapper around each generate call in the example index file:
```ts
if (require.main === module) {
  generate();
}
```
This means that it only generates the models if run from the CLI (i.e. `npm run start`).

We can then adapt the tests to only be called 1x instead of 2x. This means [changing this](https://github.com/asyncapi/modelina/blob/527cc6a84f6ae5229d9b352fcbfc4b0ad8249036/examples/adapting-input-and-output/index.spec.ts#L9) to:
```ts
    await generate();
    expect(spy.mock.calls.length).toEqual(1);
    expect(spy.mock.calls[0]).toMatchSnapshot();
```

nitintejuja:
**Description**

- Simplified adapting-input-and-output index.ts file as `generate` function was getting called 2 times and updated spec files accordingly.

**Related issue(s)**
Fixes #864


jonaslagoni (changes requested):
Hey @nitintejuja, the issue that is referred is actually to change this for all the examples 🙂 Might not have been specified that well enough 🤔 Do you wish to apply the change to the rest of the examples, or just this one?

Also, please make sure you target the `next` branch and create your branch from it, otherwise, you will get a bunch of conflicts to solve 🙂 

Let me know if you have any problems.

nitintejuja:
@jonaslagoni i'm facing issue in 2 tests . I have made required changes but its not working .
below are testfiles
`examples/typescript-generate-comments/index.spec.ts`
`examples/typescript-enum-type/index.spec.ts`

can you help me ?

jonaslagoni (commented):
Finally back from holiday, let me check @nitintejuja!

examples/typescript-enum-type/index.spec.ts
```diff
@@ -9,15 +9,15 @@ describe('Should be able to render correct enums based on options', () => {
   });
   test('and should log expected output to console', async function testIt(): Promise<void> {
     // Wait for first call to the `generate` function is over, and we can reliably know the output
-    if (spy.mock.calls.length !== 6) {
+    if (spy.mock.calls.length !== 3) {
       await timeout(50);
       return testIt();
     }
```
jonaslagoni, line 15:
```suggestion
```

You can safely remove this 🙂

examples/generate-go-enums/index.spec.ts
```diff
@@ -8,7 +8,7 @@ describe('Should be able to render Go Enums', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 3 model (root and three different enums type), we double it
```
jonaslagoni, line 10:
```suggestion
```

examples/java-from-typescript-type/index.spec.ts
```diff
@@ -8,7 +8,7 @@ describe('Should be able to generate Java model from a TypeScript type file', ()
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 4x, so even though we expect 1 model, we double it
```
jonaslagoni, line 10:
```suggestion
```

examples/javascript-use-cjs/index.spec.ts
```diff
@@ -8,8 +8,7 @@ describe('Should be able to render models to ESM module system', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 2 models, we double it
```
jonaslagoni, line 10:
```suggestion
```

examples/javascript-use-esm/index.spec.ts
```diff
@@ -8,8 +8,7 @@ describe('Should be able to render models to ESM module system', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 2 models, we double it
```
jonaslagoni, line 10:
```suggestion
```

examples/openapi-from-object/index.spec.ts
```diff
@@ -7,8 +7,8 @@ describe('Should be able to process a pure OpenAPI object', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 2 models, we double it
```
jonaslagoni, line 9:
```suggestion
```

examples/typescript-use-cjs/index.spec.ts
```diff
@@ -8,8 +8,7 @@ describe('Should be able to render models to ESM module system', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 2 models, we double it
```
jonaslagoni, line 10:
```suggestion
```

examples/typescript-use-esm/index.spec.ts
```diff
@@ -8,8 +8,7 @@ describe('Should be able to render models to ESM module system', () => {
   test('and should log expected output to console', async () => {
     await generate();
     //Generate is called 2x, so even though we expect 2 models, we double it
```
jonaslagoni, line 10:
```suggestion
```

jonaslagoni (changes requested):
I think you just have to update the snapshots 🙂 You can do this by running `npm run test:examples -- -u`

examples/typescript-enum-type/index.spec.ts
nitintejuja: i have removed this piece of code but still i'm getting failed test for this spec file. can you help me ?
Thanks

jonaslagoni: Have you updated the snapshots? `npm run test -- -u`

nitintejuja: yes but on snapshot creation it gives error.

jonaslagoni: Which error are you getting?

nitintejuja: im getting below error 
`FAIL examples/typescript-enum-type/index.spec.ts (18.144 s)
  ● Should be able to render correct enums based on options › and should log expected output to console

    expect(received).toMatchSnapshot()

    Snapshot name: `Should be able to render correct enums based on options and should log expected output to console 1`

    - Snapshot  - 14
    + Received  +  0

    @@ -1,19 +1,5 @@
    - Generator output with Union:
    - class Root {
    -   private _eventType?: EventType;
    -
    -   constructor(input: {
    -     eventType?: EventType,
    -   }) {
    -     this._eventType = input.eventType;
    -   }
    -
    -   get eventType(): EventType | undefined { return this._eventType; }
    -   set eventType(eventType: EventType | undefined) { this._eventType = eventType; }
    - }
    - type EventType = "ping" | "pong";
      Generator output with Enum:
      class Root {
        private _eventType?: EventType;

        constructor(input: {

      12 |     expect(spy.mock.calls.length).toEqual(6);
      13 |     const generatedContent = spy.mock.calls.slice(3).join('\n');
    > 14 |     expect(generatedContent).toMatchSnapshot();
         |                              ^
      15 |   });
      16 | });
      17 |

      at examples/typescript-enum-type/index.spec.ts:14:30
      at fulfilled (examples/typescript-enum-type/index.spec.ts:5:58)
`

examples/typescript-enum-type/index.spec.ts
```diff
@@ -7,17 +7,10 @@ describe('Should be able to render correct enums based on options', () => {
...
-      return testIt();
-    }
-
+  test('and should log expected output to console', async (): Promise<void> => {
     await generate();
-    //Generate is called 2 times, so even though we expect 6 console logs we double it
-    expect(spy.mock.calls.length).toEqual(12);
-    const generatedContent = spy.mock.calls.slice(6).join('\n');
+    expect(spy.mock.calls.length).toEqual(6);
+    const generatedContent = spy.mock.calls.slice(3).join('\n');
```
jonaslagoni, line 13:
```suggestion
    const generatedContent = spy.mock.calls.join('\n');
```

jonaslagoni (changes requested):
Ahh damn, found out the issue 🤦 I recently changed the test script which means that it is no longer possible to update the snapshots like this 🙃 

After applying the above comment you need to run the following command `./node_modules/.bin/jest  ./examples --testPathIgnorePatterns ./examples/integrate-with-react --testPathIgnorePatterns ./examples/TEMPLATE -u` to update all the snapshots...

So sorry for this!

examples/typescript-enum-type/index.spec.ts
jonaslagoni: please see https://github.com/asyncapi/modelina/pull/868#pullrequestreview-1117781992

examples/typescript-enum-type/index.spec.ts
nitintejuja: now its working fine and all testcases passed. Just one more doubt, should i commit latest snapshots ?

jonaslagoni: Yes please :)

nitintejuja: done :heavy_check_mark:

jonaslagoni (approved):

jonaslagoni:
Thanks for the hard work @nitintejuja! This is such a great change!

@all-contributors please add @nitintejuja for test

Issue #847, opened 2022-08-07
RustGenerator Blackbox tests failing


leigh-johnson:
#### Describe the bug
Couple of RestGenerator blackbox tests are failing in the `next` mainline branch. 

#### How to Reproduce
`npm run test:blackbox`

#### Expected behavior
Tests pass :stuck_out_tongue_closed_eyes: 

----

- [ ] `Error: Error: Command failed: cargo build /home/runner/work/modelina/modelina/test/blackbox/output/JsonSchemaDraft-4/traefik-v2/rust/struct` 
[link](https://github.com/asyncapi/modelina/runs/7707920632?check_suite_focus=true#step:9:257)
Should be `cargo build --manifest-path <test dir>`

- [ ] `TypeError: Cannot read property '0' of undefined`
[link](https://github.com/asyncapi/modelina/runs/7707920632?check_suite_focus=true#step:9:237)
Revisit the way enum member names are generated. The current implementation assumes a valid JSON draft 07 enum definition is passed and fails for all other inputs.

leigh-johnson:
:wave: Just a head's up, I've blocked out time to address this issue November 14 - 18. Finally! Sorry for the long wait, I had a lot of pre-work to complete before moving PrintNanny.ai over to a NATS + AsyncAPI event stack.

jonaslagoni:
AS this is a testing problem and does not affect anything other than the black-box tests, I think we can leave this out from the v1 milestone 🙂 It's just something that needs to be fixed at some point 😅

AnimeshKumar923:
@jonaslagoni still valid?

PR #869, opened 2022-09-03
fix: rust compile errors

YushiOMOTE:
**Description**

Locally tried Rust generator with more complicated private examples and got some compile errors in the generate Rust crate.
The PR makes compile pass fixing the following issues:

* Remove warnings in case the model name contains the dash + number (e.g. `Abc_123`)
* Compile errors due to mismatch of some Option field definitions and constructors in struct 
* Fix broken generated file names
* Fix invalid type name included in generated Union types
* Fix the missing qualifer `crate::` in some cases
* Fix undefined error when schemaUid doesn't exist

### DIFF
--- a/src/generators/rust/RustConstrainer.ts
+++ b/src/generators/rust/RustConstrainer.ts
@@ -253,7 +253,8 @@ export const RustDefaultTypeMapping: TypeMapping<RustOptions> = {
     return `${constrainedModel.name}`;
   },
   Array({ constrainedModel }): string {
-    return `Vec<${FormatHelpers.upperFirst(constrainedModel.valueModel.type)}>`;
+    let prefix = (constrainedModel.valueModel instanceof ConstrainedReferenceModel) ? 'crate::' : '';
+    return `Vec<${prefix}${FormatHelpers.upperFirst(constrainedModel.valueModel.type)}>`;
   },
   Enum({ constrainedModel }): string {
     return constrainedModel.name;
@@ -262,7 +263,9 @@ export const RustDefaultTypeMapping: TypeMapping<RustOptions> = {
     return constrainedModel.name;
   },
   Dictionary({ constrainedModel }): string {
-    return `std::collections::HashMap<${constrainedModel.key.type}, ${constrainedModel.value.type}>`;
+    let key_prefix = (constrainedModel.key instanceof ConstrainedReferenceModel) ? 'crate::' : '';
+    let value_prefix = (constrainedModel.value instanceof ConstrainedReferenceModel) ? 'crate::' : '';
+    return `std::collections::HashMap<${key_prefix}${constrainedModel.key.type}, ${value_prefix}${constrainedModel.value.type}>`;
   }
 };
 
--- a/src/generators/rust/RustConstrainer.ts
+++ b/src/generators/rust/RustConstrainer.ts
@@ -41,16 +41,84 @@ export function deriveCopy(model: ConstrainedMetaModel): boolean {
   return true;
 }
 
+export function derivePartialEq(model: ConstrainedMetaModel): boolean {
+  if (model instanceof ConstrainedAnyModel) { return false; }
+
+  if (
+    // all contents implement PartialEq trait
+    model instanceof ConstrainedArrayModel ||
+    model instanceof ConstrainedEnumModel ||
+    model instanceof ConstrainedObjectModel ||
+    model instanceof ConstrainedTupleModel ||
+    model instanceof ConstrainedUnionModel ||
+    model instanceof ConstrainedDictionaryModel
+  ) { return allPartialEq(model); }
+  return true;
+}
+
 export function deriveEq(model: ConstrainedMetaModel): boolean {
+  if (!derivePartialEq(model)) { return false }
+
   if (model instanceof ConstrainedFloatModel || model instanceof ConstrainedAnyModel) { return false; } else if (
     // all contents implement Eq trait
     model instanceof ConstrainedArrayModel ||
     model instanceof ConstrainedEnumModel ||
     model instanceof ConstrainedObjectModel ||
     model instanceof ConstrainedTupleModel ||
-    model instanceof ConstrainedUnionModel
+    model instanceof ConstrainedUnionModel ||
+    model instanceof ConstrainedDictionaryModel
   ) { return allEq(model); }
   return true;
+}
+
+export function derivePartialOrd(model: ConstrainedMetaModel): boolean {
+  if (
+    model instanceof ConstrainedAnyModel ||
+    model instanceof ConstrainedDictionaryModel
+  ) { return false; } else if (
+    // all contents implement PartialOrd trait
+    model instanceof ConstrainedArrayModel ||
+    model instanceof ConstrainedEnumModel ||
+    model instanceof ConstrainedObjectModel ||
+    model instanceof ConstrainedTupleModel ||
+    model instanceof ConstrainedUnionModel
+  ) { return allPartialOrd(model); }
+  return true;
+}
+
+export function deriveOrd(model: ConstrainedMetaModel): boolean {
+  if (!derivePartialOrd(model)) { return false }
+
+  if (
+    model instanceof ConstrainedFloatModel ||
+    model instanceof ConstrainedAnyModel ||
+    model instanceof ConstrainedDictionaryModel
+  ) { return false; } else if (
+    // all contents implement Ord trait
+    model instanceof ConstrainedArrayModel ||
+    model instanceof ConstrainedEnumModel ||
+    model instanceof ConstrainedObjectModel ||
+    model instanceof ConstrainedTupleModel ||
+    model instanceof ConstrainedUnionModel
+  ) { return allOrd(model); }
+  return true;
+}
+
+export function allPartialEq(model: ConstrainedMetaModel): boolean {
+  if (model instanceof ConstrainedUnionModel) {
+    return model.union.map(derivePartialEq).every(v => v === true);
+  } else if (model instanceof ConstrainedTupleModel) {
+    return model.tuple.map(v => derivePartialEq(v.value)).every(v => v === true);
+  } else if (model instanceof ConstrainedObjectModel) {
+    return Object.values(model.properties).map(p => derivePartialEq(p.property)).every(v => v === true);
+  } else if (model instanceof ConstrainedArrayModel) {
+    return derivePartialEq(model.valueModel);
+  } else if (model instanceof ConstrainedEnumModel) {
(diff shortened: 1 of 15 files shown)


leigh-johnson:
👋 Hey @YushiOMOTE, thank you for the contributions! Would you feel comfortable adding tests to demonstrate how the code breaks, and verify your fixes? Feel free to ping me if you're having trouble getting the tests (unit or black-box) running locally.

An excellent first step would be to add a test for `renderMacro` revisions to: https://github.com/asyncapi/modelina/blob/next/test/generators/rust/RustGenerator.spec.ts 

Can you share a schema with example data structures? That will make it easier to grok the behavior you're describing. Thanks!

The `ConstrainedUnionModel` isn't used because we use `ConstrainedEnumModel` to describe an enum of primitive type(s) and mixed/complex types ("union" in most languages). 

Regarding imports, the decision to render unit types and use fully-qualified `crate::` syntax was intentional. We follow this convention in the openapi-generator for data structures and fn signatures.
https://github.com/OpenAPITools/openapi-generator/blob/60dcf8613f6b75b606757c4a991fe018ad10ee99/samples/client/petstore/rust/reqwest/petstore/src/models/pet.rs#L26
https://github.com/OpenAPITools/openapi-generator/blob/master/samples/client/petstore/rust/reqwest/petstore/src/apis/pet_api.rs#L86

leigh-johnson:
link against #847 for ref

YushiOMOTE:
Thank you! Yes let me add tests. I cannot share the exact schema as it’s closed, but probably can share example schemas that can reproduce the issues.

src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -63,7 +63,7 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
     
     let schemaUid = schema.uid();
     //Because the constraint functionality of generators cannot handle -, <, >, we remove them from the id if it's an anonymous schema.
-    if (schemaUid.includes('<anonymous-schema')) {
+    if (typeof schemaUid !== 'undefined' && schemaUid.includes('<anonymous-schema')) {
```
leigh-johnson, line 66: I'm not confident about this revision, since it's to the AsyncAPI input processor used by all generators (not just the Rust generator). It'd help to see the input you're using. Maybe @jonaslagoni has better insight here.

test/generators/rust/presets/__snapshots__/CommonPreset.spec.ts.snap
```diff
@@ -2,18 +2,18 @@
 
 exports[`RUST_COMMON_PRESET Enum should render \`enum\` of union type with Default implementation of integer type 1`] = `
 "// Address represents a Address model.
-#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct Address {
     #[serde(rename=\\"members\\")]
-    members: Box<crate::Members>,
+    members: crate::Members,
```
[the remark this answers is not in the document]
leigh-johnson: I think we should keep references to models as pointer-heap memory by default. If you need the entire data structure to be loaded into stack memory, I think a preset would be the right mechanism for that behavior.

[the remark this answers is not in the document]
YushiOMOTE: Got it! I fixed to use `Box` for reference. Now the constructor side is fixed to use `Box`: https://github.com/asyncapi/modelina/pull/869/files#diff-9f4739c9c2692512b77fc209d671487d26937caed76f432ff6adf62a343a3ebaR18

YushiOMOTE:
> > Fix invalid type name included in generated Union types > (https://github.com/asyncapi/modelina/commit/78c44a2cd67fac5749dbf7405a849984c6d88d13)
>
> I'm trying to give the example swagger or add a test in PR to describe this case soon 👀

This spec generates the issue with union.
https://github.com/YushiOMOTE/modelina/commit/6f47a2ec5900f2c79809df4ec2531573913083b5

src/processors/AsyncAPIInputProcessor.ts
YushiOMOTE: I checked more. The issue seems from AsyncAPI side (not specific to Rust).

If one of the schemas in the spec is missing `type`, this error happens:

```
/Users/yushi.omote/dev/model-play/modelina/src/processors/AsyncAPIInputProcessor.ts:66
    if (schemaUid.includes('<anonymous-schema')) {
                  ^
TypeError: Cannot read properties of undefined (reading 'includes')
```

This is the minimal spec that reproduces the issue:

```yaml
asyncapi: 2.0.0
id: "https://github.com/yushiomote/hoge"
info:
  title: "Test"
  version: 2.0.0
defaultContentType: application/json
channels:
  hello:
    publish:
      message:
        name: hello
        payload:
          # type: object <- The issue happens if this is missing
          properties:
            data: { type: string }
```

Maybe we should define `type` (it's supposed to be `required` field?) but Asyncapi studio doesn't blame the missing type
Screen Shot 2022-09-11 at 14 42 46

YushiOMOTE: I can move this specific revision to a different PR opening a separate issue if needed 🙇

jonaslagoni: As this is not related rust specific problems, I think it makes sense to split it out if you can @YushiOMOTE 🙂

jonaslagoni: Solving in https://github.com/asyncapi/modelina/pull/876

leigh-johnson:
Sorry for the delay in responding! Thank you for sticking with this PR @YushiOMOTE 🙇‍♀️ 

The failing black-box tests are related to #874, so don't worry about those. 
https://github.com/asyncapi/modelina/actions/runs/3030534665/jobs/4877197283

Linter caught a couple of nits:
https://github.com/asyncapi/modelina/actions/runs/3030534665/jobs/4877197283

jonaslagoni:
Cant wait for this to be merged 😄 Just encountered the same issues 👍

src/generators/rust/RustConstrainer.ts
```diff
@@ -162,7 +263,9 @@ export const RustDefaultTypeMapping: TypeMapping<RustOptions> = {
     return constrainedModel.name;
   },
   Dictionary({ constrainedModel }): string {
-    return `std::collections::HashMap<${constrainedModel.key.type}, ${constrainedModel.value.type}>`;
+    let key_prefix = (constrainedModel.key instanceof ConstrainedReferenceModel) ? 'crate::' : '';
+    let value_prefix = (constrainedModel.value instanceof ConstrainedReferenceModel) ? 'crate::' : '';
```
leigh-johnson, line 267:
```suggestion
    const value_prefix = (constrainedModel.value instanceof ConstrainedReferenceModel) ? 'crate::' : '';
```

src/generators/rust/presets/CommonPreset.ts
```diff
@@ -68,15 +68,22 @@ export const RUST_COMMON_PRESET: RustPreset<RustCommonPresetOptions> = {
         const args: string[] = [];
         const fields: string[] = [];
         for (const v of Object.values(properties)) {
+          let prefix = (v.property instanceof ConstrainedReferenceModel) ? 'crate::' : '';
```
leigh-johnson, line 71:
```suggestion
          const prefix = (v.property instanceof ConstrainedReferenceModel) ? 'crate::' : '';
```

src/generators/rust/presets/CommonPreset.ts
```diff
@@ -68,15 +68,22 @@ export const RUST_COMMON_PRESET: RustPreset<RustCommonPresetOptions> = {
...
+          let fieldType = prefix + v.property.type;
```
leigh-johnson, line 72:
```suggestion
          const fieldType = prefix + v.property.type;
```

src/generators/rust/renderers/StructRenderer.ts
```diff
@@ -1,6 +1,6 @@
 import { RustRenderer } from '../RustRenderer';
 import { StructPresetType } from '../RustPreset';
-import { ConstrainedObjectModel, ConstrainedObjectPropertyModel, ConstrainedReferenceModel } from '../../../models';
+import { ConstrainedObjectModel, ConstrainedObjectPropertyModel, ConstrainedReferenceModel, ConstrainedUnionModel } from '../../../models';
```
leigh-johnson, line 3:
```suggestion
import { ConstrainedObjectModel, ConstrainedObjectPropertyModel, ConstrainedReferenceModel } from '../../../models';
```

src/generators/rust/renderers/UnionRenderer.ts
```diff
@@ -54,6 +54,10 @@ export const RUST_DEFAULT_UNION_PRESET: UnionPresetType<RustOptions> = {
     return `#[serde(${serdeArgs.join(', ')})]`;
   },
   item({ item }) {
-    return `${item.name}(${item.type})`;
+    if (item instanceof ConstrainedReferenceModel) {
+      return `${item.name}(crate::${item.ref.type})`;
+    } else {
+      return `${item.name}(${item.type})`;
+    }
```
leigh-johnson, line 61:
```suggestion
    }
    return `${item.name}(${item.type})`;
```

src/helpers/FormatHelpers.ts
```diff
@@ -82,6 +83,16 @@ export class FormatHelpers {
...
   static toPascalCase = pascalCase;
 
+  /**
+   * Transform into a string of capitalized words without separators
+   * merging numbers.
+   * @param {string} value to transform
+   * @returns {string}
+   */
+  static toPascalCaseMergingNumbers(value: string): string {
+    return pascalCase(value, { transform: pascalCaseTransformMerge })
```
leigh-johnson, line 93:
```suggestion
    return pascalCase(value, { transform: pascalCaseTransformMerge });
```

leigh-johnson (changes requested):
:wave: I'd love to merge this @YushiOMOTE, just a few suggestions to pass the linter (added in review).

leigh-johnson:
@YushiOMOTE Looks like the last thing left is to update this snapshot 🎉 👏 
https://github.com/asyncapi/modelina/actions/runs/3398564376/jobs/5652579555

leigh-johnson:
@YushiOMOTE Just a heads up, I'm going to merge this PR into a new branch, `cleanup-pr-869`, and do the last bit of cleanup. Thank you for submitting these revisions!

jonaslagoni:
@all-contributors
please add @YushiOMOTE for bug, code. 
please add @leigh-johnson for review.

PR #876, opened 2022-09-12
fix: undefined error when schemaUid isn't set

### LINKED ISSUE #875: Type error when `type` is omitted in schemas in AsyncAPI spec
#### Describe the bug

If one of the schemas in the AsyncAPI spec is missing `type`, type error happens as `schemaUid` is undefined.

For example, given this schema,

```yaml
asyncapi: 2.0.0
id: "https://github.com/yushiomote/hoge"
info:
  title: "Test"
  version: 2.0.0
defaultContentType: application/json
channels:
  hello:
    publish:
      message:
        name: hello
        payload:
          # type: object <- The issue happens if this is missing
          properties:
            data: { type: string }
```

this error happens:

```
/Users/yushi.omote/dev/model-play/modelina/src/processors/AsyncAPIInputProcessor.ts:66
    if (schemaUid.includes('<anonymous-schema')) {
                  ^
TypeError: Cannot read properties of undefined (reading 'includes')
```

This error doesn't happen on `master`. Omitting `type` isn't handled as an error in other tools like asyncapi studio.

image

#### How to Reproduce

See the description above.

#### Expected behavior

The spec should be processed without error as the type is `object`.

YushiOMOTE:
**Description**

This will fix type error when type field is omitted in schemas in AsyncAPI spec

**Related issue(s)**

Resolves #875

**Test results**

```javascript
import { JavaScriptGenerator } from '../../src';
import { parse } from 'yaml'

const generator = new JavaScriptGenerator();
const spec = `
asyncapi: 2.0.0
id: "https://github.com/yushiomote/hoge"
info:
  title: "Test"
  version: 2.0.0
defaultContentType: application/json
channels:
  hello:
    publish:
      message:
        name: hello
        payload:
          # type: object <- we omit type field
          properties:
            v1: { type: string }
            v2: { type: string }
            v3: { type: string }
`;

  const models = await generator.generate(parse(spec));
  for (const model of models) {
    console.log(model.result);
  }
}
generate();
```

```sh
npm run start

> start
> ../../node_modules/.bin/ts-node --cwd ../../ ./examples/$npm_package_config_example_name/index.ts

class AnonymousSchema_1 {
  v1;
  v2;
  v3;

  constructor(input) {
    if (input.hasOwnProperty('v1')) {
      this.v1 = input.v1;
    }
    if (input.hasOwnProperty('v2')) {
      this.v2 = input.v2;
    }
    if (input.hasOwnProperty('v3')) {
      this.v3 = input.v3;
    }
  }

  get v1() { return this.v1; }
  set v1(v1) { this.v1 = v1; }

  get v2() { return this.v2; }
  set v2(v2) { this.v2 = v2; }

  get v3() { return this.v3; }
  set v3(v3) { this.v3 = v3; }
}
```

### DIFF
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -63,7 +63,7 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
     
     let schemaUid = schema.uid();
     //Because the constraint functionality of generators cannot handle -, <, >, we remove them from the id if it's an anonymous schema.
-    if (schemaUid.includes('<anonymous-schema')) {
+    if (typeof schemaUid !== 'undefined' && schemaUid.includes('<anonymous-schema')) {
       schemaUid = schemaUid.replace('<', '').replace(/-/g, '_').replace('>', '');
     }
     


jonaslagoni (changes requested):
We really need to add a test that triggers this error to avoid the issue going forward, do you mind adding a test that uses your input in https://github.com/asyncapi/modelina/blob/next/test/processors/AsyncAPIInputProcessor.spec.ts?

jonaslagoni:
Fixed in https://github.com/asyncapi/modelina/pull/994

PR #889, opened 2022-09-22
test: adds test for AsyncAPI polymorphism

kennethaasan:
**Description**

- adds test for AsyncAPI polymorphism to prove that it's not working according to the https://www.asyncapi.com/docs/reference/specification/v2.4.0 spec

### DIFF
--- a/test/generators/typescript/TypeScriptGenerator.spec.ts
+++ b/test/generators/typescript/TypeScriptGenerator.spec.ts
@@ -453,7 +453,7 @@ ${content}`;
       }
     };
 
-    test('should render 6 models (5 classes and 2 enums)', async () => {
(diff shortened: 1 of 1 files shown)


kennethaasan:
@jonaslagoni, please review. I'm not sure where I should put this test, so feel free to tell me where to move it.

jonaslagoni:
Thats for the contribution @kennethaasan, while this is definitely a great test to have, does adding the test when its not accompanied by a fix make sense? 🤔 Cause we do have it as an issue already if I am not mistaken.

kennethaasan:
@jonaslagoni, true. I will try to have a look at the code base and try to make a fix. This is the main blocker for us to start using Modelina. We're using lots of `allOf` and `oneOf`.

PR #899, opened 2022-09-27
feat: adds support for oneOf and anyOf as UnionModel

### LINKED ISSUE #367: Typescript: Using oneOf leads to creation and use of an anonymous schema rather than use of a union type

kennethaasan:
**Description**

- This MR fixes the #367 issue. `oneOf` is now converted into an UnionModel.
- There is still an issue with properties from `Cat`, `Dog`, and `StickInsect` are merged into all the models. This seems related to merging properties by using `allOf` which is unrelated to this MR. I will make another MR to fix that.
- There is also an issue with the `discriminator` property. I still have to set a `petType` const in every schema that inherits from `Pet` for it to be included in the `PetType` enum. I see that as a separate issue that we can solve with another MR.

**Related issue(s)**
Fixes #367

### DIFF
--- a/examples/generate-python-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-python-models/__snapshots__/index.spec.ts.snap
@@ -1,0 +1,18 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Should be able to render python models and should log expected output to console 1`] = `
+Array [
+  "class Root: 
+  def __init__(self, input):
+    if hasattr(input, 'email'):
+    	self._email = input.email
+
+  @property
+  def email(self):
+  	return self._email
+  @email.setter
+  def email(self, email):
+  	self._email = email
+",
+]
+`;
--- a/examples/rust-generate-crate/__snapshots__/index.spec.ts.snap
+++ b/examples/rust-generate-crate/__snapshots__/index.spec.ts.snap
@@ -52,12 +52,12 @@ Array [
   "// Members represents a union of types: String, f64, bool
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub enum Members {
-    #[serde(rename=\\"Members0\\")]
-    Members0(String),
-    #[serde(rename=\\"Members1\\")]
-    Members1(f64),
-    #[serde(rename=\\"Members2\\")]
-    Members2(bool),
+    #[serde(rename=\\"MembersOneOf_00\\")]
+    MembersOneOf_00(String),
+    #[serde(rename=\\"MembersOneOf_11\\")]
+    MembersOneOf_11(f64),
+    #[serde(rename=\\"MembersOneOf_22\\")]
+    MembersOneOf_22(bool),
 }
 
 ",
--- a/src/helpers/CommonModelToMetaModel.ts
+++ b/src/helpers/CommonModelToMetaModel.ts
@@ -53,11 +53,21 @@ export function convertToMetaModel(jsonSchemaModel: CommonModel, alreadySeenMode
   return new AnyModel(modelName, jsonSchemaModel.originalInput);
 }
 export function convertToUnionModel(jsonSchemaModel: CommonModel, name: string, alreadySeenModels: Map<CommonModel, MetaModel>): UnionModel | undefined {
+  const unionModel = new UnionModel(name, jsonSchemaModel.originalInput, []);
+
+  if (Array.isArray(jsonSchemaModel.union)) {
+    alreadySeenModels.set(jsonSchemaModel, unionModel);
+    for (const unionCommonModel of jsonSchemaModel.union) {
+      const unionMetaModel = convertToMetaModel(unionCommonModel, alreadySeenModels);
+      unionModel.union.push(unionMetaModel);
+    }
+    return unionModel;
+  }
+
   if (!Array.isArray(jsonSchemaModel.type) || jsonSchemaModel.type.length <= 1) {
     return undefined;
   }
   // Has multiple types, so convert to union
-  const unionModel = new UnionModel(name, jsonSchemaModel.originalInput, []);
   alreadySeenModels.set(jsonSchemaModel, unionModel);
   const enumModel = convertToEnumModel(jsonSchemaModel, name);
   if (enumModel !== undefined) {
--- a/src/helpers/CommonModelToMetaModel.ts
+++ b/src/helpers/CommonModelToMetaModel.ts
@@ -53,22 +53,23 @@ export function convertToMetaModel(jsonSchemaModel: CommonModel, alreadySeenMode
   return new AnyModel(modelName, jsonSchemaModel.originalInput);
 }
 export function convertToUnionModel(jsonSchemaModel: CommonModel, name: string, alreadySeenModels: Map<CommonModel, MetaModel>): UnionModel | undefined {
+  const containsUnions = Array.isArray(jsonSchemaModel.union);
+  const containsSimpleTypeUnion = Array.isArray(jsonSchemaModel.type) && jsonSchemaModel.type.length > 1
+  if (!containsSimpleTypeUnion && !containsUnions) {
+    return undefined;
+  }
   const unionModel = new UnionModel(name, jsonSchemaModel.originalInput, []);
-
-  if (Array.isArray(jsonSchemaModel.union)) {
-    alreadySeenModels.set(jsonSchemaModel, unionModel);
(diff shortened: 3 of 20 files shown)


src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -30,8 +30,8 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
     }
     inputModel.originalInput = doc;
     // Go over all the message payloads and convert them to models
-    for (const [, message] of doc.allMessages()) {
-      const schema = AsyncAPIInputProcessor.convertToInternalSchema(message.payload());
+    for (const [messageName, message] of doc.allMessages()) {
+      const schema = AsyncAPIInputProcessor.convertToInternalSchema(message.payload(), messageName);
```
jonaslagoni, line 34: So this change in this file, I kinda understand, but not really, do you mind clarifying why this was introduced? 🤔 

The reason why I am confused about this change is that when calling `schema.uid();` the message name should already have been applied to the payload schema by the parser 🤔

jonaslagoni: This change is why many of the snapshots now all of a sudden expect `LessAnonymousMinusMessageMinus_1Greater` instead of `AnonymousMessage_1`.

src/interpreter/Interpreter.ts
```diff
@@ -78,11 +79,11 @@ export class Interpreter {
     interpretItems(schema, model, this, interpreterOptions);
     interpretProperties(schema, model, this, interpreterOptions);
     interpretAllOf(schema, model, this, interpreterOptions);
+    interpretOneOf(schema, model, this, interpreterOptions);
     interpretDependencies(schema, model, this, interpreterOptions);
     interpretConst(schema, model);
     interpretEnum(schema, model);
 
-    this.interpretAndCombineMultipleSchemas(schema.oneOf, model, schema, interpreterOptions);
     this.interpretAndCombineMultipleSchemas(schema.anyOf, model, schema, interpreterOptions);
```
jonaslagoni, line 87: Even though this PR is for `oneOf` the same behavior can be applied to `anyOf` as they have, from a data model perspective, the same expected behavior.

Dont need to adapt it, but just wanted to raise it 🙂

jonaslagoni (changes requested):

src/helpers/CommonModelToMetaModel.ts
```diff
@@ -53,11 +53,21 @@ export function convertToMetaModel(jsonSchemaModel: CommonModel, alreadySeenMode
...
+      unionModel.union.push(unionMetaModel);
+    }
+    return unionModel;
+  }
+
   if (!Array.isArray(jsonSchemaModel.type) || jsonSchemaModel.type.length <= 1) {
     return undefined;
   }
   // Has multiple types, so convert to union
-  const unionModel = new UnionModel(name, jsonSchemaModel.originalInput, []);
```
jonaslagoni, line 60: I would probably change this a bit to make it more clear to when exactly it converts to unions, what about something like this?:  
```suggestion
  const containsUnions = Array.isArray(jsonSchemaModel.union);
  const containsSimpleTypeUnion = Array.isArray(jsonSchemaModel.type) && jsonSchemaModel.type.length > 1
  if (!containsSimpleTypeUnion && !containsUnions) {
    return undefined;
  }
  const unionModel = new UnionModel(name, jsonSchemaModel.originalInput, []);
  alreadySeenModels.set(jsonSchemaModel, unionModel);
  
  // Has multiple types, so convert to union
  if (containsUnions) {
    for (const unionCommonModel of jsonSchemaModel.union) {
      const unionMetaModel = convertToMetaModel(unionCommonModel, alreadySeenModels);
      unionModel.union.push(unionMetaModel);
    }
    return unionModel;
  } 
  // Has simple union types
```

src/processors/AsyncAPIInputProcessor.ts
kennethaasan: This was the only change I added that's not related to `oneOf`. And also why I stated that I wanted to do other changes in separate MRs. I can remove it, but I added it because it made the name of the union `PetMessage` instead of `Anonymous`. For some reason, the parser doesn't add the name to the message.

src/interpreter/Interpreter.ts
kennethaasan: I can do that in this MR :)

src/processors/AsyncAPIInputProcessor.ts
jonaslagoni: Alright, let me take a look at the parser and figure out where we need to adapt this behavior then. Cause I know we have quite a few issues when it comes to the naming of the data models.

I am 100% with you it should not be `Anonymous`, but I don't think this is how we fix it 🙂

src/helpers/CommonModelToMetaModel.ts
kennethaasan: For sure a better solution. I'll make this change

src/interpreter/Interpreter.ts
kennethaasan: Please check now

examples/generate-python-models/__snapshots__/index.spec.ts.snap
```diff
@@ -0,0 +1,18 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Should be able to render python models and should log expected output to console 1`] = `
+Array [
+  "class Root:
```
jonaslagoni, line 5: I dont get why this is created 😆 Cause nothing of these changes should affect it 🧐 

Oh well, must be a missed when the PR was merged.

examples/rust-generate-crate/__snapshots__/index.spec.ts.snap
```diff
@@ -52,12 +52,12 @@ Array [
   "// Members represents a union of types: String, f64, bool
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub enum Members {
-    #[serde(rename=\\"Members0\\")]
-    Members0(String),
-    #[serde(rename=\\"Members1\\")]
-    Members1(f64),
-    #[serde(rename=\\"Members2\\")]
-    Members2(bool),
+    #[serde(rename=\\"MembersOneOf_00\\")]
```
jonaslagoni, line 55: We might have to take a double check down the line after this PR, https://github.com/asyncapi/modelina/pull/869, and https://github.com/asyncapi/modelina/pull/876 how we can properly name schemas for models.

cc @leigh-johnson

jonaslagoni (approved):
This LGTM 👍 

@magicmatatjahu do you want to give it a review just so I did not miss anything 😄

examples/generate-python-models/__snapshots__/index.spec.ts.snap
kennethaasan: Probably 😄

magicmatatjahu (approved):

Samridhi-98 (approved):

jonaslagoni:
@all-contributors please add @kennethaasan for code, test

PR #900, opened 2022-09-27
feat: upgrades GitHub actions, Node.js, and npm audit fix

kennethaasan:
**Description**

- upgrades Node.js to v16 and https://github.com/actions/setup-node to v3
- upgrades https://github.com/actions/checkout to v3
- I ran npm audit fix


jonaslagoni:
We can unfortunately not update these things as of yet I don't think, because Modelina MUST be compatible with the [AsyncAPI generator](https://github.com/asyncapi/generator#requirements). Basically the same requirement we have across all JS tools.

If I recall correctly we are having discussions about upgrading it to a newer node version, same goes for the workflows, many are inherited from https://github.com/asyncapi/.github/tree/master/.github/workflows. So they should not be changed locally.

So for now this requirement stands until we can upgrade across tools.

kennethaasan:
Makes sense 😄 Just remember that v14 has end of life at 2023-04-30. Closing for now

PR #901, opened 2022-09-27
fix: python generate unusable class for empty properties

jonaslagoni:
**Description**
If the class generator tries to render one with no properties, the class file is unable to compile. 

### DIFF
--- a/src/generators/python/renderers/ClassRenderer.ts
+++ b/src/generators/python/renderers/ClassRenderer.ts
@@ -76,13 +76,20 @@ export const PYTHON_DEFAULT_CLASS_PRESET: ClassPresetType<PythonOptions> = {
   },
   ctor({renderer, model}) {
     const properties = model.properties || {};
-    const assigments = Object.values(properties).map((property) => {
-      if (!property.required) {
-        return `if hasattr(input, '${property.propertyName}'):\n\tself._${property.propertyName} = input.${property.propertyName}`;
-      }
-      return `self._${property.propertyName} = input.${property.propertyName}`;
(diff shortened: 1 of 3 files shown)


src/generators/python/renderers/ClassRenderer.ts
```diff
@@ -76,13 +76,20 @@ export const PYTHON_DEFAULT_CLASS_PRESET: ClassPresetType<PythonOptions> = {
...
+        if (!property.required) {
+          return `if hasattr(input, '${property.propertyName}'):\n\tself._${property.propertyName} = input.${property.propertyName}`;
+        }
+        return `self._${property.propertyName} = input.${property.propertyName}`;
+      });
+      body = renderer.renderBlock(assigments);
+    } else {
+      body = `"""
+No properties
+"""`;
```
magicmatatjahu, line 91: Do we need that comment?

magicmatatjahu (approved):

src/generators/python/renderers/ClassRenderer.ts
jonaslagoni: Not necessarily that comment, but yes, otherwise compilation fails 😅

jonaslagoni: Maybe `pass` is better? 🤔 https://www.programiz.com/python-programming/pass-statement

magicmatatjahu: No, comment is good :) I don't have a problem with it, I only asked.

PR #902, opened 2022-09-27
fix: wrong typescript number type used for integers

jonaslagoni:
**Description**
There does not exist a type called `interger` for typescript numbers.


Samridhi-98 (approved):
LGTM 🚀

PR #903, opened 2022-09-27
fix: duplicate and self dependencies should not being rendered

jonaslagoni:
**Description**
In the current implementation, we sometimes render duplicate and self-dependencies, which makes some black-box tests fail.  This PR fixes this issue.

### DIFF
--- a/src/models/ConstrainedMetaModel.ts
+++ b/src/models/ConstrainedMetaModel.ts
@@ -1,4 +1,3 @@
-import { makeUnique } from '../utils/DependencyHelper';
 import { MetaModel } from './MetaModel';
 
 export abstract class ConstrainedMetaModel extends MetaModel {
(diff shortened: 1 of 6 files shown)


src/models/ConstrainedMetaModel.ts
```diff
@@ -131,20 +132,6 @@ export class ConstrainedEnumModel extends ConstrainedMetaModel {
...
-        return enumModel.value instanceof ConstrainedReferenceModel;
-      }
-    ).map((enumModel) => {
-      return enumModel.value as ConstrainedReferenceModel;
-    });
-
-    //Ensure no duplicate references
-    dependencyModels = [...new Set(dependencyModels)];
-    return dependencyModels;
-  }
```
jonaslagoni, line 147: This is removed because enums are not supposed to have ConstrainedMetaModels i.e. they will never have dependencies, they are pure values.

magicmatatjahu (approved):

PR #905, opened 2022-09-27
fix: solving blackbox tests problems

jonaslagoni:
**Description**
This PR enables and fixes the BlackBox tests.

This PR introduces the following changes:
- Each generator has it's own individual black-box test and script to speed up the process and improve DX. Down the line we should be able to only run one specific test dependant on the change that happens (no need to run black-box test if it's the typescript generator that changed)
- Add a new generic file check option for file generators, that verifies that the files indeed are completely written before returning. This is a use-case that can happen if you swamp the file writer. I.e. as we do with black-box tests, in some cases the files were tested before they were actually written.
- Removes absolute imports and replaces them with relative ones.
- Fixed a bug where a union of models defined within `CommonModel` creates models with the same name.
- Add isDctionary pattern detection because models `{ "type": "object", "additionalProperties": { "$ref": "#" } }` more often than not define a dictionary rather than a full object to be rendered.
- Fix caching of models

**Blocked by:**
- https://github.com/asyncapi/modelina/pull/901 
- https://github.com/asyncapi/modelina/pull/903
- https://github.com/asyncapi/modelina/pull/1000
- https://github.com/asyncapi/modelina/pull/1002


package.json
```diff
@@ -83,7 +84,16 @@
...
     "test:examples:regular": "cross-env CI=true jest ./examples --testPathIgnorePatterns ./examples/integrate-with-react",
     "test:examples:websites": "cd ./examples/integrate-with-react && npm i && npm run test",
-    "test:blackbox": "cross-env CI=true jest ./test/blackbox",
+    "test:blackbox": "concurrently --group -n csharp,go,java,javascript,python,rust,typescript \"npm run test:blackbox:csharp\" \"npm run test:blackbox:go\" \"npm run test:blackbox:java\" \"npm run test:blackbox:javascript\" \"npm run test:blackbox:python\" \"npm run test:blackbox:rust\" \"npm run test:blackbox:typescript\"",
+    "test:blackbox:csharp": "cross-env CI=true jest ./test/blackbox/blackbox-csharp",
+    "test:blackbox:go": "cross-env CI=true jest ./test/blackbox/blackbox-go",
+    "test:blackbox:java": "cross-env CI=true jest ./test/blackbox/blackbox-java",
+    "test:blackbox:javascript": "cross-env CI=true jest ./test/blackbox/blackbox-javascript",
+    "test:blackbox:python": "cross-env CI=true jest ./test/blackbox/blackbox-python",
+    "test:blackbox:rust": "cross-env CI=true jest ./test/blackbox/blackbox-rust",
```
leigh-johnson, line 93: :bowing_woman: Splitting these tests is smart! Thanks for doing this. I'm going to start working on the Rust black-box tests in a branch of `jonaslagoni:feature/fix_all_blackbox_tests`

jonaslagoni: Awesome 🔥 I am gonna mark rust as skipping them for now, and then you can enable them on your own 🙂

jonaslagoni: I just pushed the rest of the current changes I have.

The last issues I found for rust, was something to do with enums and unions, where if an enum model was referenced in a union it expected the model to implement some eq something function I think 🤔

src/generators/dart/DartFileGenerator.ts
```diff
@@ -11,14 +11,15 @@ export class DartFileGenerator extends DartGenerator implements AbstractFileGene
    * @param input
    * @param outputDirectory where you want the models generated to
    * @param options
+   * @param ensureFilesWritten verify that the files is completely written before returning, this can happen if the file system is swamped with write requests. 
    */
-  public async generateToFiles(input: Record<string, unknown> | InputMetaModel, outputDirectory: string, options: DartRenderCompleteModelOptions): Promise<OutputModel[]> {
+  public async generateToFiles(input: Record<string, unknown> | InputMetaModel, outputDirectory: string, options: DartRenderCompleteModelOptions, ensureFilesWritten = false): Promise<OutputModel[]> {
```
magicmatatjahu, line 16: Hmm, maybe that last argument can be a part of options? In this place and another ones? I have only problem with that, rest code is good 👌🏼

jonaslagoni: Yea it's a good point, I actually almost commit that change here 🙃 (have it in my git stash) However I decided to do it in another PR that unifies all option behaviors 🙂

magicmatatjahu: Ok, I will accept but remember about this next PR :D

magicmatatjahu (approved):

src/generators/dart/DartFileGenerator.ts
jonaslagoni: Will do, gonna create an issue first to propose it 🙂

PR #917, opened 2022-10-03
fix: merging CommonModel properties should not carry over properties to other models

kennethaasan:
**Description**

- Please check the Pet test in the MR. This fix makes sure that `CommonModel.mergeProperties` doesn't carry over any properties to other models when trying to merge Pet into multiple schemas.

### DIFF
--- a/src/models/CommonModel.ts
+++ b/src/models/CommonModel.ts
@@ -353,20 +353,25 @@ export class CommonModel {
    * @param alreadyIteratedModels
    */
   private static mergeProperties(mergeTo: CommonModel, mergeFrom: CommonModel, originalInput: any, alreadyIteratedModels: Map<CommonModel, CommonModel> = new Map()) {
-    const mergeToProperties = mergeTo.properties;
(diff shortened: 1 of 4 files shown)


jonaslagoni (changes requested):
I feel like we need a test that fails within the CommonModel tests that your change fixes 🤔 What do you think? 

https://github.com/asyncapi/modelina/blob/0a441c5c5bc29c97e21c7f25ff16068a2a94fb52/test/models/CommonModel.spec.ts#L457

The reason I am saying this is if the generators got split out from the core of Modelina in the future, we would have no tests to ensure that the properties are correctly merged.

Change looks fine though 👍

kennethaasan:
@jonaslagoni, thanks for the quick review! Yes, that makes sense. I will add one now.

kennethaasan:
@jonaslagoni , please check the test I added.

jonaslagoni (approved):
LGTM 👍

jonaslagoni:
I need to fix a problem with the CI, but lets ignore it here 🙂

Issue #916, opened 2022-10-02
Allow overriding interpreters


artur-ciocanu:
#### Reason/Context
Today `modelina` allows different customizations via `preset` mechanism. It works great for most of the cases, however there are situations when customizing the code generator output is impossible, because of some of the hard-coded logic that can be found in `processors` and `interpreter`.
For example, if we want to customize the output of the schema field that have `additionalProperties`, then we really have no choice, but to accept that any field that looks like this:
```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "title": "Test",
  "type": "object",
  "properties": {
    "prop": {
      "type": "object",
      "additionalProperties": false
    }
  },
  "additionalProperties": false,
  "required": [
    "id"
  ]
}
```
Will be translated into something like this (e.g. Java code gen):
```
public class Test {
  private Prop prop;

  public Prop getProp() { return this.prop; }
  public void setProp(Prop prop) { this.prop = prop; }
}

public class Prop {
  
}
```
Which is quite unfortunate since we can do anything useful with this Java class. I would argue that the above schema declaration should generate something like this:
```
public class Test {
  private Object prop;

  @JsonProperty("prop")
  public Object getProp() { return this.prop; }
  public void setProp(Object prop) { this.prop = prop; }
}
```
I would like to mention that if we use a schema like this:
```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "title": "Test",
  "type": "object",
  "properties": {
    "prop": true
  },
  "additionalProperties": false,
  "required": [
    "id"
  ]
}
```
Or
```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "title": "Test",
  "type": "object",
  "properties": {
    "prop": {
       "title": "Prop"
    }
  },
  "additionalProperties": false,
  "required": [
    "id"
  ]
}
```
Then we get a property of type `Object`. As far as I can tell all the 3 schema definitions are equivalent from code generation perspective, however since `InterpretAdditionalProperties` can't be customized we can't do anything about it.

#### Description
One possible solution is to allow clients to pass custom `interpreters` for all the different properties like `oneOf`, `anyOf`, `additionalItems`, `additionalProperties`, etc. This is somewhat similar to the way we can create custom `presets` and add them via config.

artur-ciocanu:
This issue is related to this discussion: https://github.com/asyncapi/modelina/discussions/870.

handrews:
[...]

PR #923, opened 2022-10-04
feat: adds support for interpreting a combination of allOf and oneOf, and properties and oneOf

kennethaasan:
**Description**

- adds support for interpreting a combination of allOf and oneOf
- adds support for interpreting a combination of properties and oneOf

AFAIK, according to the JSON Schema spec, it's possible to combine allOf and oneOf like I did in the test. This PR adds support for interpreting that combination.

I made this a draft PR so that we can discuss the approach. If you agree to this solution I will also add some tests for the `Interpreter`.


test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -488,4 +488,200 @@ ${content}`;
...
+          Pet: {
+            payload: {
+              title: 'Pet',
+              allOf: [
+                { $ref: '#/components/schemas/Animal' },
+              ],
+              oneOf: [
+                { $ref: '#/components/schemas/Cat' },
+                { $ref: '#/components/schemas/Dog' },
+              ],
```
jonaslagoni, line 511:
```json
{
              allOf: [
                { $ref: '#/components/schemas/Animal' },
                {oneOf: [
                  { $ref: '#/components/schemas/Cat' },
                  { $ref: '#/components/schemas/Dog' },
                ]}
              ],
}
```

Is this a pattern you have seen being used?

jonaslagoni (commented):
Well, we are definitely going into the realm of "interpreting" JSON Schema variants here, and there are definitely gonna be people to argue against this interpretation of it, good thing we are pursuing https://github.com/asyncapi/modelina/issues/916.

Back to this proposal... I think it makes sense a long way.

I would probably start determining this as pattern matching where we look for specific patterns of keywords together, to determine how they are to be interpreted. So I would create a new interpreter file per pattern you see and want to process. I.e. you leave the `InterpretOneOf` to just being interpreting regular oneOf instances.

The more complicated these patterns become, the harder it will be to keep the code clean and without introducing weird bugs 😅 

Just to keep it in mind, we can also do pattern matching per schema type, i.e. we only apply this to the AsyncAPI Schema object. Maybe it's relevant.

Furthermore, remember docs here, especially if we start to match specific patterns of keywords together 🙂 https://github.com/asyncapi/modelina/blob/next/docs/inputs/JSON_Schema.md

Otherwise 💪 proposal!

test/generators/typescript/TypeScriptGenerator.spec.ts
kennethaasan: No, just the one in my test.

kennethaasan:
@jonaslagoni, thanks for reviewing this draft proposal. I will try to change my code to have one interpreter per pattern and add tests accordingly. IMO, this functionality should belong to Modelina and not interpreters that we overwrite because this is core functionality of JSON Schema.

kennethaasan:
@jonaslagoni, please review again.

src/interpreter/InterpretAllOf.ts
```diff
@@ -14,7 +14,13 @@ import { isModelObject } from './Utils';
  * @param interpreterOptions to control the interpret process
  */
 export default function interpretAllOf(schema: InterpreterSchemaType, model: CommonModel, interpreter : Interpreter, interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions): void {
-  if (typeof schema === 'boolean' || schema.allOf === undefined) {return;}
+  if (
+    typeof schema === 'boolean' ||
+    schema.allOf === undefined ||
+    schema.oneOf
```
jonaslagoni, line 20: Hmm, I see we have a bit of a problem with detecting when an interpreter should be run or not 🤔Especially if we find more patterns to look for...

While implementing this, did you think about a way we can improve this?

src/interpreter/InterpretOneOfWithAllOf.ts
```diff
@@ -0,0 +1,53 @@
+import { CommonModel } from '../models/CommonModel';
+import { Interpreter, InterpreterOptions, InterpreterSchemaType } from './Interpreter';
+
+/**
+ * Interpreter function for oneOf keyword with allOf.
+ * 
+ * It puts the schema reference into the items field.
```
jonaslagoni, line 7: Do you mind clarifying a bit more in-depth the pattern it matches and how it's interpreted? 

Cause if we don't look at this for the next couple of months, it might be a bit troubling to get an overview 😅

src/interpreter/InterpretOneOfWithProperties.ts
```diff
@@ -0,0 +1,46 @@
+import { CommonModel } from '../models/CommonModel';
+import { Interpreter, InterpreterOptions, InterpreterSchemaType } from './Interpreter';
+
+/**
+ * Interpreter function for oneOf keyword with properties.
+ * 
+ * It puts the schema reference into the items field.
```
jonaslagoni, line 7: Do you mind clarifying a bit more in-depth the pattern it matches and how it's interpreted? 

Cause if we don't look at this for the next couple of months, it might be a bit troubling to get an overview 😅

src/interpreter/InterpretAllOf.ts
jonaslagoni: Maybe the patterns should export something like `isPatternX` that then can be used as `!isPatternX()` to blacklist when interpretations should not happen 🤔 

Not saying we should do this now, just want to hear your opinion 🙂

jonaslagoni (changes requested):
Just a few comments 🙂 Looking good 🔥

src/interpreter/InterpretAllOf.ts
kennethaasan: Hmm, good question. I can't think of any obvious ways to improve this right now. We need to check what schema properties exist in every interpreter anyway, so I'm not sure if `isPatternX` makes sense.

src/interpreter/InterpretOneOfWithAllOf.ts
kennethaasan: Done.

src/interpreter/InterpretAllOf.ts
jonaslagoni: What if the `interpretOneOfWithAllOf` exposed the following:

```ts
export isPatternOneOfWithAllOf = (....) => {
  if (
    typeof schema === 'boolean' ||
    !schema.oneOf ||
    !schema.allOf ||
    schema.properties ||
    interpreterOptions.allowInheritance
  ) { 
    return false;
  }
  return true;
}
```

and then the core allOf and oneOf would use it like the following: 

```ts
  if (
    typeof schema === 'boolean' ||
    schema.allOf === undefined ||
    isPatternOneOfWithAllOf(...)
```

we would of course have some duplicate checks but I think its worth it considering 🤔

jonaslagoni (approved):
LGTM 🔥

PR #926, opened 2022-10-05
fix: javascript should not split out enums

jonaslagoni:
**Description**
In the current JavaScript generator the enums are split out but never rendered because we don't have a way to render them.

The reason why this change is necessary, is because classes still try to import/require the enums, when they simply don't exist. Which means something like the black-box tests fails.

Part of https://github.com/asyncapi/modelina/pull/905


Samridhi-98 (approved):
LGTM 🚀

Issue #922, opened 2022-10-03
Java dictionary type with integer value fails to generate compilable code


jonaslagoni:
#### Describe the bug

In Java `private Map<String, int> additionalProperties;` is not a valid map as the value of the map MUST be reference `Integer` instead.

The reason this is happening is because the contrainer allows it: https://github.com/asyncapi/modelina/blob/7900bf35622f4e3c03236fbdf3a2c5acd6d0031c/src/generators/java/JavaConstrainer.ts#L27

Use the following JSON Schema input:

```json
{
    "type": "object",
    "additionalProperties": {
      "type": "integer",
      "format": "int32"
    }
  }
}
```

jonaslagoni:
We could filter and map out types here: https://github.com/asyncapi/modelina/blob/7900bf35622f4e3c03236fbdf3a2c5acd6d0031c/src/generators/java/JavaConstrainer.ts#L87

In order to make sure the types are never rendered 🤔

PR #929, opened 2022-10-05
fix: java dictionary constrainer gives unusable integer type for map value

jonaslagoni:
**Description**
This PR fixes an issue where the simple `int` type is not allowed for map values.

Fixes https://github.com/asyncapi/modelina/issues/922
Related to #905

### DIFF
--- a/src/generators/java/JavaConstrainer.ts
+++ b/src/generators/java/JavaConstrainer.ts
@@ -86,7 +86,7 @@ export const JavaDefaultTypeMapping: TypeMapping<JavaOptions> = {
   Dictionary ({constrainedModel}): string {
     //Limitations to Java is that maps cannot have specific value types...
     if (constrainedModel.value.type === 'int') {
-      constrainedModel.value.type = 'Integer'
+      constrainedModel.value.type = 'Integer';
     }
     return `Map<${constrainedModel.key.type}, ${constrainedModel.value.type}>`;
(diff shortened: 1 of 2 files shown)


src/generators/java/JavaConstrainer.ts
```diff
@@ -84,6 +84,10 @@ export const JavaDefaultTypeMapping: TypeMapping<JavaOptions> = {
     return 'Object';
   },
   Dictionary ({constrainedModel}): string {
+    //Limitations to Java is that maps cannot have specific value types...
+    if (constrainedModel.value.type === 'int') {
+      constrainedModel.value.type = 'Integer';
```
magicmatatjahu, line 89: Here, you override `constrainedModel` value type. Should we create local variable for such a thing?

jonaslagoni: I would have liked that it was originally constrained in the `ConstrainedIntegerModel` callback, but it does not have access to parent models (maybe it should 🤔) which means this is the only way to do it currently.

It needs to overwrite the type because if any preset access the value model type for the dictionary, as part of some function, it needs to return `Integer`, always.

magicmatatjahu: Thanks for clarification!

magicmatatjahu (approved):

PR #935, opened 2022-10-07
fix: remove unintended characters for typescript marshalling preset

kennethaasan:
**Description**

- fixes renderUnmarshalProperty for typescript


jonaslagoni (approved):
LGTM 👍 

No idea why the workflows don't run...

/rtm

PR #947, opened 2022-10-12
feat!: add dependency manager

### LINKED ISSUE #851: Give type constraint access to renderer
#### Reason/Context
When adding custom types, sometimes you would need a dependency to use it, therefore the constraints should have access to the renderer.

See more https://github.com/asyncapi/modelina/issues/546#issuecomment-1209340325

jonaslagoni:
**Description**
Up until now, dependencies have been part of the renderer instance as it was only the presets and renderers that needed to add dependencies, however, there is a need to let the type constrainer add dependencies. This creates a bit of a dilemma, as the renderer expects a constrained model, and to create a constrained model it needs a renderer...

So therefore I suggest we split out the dependency functionality into its own instance where each language can create its own custom logic for handling dependencies. For example, TypeScript needs specific functions to more efficiently make imports based on the options from the user.

I would love to have alternatives, and if you have something please let me know, cause it does create somewhat of a complexity 🙃 But maybe this is for version 2? 🤷 

Alternatives I considered:
- We could use something like a singleton pattern instead of passing the instance across the entire flow, but it's not possible because each rendering needs its own set of dependencies.
- Redoing the renderer + constrained model dependency so they would no longer be dependent on each other and we could then pass the renderer to the container. This just seems more complex to do than this...
- ??

Changes introduced:
- Introduce the concept of dependency managers which you have access to in constraints and presets.

**Related issue(s)**
Fixes #851


magicmatatjahu (commented):
For me these solution can be, but the `AbstractDependencyManager` itself should have a method that renders all dependencies as output like `renderDependencies()`. In addition, a better solution would be if we could pass the dependency manager class as generator option (definition of it) and create it during rendering - currently someone can pass this via an argument to the method, and it should be passed to the constructor option.

jonaslagoni:
> For me these solution can be, but the AbstractDependencyManager itself should have a method that renders all dependencies as output like renderDependencies()

I think that makes sense, let me try to look into that!

> In addition, a better solution would be if we could pass the dependency manager class as generator option (definition of it) and create it during rendering - currently someone can pass this via an argument to the method, and it should be passed to the constructor option.

I don't quite understand this suggestion, can you clarify it?

magicmatatjahu:
@jonaslagoni 

> I don't quite understand this suggestion, can you clarify it?

For example. you have signature:

```ts
public abstract render(model: MetaModel, inputModel: InputMetaModel, dependencyManager?: AbstractDependencyManager): Promise<RenderOutput>;
```

I think that's would be better if people would pass `AbstractDependencyManager` definition to the `Generator` class, not to the function signature, so `CommonGeneratorOptions` should have this new field:

```ts
{
  dependencyManager?: typeof AbstractDependencyManager // definition of class not instance!
}
```

and then if someone will run `render` function, we will create `AbstractDependencyManager` instance based on `options.dependencyManager` like:

```
render(...) {
  const dependencyManager = new options.dependencyManager(...)
}
```

Is it clarify? :D

jonaslagoni:
> Is it clarify? :D

Yes! Actually, that's brilliant!

Let me try out that approach 😄

src/generators/typescript/TypeScriptDependencyManager.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Render the model dependencies based on the options
+   * 
+   * @param model 
+   * @param opt 
+   * @returns 
+   */
+  renderCompleteModelDependencies(model: ConstrainedMetaModel, opt: TypeScriptRenderCompleteModelOptions) {
```
magicmatatjahu, line 39: That function should be included in the `AbstractDependencyManager` as abstract method and every manager should implement it.

src/generators/typescript/TypeScriptGenerator.ts
```diff
@@ -41,13 +42,19 @@ export class TypeScriptGenerator extends AbstractGenerator<TypeScriptOptions,Typ
...
+  
+  static defaultCompleteModelOptions: TypeScriptRenderCompleteModelOptions = {
+    exportType: 'default'
   };
 
   constructor(
     options?: DeepPartial<TypeScriptOptions>,
   ) {
     const realizedOptions = mergePartialAndDefault(TypeScriptGenerator.defaultOptions, options) as TypeScriptOptions;
+    realizedOptions.dependencyManagerFactory = () => { return new TypeScriptDependencyManager(realizedOptions);};
```
magicmatatjahu, line 57: Here you override `dependencyManagerFactory` field and people should have option to pass their manager.

src/generators/AbstractGenerator.ts
```diff
@@ -23,16 +26,16 @@ export const defaultGeneratorOptions: CommonGeneratorOptions = {
...
     public readonly languageName: string,
     public readonly options: Options
   ) { }
 
-  public abstract render(model: MetaModel, inputModel: InputMetaModel): Promise<RenderOutput>;
-  public abstract renderCompleteModel(model: MetaModel, inputModel: InputMetaModel, options: RenderCompleteModelOptions): Promise<RenderOutput>;
-  public abstract constrainToMetaModel(model: MetaModel): ConstrainedMetaModel;
+  public abstract render(model: MetaModel, inputModel: InputMetaModel, dependencyManager?: AbstractDependencyManager): Promise<RenderOutput>;
+  public abstract renderCompleteModel(model: MetaModel, inputModel: InputMetaModel, options: Partial<RenderCompleteModelOptions>, dependencyManager?: AbstractDependencyManager): Promise<RenderOutput>;
+  public abstract constrainToMetaModel(model: MetaModel, dependencyManager: AbstractDependencyManager): ConstrainedMetaModel;
```
magicmatatjahu, line 38: If we have `AbstractDependencyManager` in the generator option, why we still need that in arguments?

magicmatatjahu (commented):
Few comments :)

src/generators/typescript/TypeScriptGenerator.ts
jonaslagoni: Right... Good catch!

src/generators/typescript/TypeScriptDependencyManager.ts
jonaslagoni: I will add it if it makes sense 👍

src/generators/AbstractDependencyManager.ts
```diff
@@ -0,0 +1,16 @@
+export abstract class AbstractDependencyManager {
+  constructor(
+    public dependencies: string[] = []
+  ) {}
```
magicmatatjahu, line 4: Each dependency manager should have access to the generator options, by this you can avoid passing option to the `TypeScriptDependencyManager` methods.

magicmatatjahu (changes requested):
One thing, we discussed about it as I remember 😅

src/generators/AbstractDependencyManager.ts
jonaslagoni: I tried to avoid it for the complexity of generics and because providing the options on creation makes it hard to define a default value in the same manner we currently use. 

Do you think it's necessary?

jonaslagoni: Okay added the options as I saw it was an annoying experience for the user to keep track of it when adding dependencies 😄

src/generators/typescript/TypeScriptGenerator.ts
```diff
@@ -41,16 +44,41 @@ export class TypeScriptGenerator extends AbstractGenerator<TypeScriptOptions, Ty
     defaultPreset: TS_DEFAULT_PRESET,
     typeMapping: TypeScriptDefaultTypeMapping,
     constraints: TypeScriptDefaultConstraints,
-    moduleSystem: 'ESM'
+    moduleSystem: 'ESM',
+    //Temporarily sat
```
magicmatatjahu, line 48:
```suggestion
    // Temporarily set
```

magicmatatjahu (commented):
I like it. A couple of things I don't like is that you add too many things through class arguments and not through some single object, such as here https://github.com/asyncapi/modelina/pull/947/files#diff-a6a5c62c72e934883df999f53c7335c2b386754b04ca5222520015351537d546R20 but it's ok 👌🏼

jonaslagoni:
@magicmatatjahu lets merge it into a separate branch to allow each generator and maintainer to review the changes more easily.

magicmatatjahu (approved):
@jonaslagoni Yeah, good idea. If new feature has a bugs that I don't see, it will be easier to fix :)

PR #958, opened 2022-10-18
feat: handle const in properties

### LINKED ISSUE #946: Properties that use const should be initialized with that const value and not be included in the constructor of a model
#### Description
Properties that use const should be initialized with that const value and not be included in the constructor of a model. 

Example: When we set const to Cat in https://github.com/asyncapi/modelina/blob/next/test/generators/typescript/TypeScriptGenerator.spec.ts#L401, the generated class for Cat should automatically set the type property to PetType.CAT

https://github.com/asyncapi/modelina/blob/next/test/generators/typescript/__snapshots__/TypeScriptGenerator.spec.ts.snap#L7 should be:
```
class Cat {
  private _petType: PetType = PetType.CAT;
  private _reservedName: string;
  private _huntingSkill: AnonymousSchema_6;

  constructor(input: {
    petType: PetType, // this line should be deleted because you should not be able to set petType to something other than Cat
    reservedName: string,
    huntingSkill: AnonymousSchema_6,
  }) {
    this._petType = input.petType;
    this._reservedName = input.reservedName;
    this._huntingSkill = input.huntingSkill;
  }

  get petType(): PetType { return this._petType; }
  set petType(petType: PetType) { this._petType = petType; } // this is not needed because you should not be able to set petType to something other than Cat

  get reservedName(): string { return this._reservedName; }
  set reservedName(reservedName: string) { this._reservedName = reservedName; }

  get huntingSkill(): AnonymousSchema_6 { return this._huntingSkill; }
  set huntingSkill(huntingSkill: AnonymousSchema_6) { this._huntingSkill = huntingSkill; }
}
```

kennethaasan:
**Description**

- handle const in properties

**Related issue(s)**
Resolves #946

### DIFF
--- a/src/generators/AbstractRenderer.ts
+++ b/src/generators/AbstractRenderer.ts
@@ -1,5 +1,5 @@
 import { AbstractGenerator, CommonGeneratorOptions } from './AbstractGenerator';
-import { ConstrainedMetaModel, InputMetaModel, Preset } from '../models';
+import { ConstrainedEnumModel, ConstrainedEnumValueModel, ConstrainedMetaModel, ConstrainedObjectPropertyModel, ConstrainedReferenceModel, InputMetaModel, Preset } from '../models';
 import { FormatHelpers, IndentationTypes } from '../helpers';
 
 /**
@@ -80,5 +80,16 @@ export abstract class AbstractRenderer<
       }
     }
     return content;
+  }
+
+  getConstValue(property: ConstrainedObjectPropertyModel): ConstrainedEnumValueModel | undefined {
+    if (
+      property.property instanceof ConstrainedReferenceModel &&
+      property.property.ref instanceof ConstrainedEnumModel &&
+      property.property.ref.constValue
+    ) {
+      return property.property.ref.constValue;
+    }
+    return undefined;
   }
 }
--- a/src/generators/java/renderers/ClassRenderer.ts
+++ b/src/generators/java/renderers/ClassRenderer.ts
@@ -82,14 +82,23 @@ export const JAVA_DEFAULT_CLASS_PRESET: ClassPresetType<JavaOptions> = {
   self({ renderer }) {
     return renderer.defaultSelf();
   },
-  property({ property }) {
+  property({ renderer, property }) {
+    const constValue = renderer.getConstValue(property);
+
+    if (constValue) {
+      return `private final ${property.property.type} ${property.propertyName} = ${property.property.type}.${constValue.key};`;
+    }
+
     return `private ${property.property.type} ${property.propertyName};`;
   },
   getter({ property }) {
     const getterName = `get${FormatHelpers.toPascalCase(property.propertyName)}`;
     return `public ${property.property.type} ${getterName}() { return this.${property.propertyName}; }`;
   },
-  setter({ property }) {
+  setter({ renderer, property }) {
+    if (renderer.getConstValue(property)) {
+      return undefined;
+    }
     const setterName = FormatHelpers.toPascalCase(property.propertyName);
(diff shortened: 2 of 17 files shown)


jonaslagoni (commented):
Well, this is definitely a bit of a challenge 😄

Let me ponder this over the next couple of days, this is quite tricky! 😄

kennethaasan:
@jonaslagoni , yes 😄 At least we have a start and something that's working for us.

src/generators/AbstractRenderer.ts
```diff
@@ -81,4 +81,15 @@ export abstract class AbstractRenderer<
     }
     return content;
   }
+
+  getConstValue(property: ConstrainedObjectPropertyModel): ConstrainedEnumValueModel | undefined {
```
jonaslagoni, line 85: Really not a fan that we have to use support functions like here to get the constant value, but on the other hand, I don't see a way around it, as it's not really possible to have a `contantValue` property in the `MetaModel` class or `ConstrainedObjectPropertyModel` ...

src/generators/java/renderers/ClassRenderer.ts
```diff
@@ -82,14 +82,23 @@ export const JAVA_DEFAULT_CLASS_PRESET: ClassPresetType<JavaOptions> = {
   self({ renderer }) {
     return renderer.defaultSelf();
   },
-  property({ property }) {
+  property({ renderer, property }) {
+    const constValue = renderer.getConstValue(property);
+
+    if (constValue) {
+      return `private final ${property.property.type} ${property.propertyName} = ${property.property.type}.${constValue.key};`;
+    }
```
jonaslagoni, line 90: This assumes that constants are only from an enum, which definitely is not the case in the future as all models can have default values (I think).

jonaslagoni: Maybe we should move this as part of the renderer so we can just call `${renderer.getConstantValue(property)}` instead of `${property.property.type}.${constValue.key}`.

src/generators/java/renderers/ClassRenderer.ts
```diff
@@ -82,14 +82,23 @@ export const JAVA_DEFAULT_CLASS_PRESET: ClassPresetType<JavaOptions> = {
...
+
     return `private ${property.property.type} ${property.propertyName};`;
   },
   getter({ property }) {
     const getterName = `get${FormatHelpers.toPascalCase(property.propertyName)}`;
     return `public ${property.property.type} ${getterName}() { return this.${property.propertyName}; }`;
   },
-  setter({ property }) {
+  setter({ renderer, property }) {
+    if (renderer.getConstValue(property)) {
```
jonaslagoni, line 99: The better structure would have been `property.hasConstantValue()` or something similar, but I dont see a way around it, can you? 🤔

src/generators/typescript/TypeScriptObjectRenderer.ts
```diff
@@ -34,7 +34,15 @@ export abstract class TypeScriptObjectRenderer extends TypeScriptRenderer<Constr
   }
 
   renderProperty(property: ConstrainedObjectPropertyModel): string {
-    return `${property.propertyName}${property.required === false ? '?' : ''}: ${property.property.type};`;
+    const renderedProperty = `${property.propertyName}${property.required === false ? '?' : ''}: ${property.property.type}`;
+
+    const constValue = this.getConstValue(property);
+
+    if (constValue) {
+      return `${renderedProperty} = ${property.property.type}.${constValue.key};`;
```
jonaslagoni, line 42: Same as above as this only works for enums and is not future proof.

jonaslagoni (commented):
Sorry for the super long time between review 🙃

Got a few conversation starters where I would love to get your input on!

jonaslagoni:

PR #964, opened 2022-10-21
feat: add basic pydantic support for python models

### LINKED ISSUE #963: Python: add support for Pydantic models
#### Reason/Context
At the moment, Pydantic is a widely used library to work with contract model classes in python.

#### Description
- Every model class should inherit from `pydantic.BaseModel`
- Model fields should be annotated with types and assigned with `pydantic.Field`

zaytsevand:
**Description**

- Classes are inherited from `BaseModel`
- Fields are annotated

**Related issue(s)**
Resolves #963

### DIFF
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -39,6 +39,10 @@ Modelina uses something called **presets** to extend the rendered model. You can
     + [Dart](#dart)
       - [**Class**](#class-4)
       - [**Enum**](#enum-4)
+    + [Python](#python)
+      - [**Class: plain Python**](#class-5)
+      - [**Class: pydantic**](#class-6)
+      - [**Enum**](#enum-5)
 - [Limitations](#limitations)
   * [Hard for two presets to write to the exact same location within a class](#hard-for-two-presets-to-write-to-the-exact-same-location-within-a-class)
 
@@ -493,6 +497,24 @@ This preset is a generator for the meta model `ConstrainedEnumModel` and [can be
 |---|---|---|
 | `item` | A method to extend enum's item. | `item` object as a [`ConstrainedEnumValueModel`](./internal-model.md#the-constrained-meta-model) instance, which contains the value and key of enum's item. |
 
+### Python
+
+#### **Class**
+This preset is a generator for the meta model `ConstrainedObjectModel` and [can be accessed through the `model` argument](#presets-shape).
+
+| Method | Description | Additional arguments |
+|---|---|---|
+| `ctor` | A method to extend rendered constructor for a given class. | - |
+| `property` | A method to extend rendered given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
+| `setter` | A method to extend setter for a given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
+| `getter` | A method to extend getter for a given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
+#### **Enum**
+
+This preset is a generator for the meta model `ConstrainedEnumModel` and [can be accessed through the `model` argument](#presets-shape).
+
+| Method | Description | Additional arguments |
+|---|---|---|
+| `item` | A method to extend enum's item. | `item` object as a [`ConstrainedEnumValueModel`](./internal-model.md#the-constrained-meta-model) instance, which contains the value and key of enum's item. |
 # Limitations
 
 With features natually comes limitations, and same applies for presets, so here are the known limitations the architecture of presets for Modelina.
--- a/examples/generate-python-pydantic-models/index.spec.ts
+++ b/examples/generate-python-pydantic-models/index.spec.ts
@@ -7,8 +7,9 @@ describe('Should be able to render python models', () => {
   });
   test('and should log expected output to console', async () => {
     await generate();
-    //Generate is called 2x, so even though we expect 1 model, we double it
-    expect(spy.mock.calls.length).toEqual(2);
-    expect(spy.mock.calls[1]).toMatchSnapshot();
+    //Generate is called 2x, so even though we expect 2 models, we double it
+    expect(spy.mock.calls.length).toEqual(4);
+    expect(spy.mock.calls[0]).toMatchSnapshot('class-model');
+    expect(spy.mock.calls[1]).toMatchSnapshot('enum-model');
   });
 });
--- a/examples/generate-python-pydantic-models/index.spec.ts
+++ b/examples/generate-python-pydantic-models/index.spec.ts
@@ -7,8 +7,7 @@ describe('Should be able to render python models', () => {
   });
   test('and should log expected output to console', async () => {
     await generate();
-    //Generate is called 2x, so even though we expect 2 models, we double it
-    expect(spy.mock.calls.length).toEqual(4);
+    expect(spy.mock.calls.length).toEqual(2);
     expect(spy.mock.calls[0]).toMatchSnapshot('class-model');
     expect(spy.mock.calls[1]).toMatchSnapshot('enum-model');
   });
--- a/examples/generate-python-pydantic-models/index.spec.ts
+++ b/examples/generate-python-pydantic-models/index.spec.ts
@@ -1,0 +1,14 @@
+const spy = jest.spyOn(global.console, 'log').mockImplementation(() => { return; });
+import {generate} from './index';
+
+describe('Should be able to render python models', () => {
+  afterAll(() => {
+    jest.restoreAllMocks();
+  });
+  test('and should log expected output to console', async () => {
+    await generate();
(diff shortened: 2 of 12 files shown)


examples/generate-python-pydantic-models/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    // requiredField: {
+    //   type: 'string',
+    //   format: 'email',
+    //   required: true,
+    //   description: 'this field is required'
+    // },
+    noDescription: {
+      type: 'string'
+    },
+  }
```
jonaslagoni, line 26: Damn, this is some old stuff 😆 We have to go back to [draft-02](https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-02#section-5.4) for this one.

By default, all properties are optional, to define a required field you have to define it in the required field for draft-7.

I suggest we change it to the below schema 👇 

```suggestion
  $schema: 'http://json-schema.org/draft-07/schema#',
  type: 'object',
  additionalProperties: false,
  required: ["requiredField"],
  properties: {
    optionalField: {
      type: 'string',
      format: 'email',
      description: 'this field is optional'
    },
    requiredField: {
      type: 'string',
      format: 'email',
      description: 'this field is required'
    },
    noDescription: {
      type: 'string'
    },
  }
```

src/generators/python/index.ts
```diff
@@ -18,3 +18,7 @@ export {
   DefaultPropertyKeyConstraints as PythonDefaultPropertyKeyConstraints,
   defaultPropertyKeyConstraints as pythonDefaultPropertyKeyConstraints
 } from './constrainer/PropertyKeyConstrainer';
+
+export {
+  PYTHON_PYDANTIC_PRESET
+} from  './presets/Pydantic';
```
jonaslagoni, line 24: Add this to the `src/generators/python/presets/index.ts` file instead 🙂 Then it's automatically exported.

src/generators/python/presets/Pydantic.ts
```diff
@@ -0,0 +1,32 @@
...
+    return `${params.property.propertyName}: ${type} = Field(${alias})`;
+  },
+  ctor: () => '',
+  getter: () => '',
+  setter: () => ''
+};
+
+export const PYTHON_PYDANTIC_PRESET: PythonPreset<PythonOptions> = {
+  class: PYTHON_PYDANTIC_CLASS_PRESET,
+  enum: PYTHON_DEFAULT_ENUM_PRESET,
```
jonaslagoni, line 31: `PYTHON_DEFAULT_ENUM_PRESET` is part of the default presets, which means it is already rendered 🙂

```suggestion
```

jonaslagoni (changes requested):
Thanks for taking the time to provide this PR @zaytsevand!

Just a few suggestions for improving the code, and on a general note do you mind adding the following so people can know the feature exists:
- Add your new preset for the python documentation so people know it exists: https://github.com/asyncapi/modelina/blob/next/docs/languages/Python.md, you can use the [TypeScript docs as guidance](https://github.com/asyncapi/modelina/blob/next/docs/languages/TypeScript.md)
- Add the example to the list of examples: https://github.com/asyncapi/modelina/tree/next/examples

src/generators/python/presets/Pydantic.ts
```diff
@@ -0,0 +1,32 @@
...
+      : `Optional[${params.property.property.type}]`;
+    const alias = params.property.property.originalInput['description']
+      ? `alias='${params.property.property.originalInput['description']}'`
+      : '';
+
+    return `${params.property.propertyName}: ${type} = Field(${alias})`;
+  },
+  ctor: () => '',
+  getter: () => '',
+  setter: () => ''
```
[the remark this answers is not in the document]
zaytsevand: Am I doing something else wrong? I've tried not to override those presets, but I'm getting default a codegen executed too:
```
$ npm run start   

> start
> ../../node_modules/.bin/ts-node --cwd ../../ ./examples/$npm_package_config_example_name/index.ts

class Root(BaseModel): 
  optionalField: Optional[str] = Field(alias='this field is optional')
  requiredField: str = Field(alias='this field is required')
  noDescription: Optional[str] = Field()
  options: Optional[Options] = Field()

  def __init__(self, input):
    if hasattr(input, 'optionalField'):
        self._optionalField = input.optionalField
    self._requiredField = input.requiredField
    if hasattr(input, 'noDescription'):
        self._noDescription = input.noDescription
    if hasattr(input, 'options'):
        self._options = input.options

  @property
  def optionalField(self):
        return self._optionalField
  @optionalField.setter
  def optionalField(self, optionalField):
        self._optionalField = optionalField

  @property
  def requiredField(self):
        return self._requiredField
  @requiredField.setter
  def requiredField(self, requiredField):
        self._requiredField = requiredField

  @property
  def noDescription(self):
        return self._noDescription
  @noDescription.setter
  def noDescription(self, noDescription):
        self._noDescription = noDescription

  @property
  def options(self):
        return self._options
  @options.setter
  def options(self, options):
        self._options = options

class Options(Enum): 
  NUMBER_123 = 123
  NUMBER_213 = 213
  RESERVED_TRUE = "true"
  RUN = "Run"```

File: examples/generate-python-pydantic-models/index.ts (hunk 1 above)
Comment (zaytsevand) [reply examples/generate-python-pydantic-models/index.ts:26]: Thanks for the comment, I missed that

File: src/generators/python/presets/Pydantic.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (jonaslagoni) [reply src/generators/python/presets/Pydantic.ts:26]: Ahhh! You are completely disabling them! 🤦 

What's the reason behind wanting the preset to disable the getter and setter functions altogether? (I don't know much about Pydanitc and Python in general, so assume I know nothing 😆)

File: src/generators/python/index.ts (hunk 2 above)
Comment (zaytsevand) [reply src/generators/python/index.ts:24]: done

File: src/generators/python/presets/Pydantic.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (jonaslagoni) [reply src/generators/python/presets/Pydantic.ts:26]: Is it because otherwise, the Pydantic annotation does not take effect because users interact over the accessors?

File: src/generators/python/presets/Pydantic.ts (hunk 3 above)
Comment (zaytsevand) [reply src/generators/python/presets/Pydantic.ts]: done

File: src/generators/python/presets/Pydantic.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (zaytsevand) [reply src/generators/python/presets/Pydantic.ts:25]: They're currently not supporting them, for instance: https://github.com/pydantic/pydantic/issues/3395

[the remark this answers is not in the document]
Comment (zaytsevand) [reply src/generators/python/presets/Pydantic.ts:25]: If it's ok, I would leave it as it is :)

File: src/generators/python/presets/Pydantic.ts (hunk 3 above)
Comment (zaytsevand) [reply src/generators/python/presets/Pydantic.ts]: I added a test to cover proper enum generation. Looks fine :)

**Review by jonaslagoni — CHANGES_REQUESTED** [review]
Remember to [add the docs](https://github.com/asyncapi/modelina/pull/964#pullrequestreview-1150833216) and other then these two simplifications of the test, then I think we are good to merge 👍

File: docs/presets.md [hunk 5]
```diff
@@ -493,6 +497,24 @@ This preset is a generator for the meta model `ConstrainedEnumModel` and [can be
 |---|---|---|
 | `item` | A method to extend enum's item. | `item` object as a [`ConstrainedEnumValueModel`](./internal-model.md#the-constrained-meta-model) instance, which contains the value and key of enum's item. |
 
+### Python
+
+#### **Class**
+This preset is a generator for the meta model `ConstrainedObjectModel` and [can be accessed through the `model` argument](#presets-shape).
```
jonaslagoni, line 503: Nice catch, thanks!

examples/generate-python-pydantic-models/index.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+import {generate} from './index';
+
+describe('Should be able to render python models', () => {
+  afterAll(() => {
+    jest.restoreAllMocks();
+  });
+  test('and should log expected output to console', async () => {
+    await generate();
+    //Generate is called 2x, so even though we expect 2 models, we double it
+    expect(spy.mock.calls.length).toEqual(4);
```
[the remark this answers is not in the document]
zaytsevand: I've two models in output now: one for a class and another for an enum. So tests are failing with 4. Based on the comment above that line, I assume it's expected behavior. Am I getting it right?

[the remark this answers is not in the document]
zaytsevand: fixed

docs/presets.md
zaytsevand: plus I updated:
- examples readme
- Python language readme

zaytsevand:
Not sure what I should do with black-box testing

jonaslagoni:
> Not sure what I should do about black-box test failure

Ignore it, it's part of #905

jonaslagoni (approved):
LGTM 👍 Nice addition!

jonaslagoni:
@all-contributors please add @zaytsevand for code, example, doc, test

PR #969, opened 2022-10-23
fix: csharp generate not null properties

### LINKED ISSUE #941: C# generator does not render optional properties correctly
#### Describe the bug

zaytsevand:
**Description**
- Update the CSharp generator to render nullable properties (CSharp v8+ style) properly
- Test cases are added/adjusted to demonstrate the functionality

**Related issue(s)**
Fixes #941

### DIFF
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -418,6 +418,7 @@ This preset is a generator for the meta model `ConstrainedObjectModel` and [can 
 | `accessor` | A method to extend rendered given property accessor. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
 | `setter` | A method to extend setter for a given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
 | `getter` | A method to extend getter for a given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
+| `propertyType` | A method to extend type definition of a given property. | `property` object as a [`ConstrainedObjectPropertyModel`](./internal-model.md#the-constrained-meta-model) instance. |
 
 #### **Enum**
 
--- a/examples/csharp-auto-implemented-properties/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-auto-implemented-properties/__snapshots__/index.spec.ts.snap
@@ -4,9 +4,12 @@ exports[`Should be able to render auto-implemented properties in CSharp and shou
 Array [
   "public class Root
 {
-  public dynamic[]? Emails { get; set; }
-  public string? StringProp { get; set; }
-  public double? NumberProp { get; set; }
+  public dynamic[]? NullArray { get; set; }
+  public string? NullString { get; set; }
+  public double? NullNumber { get; set; }
+  public dynamic[] NonNullArray { get; set; }
+  public string NonNullString { get; set; }
+  public double NonNullNumber { get; set; }
 }",
 ]
 `;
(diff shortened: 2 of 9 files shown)


jonaslagoni (commented):
I have a bit of an issue with this change, and I am not sure how to proceed 🤔 Hope you can help @zaytsevand.

First, should the type of the model (`model.type`) contain the conditional type `?` 🤔 Or is there cases where you would like to only have non-conditional type? For example for method arguments or other cases?

Is optional (non-required properties) the same as nullable types i.e. for inputs such as `type: ['string', 'null']`?

zaytsevand:
> I have a bit of an issue with this change, and I am not sure how to proceed 🤔 Hope you can help @zaytsevand.
> 
> First, should the type of the model (`model.type`) contain the conditional type `?` 🤔 Or is there cases where you would like to only have non-conditional type? For example for method arguments or other cases?
> 
> Is optional (non-required properties) the same as nullable types i.e. for inputs such as `type: ['string', 'null']`?

Generally, for `?` we have two cases:
- nullable value types, like `bool?`, which is just a shortcut for the `Nullable<bool>` generic type. If that were the only case, `model.type` would have been a perfect solution since full type specification should definitely contain `?` in that case
- nullable reference types. That's a pickle... Depending on a project configuration and language version, reference types could be explicitly set to allow/disallow nulls. So we have things like `object?` type, which makes sense only if our project uses `?` syntax to guard against null references. Otherwise, just a plain `object` would be equivalent to an optional property.

Currently, all new projects created for dotnet v6 (release date November 8, 2021) and c# v8+ are, by default, configured to allow nullable reference types. So I guess if we're not striving to support "legacy" projects, we could go for `?` embedded in `model.type`.

zaytsevand:
@jonaslagoni I checked `TypeMapping`, and it seems that at the time when type is determined, we can access `Options` (which is good, we can introduce a C#-specific option that would determine supported language and framework versions). However, we still need access to `.required` field for a property.

jonaslagoni:
> Currently, all new projects created for dotnet v6 (release date November 8, 2021) and c# v8+ are, by default, configured to allow nullable reference types. So I guess if we're not striving to support "legacy" projects, we could go for ? embedded in model.type.

If it's easy to support legacy projects by a simple change in syntax, I would say we go for that, unless we explicitly wants to add something like a `csharpVersion`/`dotnetVersion` to switch between the syntaxes. Not sure if that would make sense, or just introduce too much complexity 🤔 

> @jonaslagoni I checked TypeMapping, and it seems that at the time when type is determined, we can access Options (which is good, we can introduce a C#-specific option that would determine supported language and framework versions). However, we still need access to .required field for a property.

In #947, it is actually being [redesigned to give you access to whether a type is being rendered for a property through `partOfProperty`](https://github.com/asyncapi/modelina/pull/947/files#diff-e87514fb218573a60d7f3bf14915df83dd40b5b346ca45a2b31b962b60620438R5) so you can have access to whether the property is required or not.

My question is, if the type for the property is normally `string` and we all of a sudden change it to be optional `string?`, would there be cases where you would much rather have access to `string` and not `string?` in a method signature or something similar? 🤔

zaytsevand:
@jonaslagoni I don't think that it would be an issue if we implement the addition of '?' in TypeMap. If somebody will need a plain `string` type instead of `string?` it's just a matter of replacing the standard TypeMap with a custom one.

zaytsevand:
After some code examination: when creating a model for a property, we've no outer context, but it's the outer context where we have that array of all required properties for an outer model. So a simple fix would be to embed a list of all required properties in `ConstrainContext`, but it will result in errors when rendering nested models since the global context will not be changed.
I can't see how I can fix the issue, unfortunately.

jonaslagoni:
Let me push forward https://github.com/asyncapi/modelina/pull/947 as it will give you access to whether a property is required or not when constraining the type.

I don't think you would need access to the entire object model, do you? 🤔

zaytsevand:
@jonaslagoni 💯 correct: the whole model is not needed for it

zaytsevand:
@jonaslagoni could you please take another look at this one? #947 is not here yet, but rendering is broken in the @next branch, making it useless, unfortunately.

PR #991, opened 2022-11-15
docs: add versioning and maintenance section

jonaslagoni:
**Description**
As version 1 creeps closer, I think it's time to address how we can keep the version strategy for the library.

My main reason for adding this section is once you incorporate this library into your ecosystem, you need to know when you can comfortably update the dependency. If you integrate Modelina into an AsyncAPI template (or any other code generator) you need to trust that updating Modelina doesn't break the integration of those who interact with the models.

This section outlines exactly what we are allowed to do and when it requires a breaking change 🙂 

@ron-debajyoti @Samridhi-98 @magicmatatjahu @leigh-johnson I would love to get your input here if you are able to, as it creates constraints on what you can accept as changes and must be aware of.


README.md
```diff
@@ -144,12 +138,26 @@ Do you need to know how to use the library in certain scenarios?
 
 We have gathered all the examples in a separate folder and they can be found under the [examples folder](./examples). 
 
+## Versioning and maintenance
+As of version 1, Modelina has a very strict set of changes we are allowed to do before it requires a major version change. In short, any changes that change the generated outcome are not allowed as it's a breaking change for the consumer of the generated models. 
+
+Here is a list of changes we are allowed to do that would not require a breaking change:
+- Adding new features (that do not change existing output), such as generators, presets, input processors, etc.
+- Change existing features, by providing options that default to current behavior.
```
magicmatatjahu, line 146: I guess this is breaking change or I don't quite understand this sentence, can it be changed?

README.md
```diff
@@ -144,12 +138,26 @@ Do you need to know how to use the library in certain scenarios?
...
+- Bug fixes where the generated code is otherwise unusable (syntax errors, etc).
+
+Breaking changes are allowed and expected at a frequent rate, of course where it makes sense we will try to bundle multiple changes together.
+
+This is what we strive to do, however, mistakes may happen, and please do let us know when they do!
```
magicmatatjahu, line 151:
```suggestion
While we do our best to do so, mistakes can happen and please let us know!
```

README.md
magicmatatjahu: Ok, fine, however I wonder if we should treat `Change existing features, by providing options that default to current behavior.` as breaking change, because it exactly means, that changed function will change output using changed options - and this can also be understood as changing more than just presets or generators etc.

README.md
magicmatatjahu: Ok :)

README.md
jonaslagoni: But where is the breaking change? 🤔

Because the API of modelina does not have any breaking changes, because it defaults to current behaviour, and the generated models does not contain any breaking functionality 🤔

magicmatatjahu: Ok, I understand it differently, but it can be, if you think it's ok.

jonaslagoni: I mean, this text should be easily understood by others, not me 😆 

What about something like.

```suggestion
- Change existing features, by providing options that default to current behavior. This could be a preset that adapts the output based on options, as long as the API of Modelina and the API of the generated models does not have any breaking changes.
```

magicmatatjahu: ok, better :)

magicmatatjahu (approved):

Samridhi-98 (dismissed):
LGTM 🎉

jonaslagoni:
Alright, it's been 3 weeks with no objections, merging 🙂

magicmatatjahu (approved):

PR #993, opened 2022-11-16
feat: work-in-progress PHP model generator

Ferror:
## Motivation

I want to start creating PHP Generator. There is no Work In Progress PR and we all know the most challenging part is to start 😄 

> Would you like to help me? Feel free to reach out

## TODO
- [x] Use Generator Template
- [x] Define PHP reserved words / keywords https://www.php.net/manual/en/reserved.keywords.php
- [ ] Generate empty PHP files
- [ ] Generate empty classes
- [ ] Generate public typed properties
- [ ] Generate SubObjects
- [ ] Generate Enums

## Thoughts, Ideas, PHP Specifics 

> let's support only PHP 8.0 and up. PHP 7.4 is already not supported: https://www.php.net/supported-versions.php

### PHP 8.0
* private properties when readonly -> getters
* public properties instead of getters + setters

### PHP 8.1
* public readonly properties
* public properties instead of getters + setters

### PHP 8.2
* public readonly class
* public properties

## Options
* PHP Version (8.0, 8.1, 8.2) default (8.1)
* Readonly properties (true/false) default (true)
* Namespace


jonaslagoni:
@Ferror thanks for taking the time to provide this!

Do let me know if the [docs](https://github.com/asyncapi/modelina/blob/next/docs/contributing.md#new-generators) are missing stuff or don't help 🙂

jonaslagoni:

PR #1001, opened 2022-11-18
fix: rust enum renderer only working for JSON Schema inputs

jonaslagoni:
**Description**
This PR fixes a problem with the Rust enum renderer that assumes JSON Schema inputs. I also converted the enum value constrainer not to constrain the JSON value and then added a support function in the enum renderer to render the enum value type. To align with the other generators.

We might need to add something along the lines of `valueType` to the `ConstrainedEnumValueModel` 🤔

### DIFF
--- a/src/generators/rust/renderers/EnumRenderer.ts
+++ b/src/generators/rust/renderers/EnumRenderer.ts
@@ -43,26 +43,48 @@ ${additionalContent}`;
   runStructMacroPreset(): Promise<string> {
     return this.runPreset('structMacro');
   }
+
+
+  /**
+   * Returns the type for the JSON value
+   */
+  renderEnumValueType(value: any): string {
+    switch (typeof value) {
+      case 'boolean':
+        return 'bool';
+      case 'bigint':
+        return 'i64';
+      case 'number': {
+        return 'f64';
+      }
+      case 'object': {
+        return 'HashMap<String, serde_json::Value>';
+      }
+      case 'string':
+      default: {
+        return 'String';
+      }
+    }
+  }
 }
 
 export const RUST_DEFAULT_ENUM_PRESET: EnumPresetType<RustOptions> = {
   self({ renderer }) {
     return renderer.defaultSelf();
   },
-  item({ item }) {
-    if (item.value === 'HashMap<String, serde_json::Value>') {
-      return `${item.key}(${item.value})`;
+  item({ item, renderer }) {
+    const typeOfEnumValue = renderer.renderEnumValueType(item.value);
+    if (typeOfEnumValue === 'HashMap<String, serde_json::Value>') {
+      return `${item.key}(${typeOfEnumValue})`;
     }
     return `${item.key}`;
   },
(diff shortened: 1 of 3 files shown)


src/generators/rust/renderers/EnumRenderer.ts
```diff
@@ -43,26 +43,47 @@ ${additionalContent}`;
...
-    const originalInput = model.originalInput.enum[Number(itemIndex)];
+  itemMacro({ item }) {
     const serdeArgs = [];
-    if (typeof originalInput === 'object') {
+    if (typeof item.value === 'object') {
       serdeArgs.push('flatten');
     } else {
-      const rename = model.originalInput.enum[Number(itemIndex)].toString();
-      serdeArgs.push(`rename="${rename}"`);
+      serdeArgs.push(`rename="${item.value}"`);
```
leigh-johnson, line 86: `rename` is a serde macro that will munge the property name on serialization/deserialization, for example munging camelCase to snake_case.

Is `${item.value}` the right value here? Usually, rename is used to rewrite the property name. Quick Example :point_down: 

https://serde.rs/field-attrs.html#field-attributes

```rust

// ExampleStruct will serialize to/from camelCased JSON: `{ "propertyOne": 10 }`
#[derive(Debug, Clone, Deserialize, Serialize)]
struct ExampleStruct {
    #[serde(rename = "propertyOne")]
    property_one: i32
}
```

jonaslagoni: It does because the enum value constrainer for Rust returns the JSON value as is now, i.e. the same you did with `model.originalInput.enum[Number(itemIndex)].toString();` 🙂 

All the tests for the enum generator are exactly the same snapshot.

src/generators/rust/renderers/EnumRenderer.ts
```diff
@@ -43,26 +43,47 @@ ${additionalContent}`;
...
   self({ renderer }) {
     return renderer.defaultSelf();
   },
-  item({ item }) {
-    if (item.value === 'HashMap<String, serde_json::Value>') {
-      return `${item.key}(${item.value})`;
+  item({ item, renderer }) {
+    const typeOfEnumValue = renderer.renderEnumValueType(item.value);
+    if (typeOfEnumValue === 'HashMap<String, serde_json::Value>') {
+      return `${item.key}(${typeOfEnumValue})`;
```
leigh-johnson, line 77: I put together a playground to help clarify when ``${item.key}(${typeOfEnumValue})` needs to be returned vs. `${item.key}`

Have a noodle around and let me know what you think: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5c82a5c0219f7e77f74f6f93e2e58208

leigh-johnson (approved):

src/generators/rust/renderers/EnumRenderer.ts
jonaslagoni: I gotta try using the Rust models for a project at some point! 😄

PR #1002, opened 2022-11-18
fix: dependencies are rendered twice

jonaslagoni:
**Description**
It can happen that a reference has the exact same name, but a different instance. This PR fixes this. It also moves the `DependencyHelper` into the `helpers` directory to stay consistent.


src/helpers/DependencyHelpers.ts
```diff
@@ -19,11 +19,9 @@ export function renderJavaScriptDependency(toImport: string, fromModule: string,
...
 export function makeUnique(array: ConstrainedMetaModel[]): ConstrainedMetaModel[] {
-  const seen: Map<ConstrainedMetaModel, boolean> = new Map();
+  const seen: Map<string, boolean> = new Map();
   return array.filter((item: ConstrainedMetaModel) => {
-    if (item instanceof ConstrainedReferenceModel) {
-      return seen.has(item.ref) ? false : seen.set(item.ref, true);
-    }
-    return seen.has(item) ? false : seen.set(item, true);
+    const naiveIdentifier = item.name + item.type;
+    return seen.has(naiveIdentifier) ? false : seen.set(naiveIdentifier, true);
```
magicmatatjahu, line 25: `set` returns the reference to the map. Also you should use `Set`, not `Map` in this function to check duplication..

magicmatatjahu (changes requested):

src/helpers/DependencyHelpers.ts
jonaslagoni: Alright switched to using a `Set` 🙂

magicmatatjahu (approved):

PR #1019, opened 2022-11-28
fix: use type as a fallback if no special handling for format is provided

### LINKED ISSUE #533: Java generation - support for 'format' classifier
In the specification when I define my schema I can define a format e.g. datetime, email, etc.

At the moment when I generate with modelina java DTOs properties having a format defines are given a type of "Object".

Having the generated code support the api specified constraints (string = email, or string is to be interpreted as a date) is important for correctness and consistency of how an API is used.

Below is an example illustrating such an issue with date format.

image

200Puls:
**Description**

- Add type as a fallback if no special handling for format has been provided.
- Fixes #533 

**Related issue(s)**

### DIFF
--- a/examples/generate-java-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-java-models/__snapshots__/index.spec.ts.snap
@@ -3,10 +3,10 @@
 exports[`Should be able to render Java Models and should log expected output to console 1`] = `
 Array [
   "public class Root {
-  private Object email;
+  private String email;
 
-  public Object getEmail() { return this.email; }
-  public void setEmail(Object email) { this.email = email; }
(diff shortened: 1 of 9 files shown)


jonaslagoni (changes requested):
You have some tests that are failing @200Puls 🙂 

Might just be the snapshots needs to be updated.

200Puls:
Hi @jonaslagoni,
I have updated the snapshots and locally all tests are green now.

200Puls:
@jonaslagoni  Sorry about the formatting. I just pushed the changes again and corrected the formatting.

jonaslagoni (approved):
👍

Issue #1020, opened 2022-11-28
Generated Struct fields are private [RustGenerator]


leigh-johnson:
#### Describe the bug
Struct fields are generated without the `pub` keyword. 

#### How to Reproduce

Example struct codegen:
```rust
// SettingsLoadRequest represents a SettingsLoadRequest model.
#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub struct SettingsLoadRequest {
    #[serde(rename="format")]
    format: Box<crate::SettingsFormat>,
    #[serde(rename="filename")]
    filename: Box<crate::SettingsFile>,
}

impl SettingsLoadRequest {
    pub fn new(format: crate::SettingsFormat, filename: crate::SettingsFile) -> SettingsLoadRequest {
        SettingsLoadRequest {
            format: Box::new(format),
            filename: Box::new(filename),
        }
    }
}
```

#### Expected behavior
Structs should be generated with `pub` keyword. Using the example above, this code should be:
```rust
// SettingsLoadRequest represents a SettingsLoadRequest model.
#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub struct SettingsLoadRequest {
    #[serde(rename="format")]
    pub format: Box<crate::SettingsFormat>,
    #[serde(rename="filename")]
    pub filename: Box<crate::SettingsFile>,
}

impl SettingsLoadRequest {
    pub fn new(format: crate::SettingsFormat, filename: crate::SettingsFile) -> SettingsLoadRequest {
        SettingsLoadRequest {
            format: Box::new(format),
            filename: Box::new(filename),
        }
    }
}
```

PR #1021, opened 2022-11-28
fix: structs missing pub keyword in RustGenerator

leigh-johnson:
**Description**

RustGenerator outputs structs missing the `pub` keyword. Without `pub`, these fields are private to the implementation of the struct. 

**Related issue(s)**
closes: https://github.com/asyncapi/modelina/issues/1020


jonaslagoni (approved):
👍

PR #1028, opened 2022-12-01
fix: generated TS interfaces extend type defined by additional properties

### LINKED ISSUE #959: additionalProperties within TS interfaces
#### Describe the bug
When generating interfaces (instead od default classes) for typescript, `additionalProperties` are added as actual property to the interface.

For example:
```json

"Foo": {
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "foo": {
      "type": "string"
    }
  },
  "required": ["id", "foo"],
  "additionalProperties": {
    "type": "string"
  }
}
```

Produces such interface:
```ts
interface Foo {
  id: string;
  foo: string;
  additionalProperties?: Record<string, string>;
}
```

AFAIK additional property should allow for adding properties not listed in the interface, probably something like this:
```ts
interface Foo extends Record<string, string> {
  id: string;
  foo: string;
}
```

At the same time if additionalProperties is not present in the asyncapi definition for example, it is considered as if additiona properties are allowed. This probably means that interfaces should by default all extend `Record<string, any>` in absence of additionalProperties?

mitjacotic:
**Description**

When generating TS interfaces right now, additionalPropertes are added as actual property to the intrerface, eg:
```ts
interface Foo {
  additionalProperties?: Map<string, any>
}
```

this is wrong, instead interface should extend such type:
```ts
interface Foo extends Map<string, any> {

}
```

This PR fixes that:
- interfaces do not contain `additionalProperties` property
- interfaces extend type defined by additional properties when using `record` and `map` mapType
- extending can be turned off using `extendInterfaceWithAdditionalProperties` config

**Related issue(s)**
Resolves #959

### DIFF
--- a/examples/typescript-generate-comments/__snapshots__/index.spec.ts.snap
+++ b/examples/typescript-generate-comments/__snapshots__/index.spec.ts.snap
@@ -6,7 +6,7 @@ Array [
     "/**
  * Main Description
  */
-interface Test {
+interface Test extends Map<string, any> {
   stringProp: string;
   /**
    * Description
@@ -17,7 +17,6 @@ interface Test {
    * @example Example 1, Example 2
    */
   objectProp?: NestedTest;
-  additionalProperties?: Map<string, any>;
 }",
   ],
   Array [
@@ -26,7 +25,6 @@ interface Test {
  */
 interface NestedTest {
   stringProp?: string;
-  additionalProperties?: Map<string, any>;
 }",
   ],
 ]
--- a/examples/typescript-generate-comments/index.ts
+++ b/examples/typescript-generate-comments/index.ts
@@ -25,6 +25,7 @@ const jsonSchemaDraft7 = {
       $id: 'NestedTest',
       properties: { stringProp: { type: 'string' } },
       examples: ['Example 1', 'Example 2'],
+      additionalProperties: false,
     },
   },
 };
--- a/src/generators/typescript/TypeScriptGenerator.ts
+++ b/src/generators/typescript/TypeScriptGenerator.ts
@@ -22,6 +22,7 @@ export interface TypeScriptOptions extends CommonGeneratorOptions<TypeScriptPres
   typeMapping: TypeMapping<TypeScriptOptions>;
   constraints: Constraints;
   moduleSystem: 'ESM' | 'CJS';
+  extendInterfaceWithAdditionalProperties: boolean;
 }
 
 export interface TypeScriptRenderCompleteModelOptions {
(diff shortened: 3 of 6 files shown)


jonaslagoni (commented):
Thanks for taking the time to provide this PR @mitjacotic! Let me know if you need anything from me to finish it 🙂

mitjacotic:
@jonaslagoni It slipped my mind, but in the past I had similar issue with mixed types where some keys are known and others not. I found various issues on this topic: https://github.com/microsoft/TypeScript/issues/17867, https://github.com/microsoft/TypeScript/issues/20597 etc.

Here are a few points that I've checked:

### 1
Extended type should always indicate that value can be undefined:
```ts
interface a extends Record<string, string | undefined> {}
```

right now what this PR produces is:
```ts
interface a extends Record<string, string> {}
const foo: a = {}

a.bar // TS thinks this is a number
```

### 2
known keys inside the interface need to be of same type/subtype as in the type interface is extending:
```ts

// OK:
interface a extends Record<string, any> {
  known_key: string
}
interface a extends Record<string, string> {
  known_key: string
}
interface a extends Record<string, string | number> {
  known_key: string
}

// NOT OK, TS errors out:
interface a extends Record<string, number> {
  known_key: string
}
```

I believe this last example is the one that should be covered, if it can't be, I don't really see this PR solves anything.

#### 2.1

The above problem can kin of be fixed using types instead:
```ts
type a = { [key: string]: number | undefined } & {
  known_key: string;
}
```
But the problem with such type is that a variable can not be defined without force casting:
```ts

// fails saying that known_string is incompatible with index signature
const foo: a = {
  known_key: 'wer'
}

// fails since known_key is required:
const foo: a = {}

// but setters/getters work ok 
const foo: a = { } as unknown as bar

a.known_key // is string
a.foo // is number or undefined

a.known_key = 'foo' // OK
a.foo = 123 // OK
a.bar = 'wer' // OK - errors since it should be number

```


Generator currently only generates classes or interfaces, so the 2.1 option with types would require a new option to generate types.
And I don't see sense in having this functionality in just for the types and not interfaces too, especially when even using types would require force casting. What do you think? I'm thinking of just closing this PR, hopefully TS will gain support for such mixed types in the future.

jonaslagoni:
Hmm, quite a dilemma 😆 

Do you see it as feasible to do the following:
- Look through all the properties and see if one contain `ConstrainedDictionaryModel` with serialization type `unwrap` (this is where the `additionalProperty` will be located).
- If it contains it you know you need to extend the interface, now it's time to determine the type for the values. This is where we can do it complex, or easy.
  - Easy approach (or good first approach): Use `any`
  - Harder approach: Gather all the property types as a union, and use that value. This is the best approximation of which types can be used.

jonaslagoni:

PR #1033, opened 2022-12-04
Fix: CSharp: Using dependency management / partOfProperty to render optional types with question mark

### LINKED ISSUE #941: C# generator does not render optional properties correctly

zaytsevand:
**Description**

- I wasn't able to test changes since tests are falling with
```
    Internal error, could not find dependency manager instance

      55 |   protected getDependencyManagerInstance(options: Options): AbstractDependencyManager {
      56 |     if (options.dependencyManager === undefined) {
    > 57 |       throw new Error('Internal error, could not find dependency manager instance');
         |             ^
      58 |     }
      59 |     if (typeof options.dependencyManager === 'function') {
      60 |       return options.dependencyManager();

      at CSharpGenerator.getDependencyManagerInstance (src/generators/AbstractGenerator.ts:57:13)
      at map (src/generators/AbstractGenerator.ts:94:38)
          at Array.map (<anonymous>)
      at CSharpGenerator.generate (src/generators/AbstractGenerator.ts:93:54)
      at generate (examples/csharp-generate-required-properties/index.ts:26:18)
      at Object.<anonymous> (examples/csharp-generate-required-properties/index.spec.ts:9:5)
```

**Related issue(s)**
Fixes #941

### DIFF
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -283,16 +283,20 @@ const generator2 = new TypeScriptGenerator({
 ```
 
 ### Adding new dependencies
-Sometimes the preset might need to use some kind of foreign dependency. To achieve this each preset hook has the possibility of adding its own dependencies through the `addDependency` function from `renderer` property.
+Sometimes the preset might need to use some kind of foreign dependency. To achieve this each preset hook has the possibility of adding its own dependencies through a dependency manager, which can be accessed in `dependencyManager`.
 
 ```ts
 ...
-self({ renderer, content }) {
-  renderer.addDependency('import java.util.*;');
+self({ dependencyManager, content }) {
+  dependencyManager.addDependency('import java.util.*;');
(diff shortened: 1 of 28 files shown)


zaytsevand:
> Thanks for raising the PR @zaytsevand!
> 
> Do you mind basing the change on the `next` branch instead of the dependency manager? You still are able to perform exactly this there without the errors 🙂

Sure thing, no prob. It seems that I missed the fact that the dependency manager is already available in `next`. Unfortunately, it looks like I've to create a new PR for that. Give me a second :)

jonaslagoni:
> It seems that I missed the fact that the dependency manager is already available in next.

It is not, it is still a work in progress.

But the changes you proposed did not depend on that feature, did they? Or did I simply miss that😅?

PR #1037, opened 2022-12-06
fix: seesharp: Using dependency management / partOfProperty to render optional types with question mark

### LINKED ISSUE #941: C# generator does not render optional properties correctly

zaytsevand:
**Description**

**Related issue(s)**
Fixes #941

### DIFF
--- a/test/generators/csharp/presets/__snapshots__/JsonSerializerPreset.spec.ts.snap
+++ b/test/generators/csharp/presets/__snapshots__/JsonSerializerPreset.spec.ts.snap
@@ -5,10 +5,10 @@ exports[`JSON serializer preset should render serialize and deserialize converte
 public class Test
 {
   private string stringProp;
-  private double numberProp;
-  private EnumTest enumProp;
-  private NestedTest objectProp;
-  private Dictionary<string, dynamic> additionalProperties;
+  private double? numberProp;
+  private EnumTest? enumProp;
+  private NestedTest? objectProp;
+  private Dictionary<string, dynamic>? additionalProperties;
 
   public string StringProp 
   {
@@ -16,25 +16,25 @@ public class Test
     set { stringProp = value; }
   }
 
-  public double NumberProp 
+  public double? NumberProp 
   {
     get { return numberProp; }
     set { numberProp = value; }
   }
 
-  public EnumTest EnumProp 
+  public EnumTest? EnumProp 
   {
     get { return enumProp; }
     set { enumProp = value; }
   }
 
-  public NestedTest ObjectProp 
+  public NestedTest? ObjectProp 
   {
     get { return objectProp; }
     set { objectProp = value; }
   }
 
-  public Dictionary<string, dynamic> AdditionalProperties 
+  public Dictionary<string, dynamic>? AdditionalProperties 
   {
     get { return additionalProperties; }
     set { additionalProperties = value; }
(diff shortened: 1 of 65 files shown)


test/generators/csharp/presets/__snapshots__/JsonSerializerPreset.spec.ts.snap
```diff
@@ -79,24 +79,24 @@ internal class TestConverter : JsonConverter<Test>
...
         {
-          var value = JsonSerializer.Deserialize<double>(ref reader, options);
+          var value = JsonSerializer.Deserialize<double?>(ref reader, options);
           instance.numberProp = value;
           continue;
         }
       if (propertyName == \\"enumProp\\")
         {
-          var value = EnumTestExtension.ToEnumTest(JsonSerializer.Deserialize<dynamic>(ref reader, options));
+          var value = EnumTest?Extension.ToEnumTest?(JsonSerializer.Deserialize<dynamic>(ref reader, options));
```
jonaslagoni, line 88: This is definitely wrong 😆 

You need to change this line: https://github.com/asyncapi/modelina/blob/3cfd68b877d04f48c492b1eb53cf58b1c2d7179d/src/generators/csharp/presets/JsonSerializerPreset.ts#L91

to use the `.name` instead of type.

test/generators/csharp/presets/__snapshots__/NewtonsoftSerializerPreset.spec.ts.snap
```diff
@@ -52,13 +52,13 @@ public class TestConverter : JsonConverter<Test>
   value.StringProp = jo[\\"string prop\\"].ToObject<string>(serializer);
 }
 if(jo[\\"numberProp\\"] != null) {
-  value.NumberProp = jo[\\"numberProp\\"].ToObject<double>(serializer);
+  value.NumberProp = jo[\\"numberProp\\"].ToObject<double?>(serializer);
 }
 if(jo[\\"enumProp\\"] != null) {
-  value.EnumProp = EnumTestExtensions.ToEnumTest(jo[\\"enumProp\\"]);
+  value.EnumProp = EnumTest?Extensions.ToEnumTest?(jo[\\"enumProp\\"]);
```
jonaslagoni, line 58: Same as the other comment, need to change the following lines: 
- https://github.com/asyncapi/modelina/blob/3cfd68b877d04f48c492b1eb53cf58b1c2d7179d/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts#L19
- https://github.com/asyncapi/modelina/blob/3cfd68b877d04f48c492b1eb53cf58b1c2d7179d/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts#L71

jonaslagoni (changes requested):
Before this PR, `type` and `name` were interchangeable because they contained the same value. Now they no longer do that, and we were not strict in the past when what was used 😅 

Other then that you have some linting errors 🙂

zaytsevand:
Updated. Thanks, I missed that over all the changes!
Btw, from my understanding, all the hassle with custom Converter classes are there because of enums. Am I getting it correctly?
If so, do you mind shortening a lot of what is generated by using another technique to customize output?
Code example: https://dotnetfiddle.net/365DN6

jonaslagoni:
> Btw, from my understanding, all the hassle with custom Converter classes is there because of enums. Am I getting it correctly?

Do you mean the to model and GetValue methods? 

They are there because the core models are not meant to depend on any outside library, that is what the presets are for 🙂 

You could definitely argue that the `NewtonsoftSerializerPreset` should add this annotation to the enums 👍

jonaslagoni (changes requested):
@zaytsevand it seems you have merged the dependency manager changes into the branch again. I have a feeling you have tainted your local `next` branch as you merged it into your branch here: https://github.com/asyncapi/modelina/pull/1037/commits/d2a2885b9cf26a86562f3285b90cc73cb014d639

PR #1038, opened 2022-12-06
chore: add dependency manager to Rust

jonaslagoni:
**Description**
This PR adds the new dependency manager to the Rust generator alongside a few changes:
- Allow the user to add dependencies for each model in the constraint and presets phase.
- As a side effect of the implementation, this enables the user to selectively overwrite options so you do not have to create a new generator instance all the time. 

These changes currently live on the `dependency_-manager` (wups) branch and are a WIP until all generators are adapted, that is why the tests are failing. 

None of these changes affect the current output of the Rust generator.

**Related issue(s)**
Read more about the change here: https://github.com/asyncapi/modelina/pull/947
Fixes: https://github.com/asyncapi/modelina/issues/851


src/generators/rust/RustDependencyManager.ts
```diff
@@ -0,0 +1,12 @@
+import { AbstractDependencyManager } from '../AbstractDependencyManager';
+import { RustOptions } from './RustGenerator';
+
+export class RustDependencyManager extends AbstractDependencyManager {
```
jonaslagoni, line 4: This can be very diverse in what it can be used for, for example in TypeScript we added support functions to help better add dependencies based on the different module systems 🙂 https://github.com/asyncapi/modelina/blob/bfe8d78df9db61f467c1f0aee6e2d4d4d1b8f7b7/src/generators/typescript/TypeScriptDependencyManager.ts#L32

jonaslagoni: I also imagine that in the future, we can have it include package imports etc, to dynamically add dependency libraries for the support files. However, that is for the future if it's ever requested 🙂

magicmatatjahu (approved):

PR #1051, opened 2022-12-08
fix: csharp generator not render optional types for optional properties

### LINKED ISSUE #941: C# generator does not render optional properties correctly

zaytsevand:
**Description**

**Related issue(s)**
Fixes #941

### DIFF
--- a/src/generators/csharp/CSharpConstrainer.ts
+++ b/src/generators/csharp/CSharpConstrainer.ts
@@ -1,5 +1,4 @@
-import { ConstrainedObjectPropertyModel } from 'models';
-import { ConstrainedEnumValueModel } from '../../models';
+import { ConstrainedEnumValueModel, ConstrainedObjectPropertyModel } from '../../models';
 import { TypeMapping } from '../../helpers';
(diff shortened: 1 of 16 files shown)


zaytsevand:
@jonaslagoni I don't get what's wrong here, but black-box tests are constantly failing

src/generators/csharp/CSharpConstrainer.ts
```diff
@@ -1,10 +1,17 @@
+import { ConstrainedObjectPropertyModel } from 'models';
 import { ConstrainedEnumValueModel } from '../../models';
```
jonaslagoni, line 2:
```suggestion
import { ConstrainedEnumValueModel, ConstrainedObjectPropertyModel } from '../../models';
```

jonaslagoni (changes requested):
My comment fixes your blackbox tests and the linter you can probably fix with `npm run lint:fix`

jonaslagoni (approved):
LGTM 🔥

Issue #1055, opened 2022-12-14
oneOf models are rendered with incorrect member names [RustGenerator]


leigh-johnson:
#### Describe the bug
Given the following schema, member names and serde macros are rendered with the index of the member appended to the model name. 

```yaml
    WebrtcSettingsApplyRequest:
      name: WebrtcSettingsApplyRequest
      type: object
      additionalProperties: false
      properties:
        video_src:
          x-parser-schema-id: VideoSource
          oneOf:
            - $ref: "#/components/schemas/Camera"
            - $ref: "#/components/schemas/PlaybackVideo"
      required:
        - video_src
 ```

The rust code currently rendered:

```rust
// VideoSource represents a union of types: Camera0, PlaybackVideo1
#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub enum VideoSource {
    #[serde(rename="Camera0")]
    Camera0(crate::Camera),
    #[serde(rename="PlaybackVideo1")]
    PlaybackVideo1(crate::PlaybackVideo),
}
```

#### Expected behavior

The schema above should render the following model:

```rust
// VideoSource represents a union of types: Camera, PlaybackVideo
#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub enum VideoSource {
    #[serde(rename="Camera")]
    Camera(crate::Camera),
    #[serde(rename="PlaybackVideo")]
    PlaybackVideo(crate::PlaybackVideo),
}
```

leigh-johnson:
Dropping some debugging notes. Looks like `ConstrainedUnionModel` is getting an array of `ConstrainedReferenceModel` with the wrong types (`Cat0`, `Dog1`, `StickInsect2`). 

![Screenshot from 2022-12-14 10-45-24](https://user-images.githubusercontent.com/2601819/207686177-ba49e2d7-1e22-4f31-86d7-4bd9781124be.png)

leigh-johnson:
I just noticed `discriminator` isn't being handled correctly either, going to fix that in #1056 as well.

PR #1056, opened 2022-12-14
fix: polymorphic / union models rendered with index in enum member name

leigh-johnson:
**Description**
This fixes up an implementation that has never been handled correctly in the RustGenerator. Figured we should get this right before the v1.0.0 release.

IIRC, I added this before understanding that the MetaModel base classes already handle de-dupling union member models:
```
    // How to constrain Union member names?
    if (model instanceof UnionModel) {
      model.union = model.union.map((v, idx) => {
        v.name += idx;
        return v;
      });
    }
```

**Related issue(s)**
Closes: https://github.com/asyncapi/modelina/issues/1055


src/generators/rust/renderers/UnionRenderer.ts
```diff
@@ -46,7 +47,21 @@ export const RUST_DEFAULT_UNION_PRESET: UnionPresetType<RustOptions> = {
   },
 
   structMacro({ model, renderer }) {
-    return renderer.renderMacro(model);
+    const blocks = [renderer.renderMacro(model)];
+    if (model.originalInput.discriminator !== undefined) {
+      // when discriminator is provided, polymorphic type is internally-tagged: 
+      // https://serde.rs/enum-representations.html#internally-tagged
+      const discriminatorBlock = `#[serde(tag = "${model.originalInput.discriminator}")]`;
+      blocks.push(discriminatorBlock);
```
jonaslagoni, line 55: Eventually, we should probably look at how a discriminator keyword can live within the metamodel itself, instead of relying on originalInput.

jonaslagoni (approved):
LGTM 👍

src/generators/rust/renderers/UnionRenderer.ts
leigh-johnson: SGTM. I'll stay tuned for that! 👀

PR #1057, opened 2022-12-14
fix: csharp simplification for newtonsoft serialization

### LINKED ISSUE #1052: Feature: For Newtonsoft.Json preset use EnumMember and StringEnumConverter to serialize and deserialize enum values [CSharpGenerator]
#### Reason/Context
- It will allow removing many autogenerated classes: Converters, static maps from enum values to strings
- It follows a common declarative approach to defining DTO classes in C#
- Less code -- more error-prone and flexible for future change

#### Description
Simple implementation:
- https://dotnetfiddle.net/365DN6

zaytsevand:
**Description**
- ⚠️ This is a breaking change:
   - enum fields in models can no longer be converted to dynamic (object) values. If an enum has mixed value types, it's no longer supported: values are always strings
   - Serializer and Deserializer classes are removed. If anyone previously used them directly, the code would fail to compile.
- Switching to declarative serialization & deserialization of enums
- Custom `*Serializer` and `*Deserializer` classes are removed from the generated code because they aren't required anymore
- Classes to map enum values to enum members are removed from the generated code
- All for the cause of reducing complexity: image

**Related issue(s)**

Resolves #1052

### DIFF
--- a/examples/csharp-generate-newtonsoft-serializer/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-generate-newtonsoft-serializer/__snapshots__/index.spec.ts.snap
@@ -1,9 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Should be able to generate a model with functions to serialize the data model into JSON  and should log expected output to console 1`] = `
+exports[`Should be able to generate a model  and should log expected output to console 1`] = `
 Array [
-  "[JsonConverter(typeof(RootConverter))]
-public class Root
+  "public class Root
 {
   private string email;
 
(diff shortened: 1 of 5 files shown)


zaytsevand:
It's possible to support numeric values for enums by assigning them directly to members, like:
```csharp
public enum Example
{
    FirstValue = 1,
    SecondValue = 12
}
```
This approach would work with both `Newtonsoft` and `System.Text` serializers. But it would fail to support mixed types of values for members, and values other than integers. So it would solve only a part of a problem.

jonaslagoni (commented):
Lets discuss further in https://github.com/asyncapi/modelina/issues/1052 🙂

jonaslagoni:
Closing as [issue got closed](https://github.com/asyncapi/modelina/issues/1052), feel free to reopen 🙂

Issue #1066, opened 2022-12-20
Javax Constraints annotations should be added at the field level


artur-ciocanu:
#### Describe the bug
Currently Java constraints preset is applied at the "getter" level. 

#### How to Reproduce
Java constraints annotations are applied at the "getter" hence the validation rules won't be applied during object instantiations.

#### Expected behavior
The Java constraints annotations should be applied at the field level to ensure that data is properly validated before object is instantiated.

PR #1067, opened 2022-12-20
fix: add Java contraints annotations at the field level

artur-ciocanu:
**Description**
Javax annotations should be applied at the field level to ensure that all the validation rules are properly invoked. Currently these are applied at the "getter" level.

**Related issue(s)**
Resolves https://github.com/asyncapi/modelina/issues/1066


artur-ciocanu:
@jonaslagoni could you please take a look when you have a chance. Thank you.

jonaslagoni (commented):
Just had to look it up as I couldn't remember which was the right way, but yea this is 100% correct 👍 

What is the reason for renaming the presets files? 🤔

artur-ciocanu:
> Just had to look it up as I couldn't remember which was the right way, but yea this is 100% correct 👍
> 
> What is the reason for renaming the presets files? 🤔

I just wanted to have those reflect the name of the preset constants. If you think it should be reverted let me know and I'll revert.

jonaslagoni:
@artur-ciocanu since most presets do not have the language as part of their name, I think it makes the most sense to keep it as is 🙂 But feel free to start a discussion in an issue 👍

artur-ciocanu:
> @artur-ciocanu since most presets do not have the language as part of their name, I think it makes the most sense to keep it as is 🙂 But feel free to start a discussion in an issue 👍

@jonaslagoni I have reverted the renaming, I haven't checked other languages, but it makes sense to be consistent. Thank you.

PR #674, opened 2022-03-08
feat: release next

jonaslagoni:
**Description**

This PR is for making sure next stay up to date with master.

PR #1071, opened 2022-12-29
feat: adding interpreter options override

artur-ciocanu:
**Description**
This is an enhancement that would allow anyone to override default interpreter behavior. Some of use cases are overriding default behavior of `additionalProperties`, `additionalItems` that would allow anyone to customize how these schema properties are handled. For example by default these properties are treated as `true`, however we might want to treat them as `false` during code gen.

A typical example using interpreter options would look like this:
```
const generator = new JavaGenerator({
  processorOptions: {
    interpreter: {
      allOf: (schema, model, interpreter, interpreterOptions) => {
        // some custom logic for interpreting "allOf"
      }
    }
  }
});
```

**Related issue(s)**
https://github.com/asyncapi/modelina/issues/916

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -15,6 +15,7 @@ This document contains many of the advanced use-cases that you may stumble upon 
 - [Change the generated indentation type and size](#change-the-generated-indentation-type-and-size)
 - [Change the type mapping](#change-the-type-mapping)
 - [Changing the constrain rules](#changing-the-constrain-rules)
+- [Changing the interpreter options](#changing-the-interpreter-options)
 
 <!-- tocstop -->
 
@@ -98,3 +99,11 @@ There can be multiple reasons why you want to change the default constrain rules
 Check out this [example out for a live demonstration](../examples/overwrite-default-constraint/) for how to overwrite the default constraints.
 
 Check out this [example out for a live demonstration](../examples/overwrite-naming-formatting/) for how to overwrite the naming formatting for models.
+
+## Changing the interpreter options
+
+Sometimes it is desirable to change the default behavior of the `Interpreter`. While most of the default behavior of the `Interpreter` is fine, there are areas that would be nice to customize. Some of these are:
+- `additionalProperties` - according to JSON schema spec, by default, `additionalProperties` is set to `true`
+- `additionalItems` - according to JSON schema spec, by default, `additionalItems` is set to `true`
+
+Check out this [example out for a live demonstration](../examples/overwrite-interpreter-options/) for how to overwrite the interpreter options.
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -106,4 +106,4 @@ Sometimes it is desirable to change the default behavior of the `Interpreter`. W
 - `additionalProperties` - according to JSON schema spec, by default, `additionalProperties` is set to `true`
 - `additionalItems` - according to JSON schema spec, by default, `additionalItems` is set to `true`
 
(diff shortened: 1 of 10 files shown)


artur-ciocanu:
@jonaslagoni this one is back 😄 . Could you please take a look. Thank you.

jonaslagoni (changes requested):
@artur-ciocanu nice!

We do need some extra changes though, cause without them, the users won't the feature is there! So we need to make the documentation reflect the use-case and then provide an example where you show how it can be done 🙂

200Puls:
This is really super useful. 
We were debating on creating something similar for our usecase (turning additionalProperties to false)

artur-ciocanu:
> @artur-ciocanu nice!
> 
> We do need some extra changes though, cause without them, the users won't the feature is there! So we need to make the documentation reflect the use-case and then provide an example where you show how it can be done 🙂

@jonaslagoni I am still new to this, so could you please let me know where should I add the examples and the docs?

jonaslagoni:
@artur-ciocanu you can find out how to create new examples here: https://github.com/asyncapi/modelina/blob/next/docs/contributing.md#adding-examples

For the docs it's slightly trickier, I think we need a section here: https://github.com/asyncapi/modelina/blob/next/docs/advanced.md

It should explain the use-case of when you would want to use this feature of overwriting the interpreters as well as a link to the example you created 🙂

artur-ciocanu:
@jonaslagoni I have added an example and a new section to docs. Let me know if it is looking good.

jonaslagoni (commented):
Something I have been wondering, cant recall if we already discussed this, but if it's really only about turning interpretation on or off for `additionalProperties` and `additionalItems`, we could just add the options `ignoreAdditionalProperties` and `ignoreAdditionalItems` instead of allowing changing the entire interpretation callbacks. 

It would be both easier for the end-user and easier to implement.

But if the use-case is really to allow custom interpretations, that can't be solved by simple boolean options, then I am all aboard this feature. Just need to be for the right reasons 😄

jonaslagoni:
> Also should I rebase from `master`?

I actually cant recall what needs to change once we merge #674, so lets see 🙂

artur-ciocanu:
> Something I have been wondering, cant recall if we already discussed this, but if it's really only about turning interpretation on or off for `additionalProperties` and `additionalItems`, we could just add the options `ignoreAdditionalProperties` and `ignoreAdditionalItems` instead of allowing changing the entire interpretation callbacks.
> 
> It would be both easier for the end-user and easier to implement.
> 
> But if the use-case is really to allow custom interpretations, that can't be solved by simple boolean options, then I am all aboard this feature. Just need to be for the right reasons 😄

@jonaslagoni now that you mentioned it, I do think that in some cases we might want to have the full power of interpreter options overwrite. However I would go with simple booleans for addtionalProperties and additionaltems for now and later we can add other options in the form of function overrides.

jonaslagoni:
Sounds good @artur-ciocanu  👍 Also feel free to re-target the PR to the `master` branch 🙂  I don't think you will encounter any problems, but lets see.

artur-ciocanu:
Closing this one in favor of https://github.com/asyncapi/modelina/pull/1094.

artur-ciocanu:
> This is really super useful. We were debating on creating something similar for our usecase (turning additionalProperties to false)

@200Puls I have opened another https://github.com/asyncapi/modelina/pull/1094, please take a look. I have to close this one since `next` branch has been merged to `master`.

PR #838, opened 2022-08-03
chore: added prettier config

akshatnema:
**Description**

Added `.prettierrrc` file to add the Prettier configuration along with ESlint configuration for typescript.

**Related issue(s)**
Resolves #646

PR #1084, opened 2023-01-19
fix: java generator could generate illegal package names

### LINKED ISSUE #1075: Java Generator can generate invalid package names
#### Preamble
Java packages are the main means of source file organising in Java.
A Java package is built as a sort of folder structure, each sub-packages' name is added to the name of the package it resides it, delimited by a dot.
For example, the standard lib util package is located in: `java.util`, so if you wanted to use a HashMap, you'd import it by `import java.util.HashMap;`, since inside the HashMaps' source file, you find it's package declaration as `package java.util;`

The single packages, or sub-packages, can't be any of Java's reserved keywords.
For example, `package test.package`, `package test.class` and `package enum` would all be illegal, since `package`, `class` and `enum` are reserved.

#### Bug

The Java Generator guards against this by checking package names for the list of said reserved keywords.

However it only checks whether the package names **exactly match** one of those keywords, so it is still possible to generate invalid package names by combining them with valid root or sub packages.
So, when generating Java Code, the package name `class` would cause an Exception, but `test.class` wouldn't, which is wrong, since `class` is a keyword.

Funnily enough, the test case `should render models and their dependencies` in `JavaGenerator.spec.ts` generates an invalid package name itself, by testing whether model `Address` depending on model `OtherModel` has the dependency `import test.package.OtherModel`.

### How to Reproduce
To reproduce, describe any schema with an object, which has got another object as a property.
Use `generator.generateCompleteModels` with the config `{packageName: 'test.package'}` and look at the resulting models.

```ts
const generator = new JavaGenerator();
const jsonSchemaDraft7 = {
  $id: 'Address',
  type: 'object',
  properties: {
    other_model: { type: 'object', $id: 'OtherModel', properties: {street_name: { type: 'string' }} },
  }
};

const config = {packageName: 'test.package'};
  const models = await generator.generateCompleteModels(jsonSchemaDraft7, config);
  for (const model of models) {
    console.log(model.result);
  }
}
if (require.main === module) {
  generate();
}
```

Resulting models are:
```java
package test.package;
import test.package.OtherModel;
import java.util.Map;

public class Address {
  private OtherModel otherModel;
      
    // omitted for brevity
  }
}

package test.package;
  
import java.util.Map;
public class OtherModel {
    // omitted for brevity
}
```
With illegal sub-package `package`.

### Expected behavior
The Java Generator should throw an exception if **any of the packages and sub-packages** are a reserved keyword.
The bug itself can be found here
https://github.com/asyncapi/modelina/blob/9494ac1863dd4e8229f710c5afdaf07844146c0e/src/generators/java/JavaGenerator.ts#L62-L64

LouisXhaferi:
I adapted the JavaGenerator so that it throws an exception if *any* of the package names components is a reserved keyword.

Fixes #1075 

P.S.:
Since IntelliJ and the linter complained, I inlined some of the untyped arrays in the KotlinGenerator test.

### DIFF
--- a/src/generators/java/JavaGenerator.ts
+++ b/src/generators/java/JavaGenerator.ts
@@ -1,5 +1,5 @@
-import { 
-  AbstractGenerator, 
+import {
+  AbstractGenerator,
   CommonGeneratorOptions,
   defaultGeneratorOptions
 } from '../AbstractGenerator';
@@ -10,9 +10,9 @@ import { ClassRenderer } from './renderers/ClassRenderer';
 import { EnumRenderer } from './renderers/EnumRenderer';
 import { isReservedJavaKeyword } from './Constants';
 import { Logger } from '../../';
-import { constrainMetaModel, Constraints } from '../../helpers/ConstrainHelpers';
+import { constrainMetaModel, Constraints } from '../../helpers';
 import { JavaDefaultConstraints, JavaDefaultTypeMapping } from './JavaConstrainer';
-import { DeepPartial, mergePartialAndDefault } from '../../utils/Partials';
+import { DeepPartial, mergePartialAndDefault } from '../../utils';
 
 export interface JavaOptions extends CommonGeneratorOptions<JavaPreset> {
   collectionType: 'List' | 'Array';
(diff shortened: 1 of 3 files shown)


LouisXhaferi (commented):
This is the code that actually changed.

jonaslagoni (approved):
LGTM, we got https://github.com/asyncapi/modelina/pull/838 waiting to be merged, which should solve all those extra commits 🙂

PR #1095, opened 2023-01-24
chore: revert the last two commits

jonaslagoni:
**Description**
As the release workflow didn't trigger, this reverts it and re-commits it afterward.


KhudaDad414:
@jonaslagoni I think it would be better to use `revert: ` as type instead of `chore: ` in the title. https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits 🤔

jonaslagoni:
@KhudaDad414 I am using chore because I know it won't trigger a release 🙂 Not too sure about revert, as they also state, up to the tooling providers, and I have no idea what our tool will do.

magicmatatjahu (approved):

Samridhi-98 (approved):

PR #1099, opened 2023-01-25
fix!: use preferred ids over anonymous ids

kennethaasan:
**Description**

We're using [CloudEvents](https://cloudevents.io/) and found an issue when trying to use _allOf_ with the [CloudEvent schema](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/cloudevents.json). When extending/allOf the CloudEvents schema, and we try to set the type property to a const with title, it uses the anonymous id of the CloudEvent schema and not our own schema.

- fixes const when merging properties when mergeTo $id/title is anonymous


kennethaasan:
@jonaslagoni , please review. If you know a better way to solve this, I'm totally open to it.

src/models/CommonModel.ts
```diff
@@ -471,6 +471,16 @@ export class CommonModel {
...
         );
+
+        if (
+          mergeTo.properties[String(propName)].$id?.includes(
+            'anonymous_schema'
+          ) &&
+          !propValue.$id?.includes('anonymous_schema')
+        ) {
+          mergeTo.properties[String(propName)].$id = propValue.$id;
+        }
```
jonaslagoni, line 482: Instead of here, I think it should be added to the mergeCommonModels function as it's really there you want to selectively decide whether mergeFrom $id is better than the mergeTo is: https://github.com/asyncapi/modelina/blob/7a8197370a0c679f588e98f471558cb766b19dc4/src/models/CommonModel.ts#L773

```ts
    // Base the $id of the model on the most suited one.
    // If mergeTo is using anonymous_schema id, and the mergeFrom does not, use the mergeFrom $id.
    if (
      mergeTo.$id?.includes(
        'anonymous_schema'
      ) &&
      !mergeFrom.$id?.includes('anonymous_schema')
    ) {
      mergeTo.$id = mergeFrom.$id;
    } else {
      mergeTo.$id = mergeTo.$id || mergeFrom.$id;
    }
```

jonaslagoni (changes requested):
This is the problem I guess with JSON Schema in code-gen, that it's rather subjective what is "right".

Because if both `type`s has `title` i.e. the cloud event has `title: 'CloudEventType',`, it would be the index in the `allOf` that determines that title of the enum, which is, weird I guess 😆 

```suggestion
{
                title: 'Dog',
                allOf: [
                  {
                    type: 'object',
                    properties: {
                      type: {
                        title: 'DogType',
                        const: 'Dog'
                      }
                    }
                  },
                  {
                    $ref: '#/components/schemas/CloudEvent'
                  }
                ]
              }
```

But yea, this is definitely an improvement either way.

I think you need to add an extra test in the CommonModel or interpreter, as the test you added is only relevant for TypeScript, think it would be better to ensure that the interpreter/CommonModel creates an accurate CommonModel.

src/models/CommonModel.ts
kennethaasan: Makes sense. Done

kennethaasan:
I already have a test for the CommonModel. Let me know if it needs more.

jonaslagoni (changes requested):
This is [actually a breaking change now that I think about it](https://github.com/asyncapi/modelina#versioning-and-maintenance), even though it is a preferred behavior, it does change the model names and will break anyone having generated code with this version of Modelina.

So you have two options, hiding this behind an option for it to be released as part of V1, otherwise, it needs to be released as part of V2. I can set up the `next` branch for V2 right now so you can retarget the PR. And it will be released as part of `2.0.0-next.1`.

We don't have any specific release schedule for V2, but we can start a discussion around that afterward.

What would you like to do @kennethaasan?

kennethaasan:
Makes sense. Let's go with _2.0.0-next.1_

jonaslagoni:
> Makes sense. Let's go with _2.0.0-next.1_

Gonna prepare the `next` branch in another PR, or recreating the branch, not sure yet.

jonaslagoni:
Blocked by https://github.com/asyncapi/modelina/pull/1102

jonaslagoni (changes requested):
As this is the first change containing a breaking change, @kennethaasan do you mind adding this change in a v2 migration document? https://github.com/asyncapi/modelina/tree/master/docs/migrations

Just create it, and add a small section about this change, don't worry about the larger picture I will prettify it later 😆

jonaslagoni:
Changes look good 👍

test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -686,4 +686,117 @@ ${content}`;
...
+      asyncapi: '2.5.0',
+      info: {
+        title: 'CloudEvent example',
+        version: '1.0.0'
+      },
+      channels: {
+        pet: {
+          publish: {
+            message: {
+              // this should create a union type, but currently does not
```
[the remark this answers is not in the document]
kennethaasan: Yes, as you can see in the snap output below. https://github.com/asyncapi/modelina/issues/1111

PR #1104, opened 2023-01-25
docs: further clearify interpreter options

jonaslagoni:
**Description**
This PR further clarifies the new interpreter options and its intended use.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -102,10 +102,12 @@ Check out this [example out for a live demonstration](../examples/overwrite-nami
 
 ## Customizing the interpreter options
 
-According to JSON schema if `additionalProperties` or `additionalItems` are omitted, their default value is `true`. `Interpreter` class honors this behavior, however for people coming from other code generators like `Open API Genertor` or `JSON Schema 2 POJO` this might come as a surprise.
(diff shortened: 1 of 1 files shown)


docs/advanced.md
```diff
@@ -102,10 +102,12 @@ Check out this [example out for a live demonstration](../examples/overwrite-nami
...
+According to JSON schema draft 7, if `additionalProperties` or `additionalItems` are omitted, their default value is `true`. The `Interpreter` honors this behavior, however this creates more loose models that might not be the intention for the developer. 
 
-To avoid adding `additionalProperties: false` or `additionalItems: false` to all your specs just to get the same behavior as other code generators, this can be achieved by configuring interpreter options:
-- `ignoreAdditionalProperties` - if set, it ensures that `additionalProperties` is considered to be set to `false`
-- `ignoreAdditionalItems` - if set, it ensures that `additionalItems` is considered to be set to `false`
+We suggest to not use this option, as it limits your models when they are in compliance with the schema it was generated from. Instead adapt your schemas to be more strict by setting these keywords to `false`. This option should really only be used when you have no control over your input.
```
jonaslagoni, line 107:
```suggestion
We suggest not using this option if it can be avoided, as it limits your generated models to include any additional properties, and would simply be ignored. Instead adapt your schemas to be more strict by setting these keywords to `false`. This option should really only be used when you have no control over your input and the output is unintended.
```

magicmatatjahu (approved):

PR #1105, opened 2023-01-25
fix: do not honor ignoreAdditionalProperties and ignoreAddtionalItems if properties exist

artur-ciocanu:
**Description**
Recently we have introduced the ability to handle the default behavior of `additionalProperties` and `additionalItems`. However we might have introduced an unintended side effect. For example for a schema like:
```
{
  type: 'object',
  properties: {
    email: {
      type: 'string',
      format: 'email'
    }
  }
}
```
If we use `ignoreAdditionalProperties` then for TypeScript generator we would get something like:
```
class Root {
  private _email?: string;

  constructor(input: {
    email?: string,
  }) {
    this._email = input.email;
  }

  get email(): string | undefined { return this._email; }
  set email(email: string | undefined) { this._email = email; }
}
```
For a schema like this:
```
{
  type: 'object',
  properties: {
    email: {
      type: 'string',
      format: 'email'
    }
  },
  additionalProperties: {
     type: 'string'
  }
}
```
We would get the same TypeScript output.

My proposal is to fix this and ensure that the output is something like:
```
class Root {
  private _email?: string;
  private _additionalProperties?: Map<string, string>;

  constructor(input: {
    email?: string,
    additionalProperties?: Map<string, string>,
  }) {
    this._email = input.email;
    this._additionalProperties = input.additionalProperties;
  }

  get email(): string | undefined { return this._email; }
  set email(email: string | undefined) { this._email = email; }

  get additionalProperties(): Map<string, string> | undefined { return this._additionalProperties; }
  set additionalProperties(additionalProperties: Map<string, string> | undefined) { this._additionalProperties = additionalProperties; }
}
```

The reasoning being that `ignoreAdditionalProperties` should affect ONLY the default JSON schema behavior, however if user explicitly sets it to some other value `ignoreAdditionalProperties` should have no effect.

### DIFF
--- a/src/interpreter/InterpretAdditionalItems.ts
+++ b/src/interpreter/InterpretAdditionalItems.ts
@@ -19,14 +19,13 @@ export default function interpretAdditionalItems(
   interpreter: Interpreter,
   interpreterOptions: InterpreterOptions = Interpreter.defaultInterpreterOptions
 ): void {
-  if (interpreterOptions.ignoreAdditionalItems === true) {
-    return;
-  }
   if (typeof schema === 'boolean' || model.type?.includes('array') === false) {
     return;
   }
   const additionalItemsModel = interpreter.interpret(
-    schema.additionalItems === undefined ? true : schema.additionalItems,
+    schema.additionalItems === undefined
+      ? !interpreterOptions.ignoreAdditionalItems
+      : schema.additionalItems,
     interpreterOptions
   );
   if (additionalItemsModel !== undefined) {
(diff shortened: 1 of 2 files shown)


artur-ciocanu:
@jonaslagoni let me know your thoughts. Thank you.

jonaslagoni (changes requested):
I was working on a similar fix, but I also included an additional perspective where it's applied to unintended scenarios.

Can you take a look at https://github.com/asyncapi/modelina/pull/1106?

Feel free to apply the changes to your PR if you want to continue here, otherwise, we can use https://github.com/asyncapi/modelina/pull/1106 going forward, up to you 🙂

artur-ciocanu:
Thanks @jonaslagoni I'lll close it in favor of #1106

PR #1106, opened 2023-01-25
fix: add conditions for when to ignore additionalProperties and additionalItems

jonaslagoni:
**Description**
This PR fixes the two options always ignoring additionalProperties and additionalItems when it should be conditional on two things:

1. Should only apply on when defaulting, not when explicitly defined.
2. ignoreAdditionalProperties should only apply when the schema has properties, otherwise its normally later interpreted as a map type which is intended.
3. ignoreAdditionalProperties should only apply when the schema has explicit types defined, otherwise its the same as setting types explicitly.

### DIFF
--- a/src/interpreter/InterpretAdditionalItems.ts
+++ b/src/interpreter/InterpretAdditionalItems.ts
@@ -25,11 +25,15 @@ export default function interpretAdditionalItems(
   const hasArrayTypes = schema.items !== undefined;
   let defaultAdditionalItems = true;
   if (hasArrayTypes && interpreterOptions.ignoreAdditionalItems !== undefined) {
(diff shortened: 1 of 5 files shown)


magicmatatjahu:
I can accept PR, but if properties/items aren't defined it not means that given object cannot have it, e.g. schemas like:

```yaml
type: object
``` 

or

```yaml
type: array
```

means that every property/item can be defined, so I think that `ignoreAdditionalItems` and `ignoreAdditionalProperties` should have higher priority than logic in PR.

jonaslagoni:
If you dont explicitly define `properties` and `items`, it defaults to `additionalProperties` and additionalItems` which allow everything. This is why you cannot ignore the keywords in that scenario, cause then the type defined in those is the `best possible type` you have access to.

magicmatatjahu (approved):

PR #1116, opened 2023-01-30
chore: add Next integration example

jonaslagoni:
**Description**
This PR add an example for showing how to integrate Modelina into a Next project.


examples/integrate-with-next/src/pages/index.tsx
```diff
@@ -0,0 +1,76 @@
+import { getGeneratedCode } from '../modelina';
+import Head from 'next/head';
+import styles from '@/styles/Home.module.css';
+
+function HomePage({ models }: any): any {
```
kennethaasan, line 5: any reason why this is any? I always question myself whenever I have to use any :D

jonaslagoni: Nothing but laziness 😆

kennethaasan: I'd like to forbid the use of any in this repo, and only allow exceptions when there's no way around it 😆

jonaslagoni: Thats fair, I just thought it didnt matter much for an example 😆 

Let me change it 👍

kennethaasan: that's true, but now you have a better example 😄

jonaslagoni: That's true 😄

kennethaasan (approved):

jonaslagoni:
@kennethaasan thanks for the review, however in the CI the example actually fails by some weird dependency error... Something to do with node 16 and below..

jonaslagoni:
We could do it as part of `v2` release I guess? Might as well just upgrade now I guess 🤔

kennethaasan:
> We could do it as part of `v2` release I guess? Might as well just upgrade now I guess 🤔

Sounds good to me!

jonaslagoni:
Continued in https://github.com/asyncapi/modelina/pull/1488

PR #1136, opened 2023-02-13
fix!: adds union type when operation.message.oneOf is set

### LINKED ISSUE #1111: Union type when operation.message.oneOf is set is missing
#### Describe the bug
Union type when operation.message.oneOf is set is missing

#### How to Reproduce
```ts
const asyncapiDoc = {
  asyncapi: '2.5.0',
  info: {
    title: 'CloudEvent example',
    version: '1.0.0'
  },
  channels: {
    pet: {
      publish: {
        message: {
          // this should create a union type, but currently does not
          oneOf: [
            {
              $ref: '#/components/messages/Dog'
            },
            {
              $ref: '#/components/messages/Cat'
            }
          ]
        }
      }
    }
  },
  components: {
    messages: {
      Dog: {
        payload: {
          title: 'Dog',
          allOf: [
            {
              $ref: '#/components/schemas/CloudEvent'
            },
            {
              type: 'object',
              properties: {
                type: {
                  title: 'DogType',
                  const: 'Dog'
                }
              }
            }
          ]
        }
      },
      Cat: {
        payload: {
          title: 'Cat',
          allOf: [
            {
              $ref: '#/components/schemas/CloudEvent'
            },
            {
              type: 'object',
              properties: {
                type: {
                  title: 'CatType',
                  const: 'Cat'
                }
              }
            }
          ]
        }
      }
    },
    schemas: {
      CloudEvent: {
        title: 'CloudEvent',
        type: 'object',
        properties: {
          id: {
            type: 'string'
          },
          source: {
            type: 'string',
            format: 'uri-reference'
          },
          specversion: {
            type: 'string',
            default: '1.0',
            examples: ['1.0']
          },
          type: {
            type: 'string'
          },
          dataschema: {
            type: 'string',
            format: 'uri'
          },
          time: {
            type: 'string',
            format: 'date-time'
          }
        },
        required: ['id', 'source', 'specversion', 'type']
      }
    }
  }
};
```

#### Expected behavior
Union type should be made for the operation.message.oneOf

kennethaasan:
**Description**

- adds union type when operation.message.oneOf is set

**Related issue(s)**
Resolves #1111

### DIFF
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -7,10 +7,7 @@ import {
   SchemaInterface as AsyncAPISchemaInterface,
   SchemaV2 as AsyncAPISchema
 } from '@asyncapi/parser';
-import {
-  OperationObject,
-  AsyncAPISchemaObject
-} from '@asyncapi/parser/cjs/spec-types/v2';
+import { AsyncAPISchemaObject } from '@asyncapi/parser/cjs/spec-types/v2';
 import { createDetailedAsyncAPI } from '@asyncapi/parser/cjs/utils';
 import { AbstractInputProcessor } from './AbstractInputProcessor';
 import { JsonSchemaInputProcessor } from './JsonSchemaInputProcessor';
@@ -106,16 +103,13 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
     if (channels.length) {
       for (const channel of doc.channels()) {
         for (const operation of channel.operations()) {
-          const operationJson = operation.json<OperationObject>();
+          const operationMessages = operation.messages();
 
-          if (!operationJson.message) {
-            continue;
-          }
-
-          if ('oneOf' in operationJson.message) {
+          // treat multiple messages as oneOf
+          if (operationMessages.length > 1) {
             const oneOf: AsyncAPISchemaObject[] = [];
 
-            for (const message of operation.messages()) {
+            for (const message of operationMessages) {
               const payload = message.payload();
 
               if (!payload) {
@@ -132,12 +126,10 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
             });
 
             addToInputModel(payload);
-          } else {
-            for (const message of operation.messages()) {
-              const payload = message.payload();
-              if (payload) {
-                addToInputModel(payload);
-              }
+          } else if (operationMessages.length === 1) {
+            const payload = operationMessages[0].payload();
+            if (payload) {
+              addToInputModel(payload);
             }
           }
         }
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -11,10 +11,15 @@ import { InputMetaModel, ProcessorOptions } from '../models';
 import { Logger } from '../utils';
 import { AsyncapiV2Schema } from '../models/AsyncapiV2Schema';
 import { convertToMetaModel } from '../helpers';
-import type {
+import {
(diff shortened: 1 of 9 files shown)


kennethaasan:
@jonaslagoni, please review :)

src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -74,30 +77,80 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
...
+          newCommonModel
+        );
+      }
+    };
+
+    // Go over all the message payloads and convert them to models
+    const channels = doc.channels();
+
+    if (channels.length) {
+      for (const channel of doc.channels()) {
```
jonaslagoni, line 107: In its current setup, Modelina would only generate the message payloads used in operations but not lone survivors under `components.messages`. Is that the intention @kennethaasan ? 🤔

src/processors/AsyncAPIInputProcessor.ts
```diff
@@ -74,30 +77,80 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
...
+
+            addToInputModel(payload);
+          } else {
+            for (const message of operation.messages()) {
+              const payload = message.payload();
+              if (payload) {
+                addToInputModel(payload);
+              }
+            }
+          }
```
jonaslagoni, line 142: I am pretty sure that if `operation.messages()` returns more than one, it means that it's within a list of one or more messages i.e. part of `oneOf` and we can add the models to the union type then. That would simplify this a bit 👍 That is if that's the case of course, have you checked @kennethaasan?

jonaslagoni (changes requested):
Just two comments 🙂 

Also, make sure to add this change under the migration guideline as well 👍

src/processors/AsyncAPIInputProcessor.ts
kennethaasan: `doc.allMessages()` or `operation.messages()` from the parser doesn't return the `oneOf` message in `operation.message.oneOf`. That's why we have to this to include the union type in `operation.message.oneOf`.

src/processors/AsyncAPIInputProcessor.ts
kennethaasan: Good idea. I'll check

kennethaasan: Yes, that worked :)

src/processors/AsyncAPIInputProcessor.ts
jonaslagoni: I wonder if that's not a bug in the parser 🤔 I will check that.

I want to ensure that we generate all the models as before, for all situations otherwise this can quickly become a breaking change. That's why `allMessages()` was a safe choice, cause we didn't have to keep track of all the places where messages could exist 😆 

I think the only thing we are missing is loop through the component messages:
```
for (const message of doc.components().messages()) {
        const payload = message.payload();
        if (payload) {
          addToInputModel(payload);
        }
}
```

kennethaasan: Please check the parser :)

Do you still want that check for `doc.components().messages()` even though channels exist?

jonaslagoni: In order to not make this a breaking change, then yea, we need to include all those messages 🙂

kennethaasan: Done :)

jonaslagoni:
As this is a breaking change, I am gonna change the title to `fix!:` 🙂

jonaslagoni (approved):
LGTM 👍

Issue #1138, opened 2023-02-20
additionalItems should only apply for tuples


jonaslagoni:
#### Describe the bug
Currently, additionalItems are applied even when the items are not a tuple. 

https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.4.2

This means that this:

```
    "tags": {
      "type": "array",
      "items": {
        "$ref": "http://asyncapi.com/definitions/2.6.0/tag.json"
      },
      "uniqueItems": true
    },
```
generates this union type:
```
private _tags?: (Tag | any)[];
```

When it should only be:

```
private _tags?: Tag[];
```

jonaslagoni:
The simple array type above makes sense.

But what about those tuples... Say we have this:
```
    "tags": {
      "type": "array",
      "items": [{type: string}, {type: boolean}],
      "additionalItems": true
    },
```

Would we expect to generate:
```
private _tags?: (string | boolean | any)[];
```

or rather a tuple and an array type in union 🤔
```
private _tags?: [string, boolean] | (string | boolean | any)[];
```

I am leaning towards the last option because the following input is valid:
```
{
  tags: ["teest", true, "test"]
}
```
Same with:
```
{
  tags: ["teest", true, false]
}
```
Same with:
```
{
  tags: ["teest", true]
}
```
And in most languages its either a tuple or array, it cant be partly tuple and partly array. So even though you lose the tuple structure for the array type `(string | boolean | any)[]` it's the best possible type to generate 🤔

jonaslagoni:
Continued with tuple discussion here: https://github.com/asyncapi/modelina/issues/1139

kennethaasan:
I've had issues with differences between draft 7 and draft 2020-12: https://ajv.js.org/json-schema.html#draft-2020-12
But I guess that's a different issue and should be handled by checking draft version

jonaslagoni:
Gonna be solved in `v2` and currently released as part of `2.0.0-next.4`

PR #1140, opened 2023-02-21
fix!: additionalItems being applied for regular arrays

jonaslagoni:
**Description**
When I first created this logic I mistakenly thought `additionalItems` was applied to non-tuple array types i.e. when you use a single entry for `items`. This is not the case.

This PR:
- fixes that implementation and adapts the tests.
- add migration guidelines for how the models change.

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/1138


kennethaasan (approved):
Looks good to me :)

jonaslagoni:
Thanks @kennethaasan 👍

PR #1156, opened 2023-03-14
feat: add basic C++ support

### LINKED ISSUE #953: Add Support for CPP language
#### Reason/Context

I have JSON objects that are consumed/produced by Java and CPP (c++) programs. I would love to be able to generate the model for both of them with a single tool. 

 #### Description

Implement a CPP (c++) class generator

jonaslagoni:
**Description**
This PR adds a basic C++ generator.

**Remaining tasks**
This can either be done after or now.
- [ ] Add enum key to value support function for each enum
- [ ] Add serialization support

**Related issue(s)**
Fixes #953


src/generators/cplusplus/CplusplusConstrainer.ts
```diff
@@ -0,0 +1,71 @@
...
+import { CplusplusOptions } from './CplusplusGenerator';
+
+export const CplusplusDefaultTypeMapping: TypeMapping<
+  CplusplusOptions,
+  CplusplusDependencyManager
+> = {
+  Object({ constrainedModel }): string {
+    //Returning name here because all object models have been split out
+    return `${constrainedModel.name}*`;
+  },
```
beku-epitome, line 18: IIUC the Object should be different from Reference in that it owns the property / field data, instead of referencing another object. In that case the `*` for pointer type should not be here, and the property's type definition must be visible (by including that type's header).

src/generators/cplusplus/CplusplusConstrainer.ts
```diff
@@ -0,0 +1,71 @@
...
+  Reference({ constrainedModel }): string {
+    return `${constrainedModel.name}*`;
```
beku-epitome, line 20: Here there is the additional question of whether or not the reference allows the user to modify the referenced value. If modifications are not intended, we should add the `const` keyword to before the `*`.

src/generators/cplusplus/CplusplusFileGenerator.ts
```diff
@@ -0,0 +1,45 @@
...
+      options || {}
+    );
+    //Filter anything out that have not been successfully generated
+    generatedModels = generatedModels.filter((outputModel) => {
+      return outputModel.modelName !== '';
+    });
+    for (const outputModel of generatedModels) {
+      const filePath = path.resolve(
+        outputDirectory,
+        `${outputModel.modelName}.cpp`
```
beku-epitome, line 35: It would make more sense to make the generated files headers, with *.hpp extension.

jonaslagoni: So we only generate header files, no `.cpp` files? 🤔

src/generators/cplusplus/CplusplusConstrainer.ts
jonaslagoni: I dont see how we can force no modifications here 🤔 

We could potentially add this as an option later down the line if it gets requested?

src/generators/cplusplus/CplusplusFileGenerator.ts
beku-epitome: Yes, we are basically creating an interface for data storage, and interfaces need to be shared between compilation units, that's what headers achieve.

src/generators/cplusplus/CplusplusConstrainer.ts
beku-epitome: If the output were `${constrainedModel.name} const*`, that would mean that whoever is accessing this property, cannot modify it. My question was more about the theoretical model than the C++ output though, how these References are intended to behave.

jonaslagoni: Is that not up to the individual use-case, i.e. not really something we can hardcode but should be conditional based on what the user want?

src/generators/cplusplus/CplusplusFileGenerator.ts
jonaslagoni: Okay, so we can change all the models to be located in header files instead?

jonaslagoni: Alright, switched 👍

examples/generate-cplusplus-models/README.md
```diff
@@ -0,0 +1,17 @@
+# TypeScript Data Models
+
+A basic example of how to use Modelina and output a TypeScript data model.
```
beku-epitome, line 3: Copy paste error, need to replace TypeScript to C++

src/generators/cplusplus/CplusplusConstrainer.ts
```diff
@@ -0,0 +1,71 @@
...
+  },
+  Any(): string {
+    return 'void*';
```
beku-epitome, line 23: I recently learned about [std::any](https://en.cppreference.com/w/cpp/utility/any), let's use that here instead of `void*`. It needs a `'#include <utility>'` dependency.

src/generators/cplusplus/CplusplusConstrainer.ts
```diff
@@ -0,0 +1,71 @@
...
+    dependencyManager.addDependency('#include <tuple>');
+    return `std::tuple<${types.join(', ')}>`;
+  },
+  Array({ constrainedModel, dependencyManager }): string {
+    dependencyManager.addDependency('#include <vector>');
+    return `std::vector<${constrainedModel.valueModel.type}>`;
+  },
+  Enum({ constrainedModel }): string {
+    //Returning name here because all enum models have been split out
+    return `${constrainedModel.name}*`;
```
beku-epitome, line 51: It doesn't make sense to use indirection here, store the enum by value instead. The `*` should be removed.

src/generators/cplusplus/CplusplusConstrainer.ts
beku-epitome: Now I finally understand this, and the desired solution is to have only `${constrainedModel.name}`, without the `*`. I was initially confused by the naming, but here we are creating a reference to a type, not to an object. It is important to note that the referenced type must be defined before the current type is defined.

src/generators/cplusplus/CplusplusConstrainer.ts
jonaslagoni: Okay 👍

src/generators/cplusplus/Constants.ts
```diff
@@ -0,0 +1,139 @@
...
+  'line',
+  'error',
+  'warning',
+  'pragma',
+  'defined',
+  '__has_include',
+  '__has_cpp_attribute',
+  'export',
+  'import',
+  'module'
```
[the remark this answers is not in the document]
jonaslagoni: Alright, removed those you highlighted here.

jonaslagoni:
We have a remaining problem when the struct contains properties with references, it seems that in those cases we have to instead of 
```
std::variant<core_space_schema_space_meta_minus_schema_object, bool> additional_items;
```
do the following:
```
std::variant<core_space_schema_space_meta_minus_schema_object, bool> *additional_items;
```

Does that make sense to you @beku-epitome?

beku-epitome:
> We have a remaining problem when the struct contains properties with references, it seems that in those cases we have to instead of
> 
> ```
> std::variant<core_space_schema_space_meta_minus_schema_object, bool> additional_items;
> ```
> 
> do the following:
> 
> ```
> std::variant<core_space_schema_space_meta_minus_schema_object, bool> *additional_items;
> ```
> 
> Does that make sense @beku-epitome?

No, this is not what we should do, as this way we'd need to do dynamic memory management for trivially copyable types. The proper solution is to add the reference type into the dependencies, so the header file where the reference type is defined gets included into the current header. If there is only one file generated, then the reference type definition must stand before any occurrences of references to it. This also mandates that there is no circular referencing.

jonaslagoni:
Okay we are down to the last problem I think 🤔 

Currently a property is often named the same as the referenced model i.e. look at `category`

```
```

beku-epitome:
@jonaslagoni I believe if you fully qualify the type's namespace, the problem should go away:
```
namespace TestNamespace{
  struct pet_post_body {
    int id;
    TestNamespace::category category;
    std::string name;
    std::vector<std::variant<std::string, std::any>> photo_urls;
    std::vector<std::variant<TestNamespace::tags_item, std::any>> tags;
    TestNamespace::status status;
    std::map<std::string, std::any> additional_properties;
  };
}
```

jonaslagoni:
> SonarCloud Quality Gate failed.    [Quality Gate failed](https://sonarcloud.io/dashboard?id=asyncapi_generator-model-sdk&pullRequest=1156)
> 
> [Bug](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=BUG) [A](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=BUG) [0 Bugs](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=BUG) [Vulnerability](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=VULNERABILITY) [A](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=VULNERABILITY) [0 Vulnerabilities](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=VULNERABILITY) [Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=SECURITY_HOTSPOT) [E](https://sonarcloud.io/project/security_hotspots?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=SECURITY_HOTSPOT) [2 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=SECURITY_HOTSPOT) [Code Smell](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=CODE_SMELL) [A](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=CODE_SMELL) [4 Code Smells](https://sonarcloud.io/project/issues?id=asyncapi_generator-model-sdk&pullRequest=1156&resolved=false&types=CODE_SMELL)
> 
> [No Coverage information](https://sonarcloud.io/component_measures?id=asyncapi_generator-model-sdk&pullRequest=1156) No Coverage information [](https://sonarcloud.io/component_measures?id=asyncapi_generator-model-sdk&pullRequest=1156&metric=new_duplicated_lines_density&view=list) [0.1% Duplication](https://sonarcloud.io/component_measures?id=asyncapi_generator-model-sdk&pullRequest=1156&metric=new_duplicated_lines_density&view=list)

I am fine with these errors 👍

jonaslagoni:
> @jonaslagoni I believe if you fully qualify the type's namespace, the problem should go away:

@beku-epitome worked 👌

jonaslagoni:
hey @beku-epitome just pinging you in-case you missed it ✌️

beku-epitome:
Sorry for the delay, I already wrote a response, but my session got messed up in that window, so it wasn't sent :sweat_smile: Here's what I wanted to say:
Great job so far! My notes on the current output:
1. `additional_properties` field shouldn't be added by default, only if the model contains this flag explicitly.
2. Don't add `std::any` variant to arrays, only have the explicitly defined types in the array, and if only one type is defined for the array, don't use variant at all, just have `std::vector<int> array;` in the simple case.
3. It would be nice to code into C++ which properties are required, and which are optional. The current output tells that everything is required. To make properties optional, their type should be wrapped in `std::optional<int>` (`#include <optional>` dependency is added).

jonaslagoni:
>Sorry for the delay, I already wrote a response, but my session got messed up in that window, so it wasn't sent 😅 Here's what I wanted to say:

No worries, thanks for the review§

> additional_properties field shouldn't be added by default, only if the model contains this flag explicitly.

This is the default behavior of the input that affects this, not the generator, it just does what it's told from the inputs perspective, because in AsyncAPI and JSON Schema draft 7 `additionalProperties: true` it renders the property, if your input is set to having `additionalProperties: false`, the property will no be there 🙂 

> Don't add std::any variant to arrays, only have the explicitly defined types in the array, and if only one type is defined for the array, don't use variant at all, just have std::vector<int> array; in the simple case.

I don't think I fully understand this, can you clarify this?

> It would be nice to code into C++ which properties are required, and which are optional. The current output tells that everything is required. To make properties optional, their type should be wrapped in std::optional<int> (#include <optional> dependency is added). 

This should definitely be part of the default yea, gonna adapt the generator 🙂

beku-epitome:
Regarding the array output, if the input is an array property with multiple types, e.g.
```
{
    "type": "array",
    "items":
    [
        {
            "type": "string"
        },
        {
            "type": "number"
        }
    ]
}
```
then the current output is `std::vector<std::variant<std::string, double, std::any>> array_type;`. I'd like the output to be `std::vector<std::variant<std::string, double>> array_type;` instead.
If the input property is like this:
```
{
    "type": "array",
    "items":
    [
        {
            "type": "number"
        }
    ]
}
```
then the current output is `std::vector<std::variant<double, std::any>> array_type;`. I'd like the output to be `std::vector<double> array_type;` in this case.

jonaslagoni:
@beku-epitome thats a problem with the underlying interpreting functionality, it's fixed as part of `v2`, nothing we need to do on our side 🙂

jonaslagoni:
@beku-epitome alright, adapted the types to include optional ones, please take a look 🙂

docs/usage.md
```diff
@@ -198,3 +199,7 @@ Python is one of the many output languages we support. Check out this [basic exa
 ## Generate Kotlin models
 
 Kotlin is one of the many output languages we support. Check out this [basic example for a live demonstration](../examples/generate-kotlin-models) as well as [how to generate enums](../examples/generate-kotlin-enums) and the following [Kotlin documentation for more advanced use-cases](./languages/Kotlin.md).
+
+## Generate C++ models
+
+Kotlin is one of the many output languages we support. Check out this [basic example for a live demonstration](../examples/generate-cplusplus-models) and the following [C++ documentation for more advanced use-cases](./languages/Cplusplus.md).
```
beku-epitome, line 205: Typo: `Kotlin` -> `C++` or `modern C++`

src/generators/cplusplus/CplusplusConstrainer.ts
```diff
@@ -0,0 +1,110 @@
...
+  type: string,
+  partOfProperty: ConstrainedObjectPropertyModel | undefined,
+  dependencyManager: CplusplusDependencyManager
+): string {
+  if (partOfProperty !== undefined) {
+    dependencyManager.addDependency('#include <optional>');
+    return `std::optional<${type}>`;
+  }
+  return type;
+}
```
beku-epitome, line 22: This doesn't seem to do what I expect it to do. Even if I mark an object's property as required, it still ends up being generated as an optional field.

beku-epitome (changes requested):
I have only two notes, see inline, otherwise looks good :)

src/generators/cplusplus/CplusplusConstrainer.ts
jonaslagoni: A bit too fast with the implementation 😅 Was missing `&& partOfProperty.required !== true`.

✅

jonaslagoni:
✅ @beku-epitome

beku-epitome (approved):
If the whole array trouble is resolved somewhere else, then I'm happy with the result :)

jonaslagoni:
@beku-epitome would you like to become a champion for the c++ generator? You dont necessarily need to code, but it can evolve around issues, discussions and reviews of PRs? 🙂

We have some more information about championing in the docs.

jonaslagoni:
@magicmatatjahu @kennethaasan need a review/approval 👍

beku-epitome:
> @beku-epitome would you like to become a champion for the c++ generator? You dont necessarily need to code, but it can evolve around issues, discussions and reviews of PRs? slightly_smiling_face
> 
> We have some more information about championing in the docs.

If it's a similar level of involvement that was required here I can be of service :)

kennethaasan (approved):

jonaslagoni:
@beku-epitome thanks for all the reviews, really appreciate your time 🙏 

@kennethaasan thanks for the review 💪 

> If it's a similar level of involvement that was required here I can be of service :)

@beku-epitome yep more or less, you can lead the C++ development how you see fit, if a PR targets C++ generator, you can review PRs all on your own 🙂 

If you have ways you want the generator improved, a good way is to create issues for them in a good first issue type of approach, enabling others to pick them up. That way you can lead it without touching typescript 😄 So yea, basically championing the generator 🙂 

If you want to, just add yourself to the [codeowners](https://github.com/asyncapi/modelina/blob/cbb240bdbf7caedc0bc6abc5465946428c63bad5/CODEOWNERS#L49) file for C++, afterwards I'll invite you to the team, and if you want, you can join the AsyncAPI TSC 🙂

jonaslagoni:
@all-contributors please add @beku-epitome for code, review, test

PR #1169, opened 2023-03-21
feat!: handle const and discriminator

### LINKED ISSUE #946: Properties that use const should be initialized with that const value and not be included in the constructor of a model

kennethaasan:
I started on this feature at https://github.com/asyncapi/modelina/pull/958, but had to abandon it because of other priorities. 

**Description**

Properties that use `const` should be initialized with that const value and not be included in the constructor of a model. This functionality is added for TypeScript and Java.

If `discriminator` is set in a schema that other schemas are extending, it will create an enum instead of a string const.
- Test Java: https://github.com/kennethaasan/modelina/blob/const/test/generators/java/JavaGenerator.spec.ts#L389-L421
- Test TypeScript: https://github.com/kennethaasan/modelina/blob/const/test/generators/typescript/TypeScriptGenerator.spec.ts#L691-L803 and https://github.com/kennethaasan/modelina/blob/const/test/generators/typescript/TypeScriptGenerator.spec.ts#L911-L990

**Interpreters**
The interpreters are now reading the `discriminator` property from schemas and saves it to the `interpreterOptions` so that it can be used recursively inside properties. In https://github.com/kennethaasan/modelina/blob/const/src/interpreter/InterpretConst.ts#L29-L31, we now only make the const an enum if discriminator is set. By doing that, we allow const to be a string const if the discriminator is not set. In https://github.com/kennethaasan/modelina/blob/const/src/interpreter/InterpretAllOf.ts#L35-L42, we make sure to pass the discriminator to schemas that are extending the schema with discriminator set. The same is done in https://github.com/kennethaasan/modelina/blob/const/src/interpreter/InterpretOneOfWithProperties.ts#L33-L38 and https://github.com/kennethaasan/modelina/blob/const/src/interpreter/InterpretOneOfWithAllOf.ts#L34-L41. In https://github.com/kennethaasan/modelina/blob/const/src/interpreter/InterpretProperties.ts#L34-L40, it will now pass the `discriminator` to properties if the value of the `discriminator` is the same as the property name so that we can handle it in `interpretConst`.

**MetaModel**
I've added `options` to the `MetaModel` in https://github.com/kennethaasan/modelina/blob/const/src/models/MetaModel.ts#L9. This caused a major diff in this PR because the `MetaModel` was used in a lot of places. By adding an options object, we can easier add things like `default` later. At first, I tried to add it to the `ObjectPropertyModel`, but that didn't work because I couldn't read the `constValue` in the `UnionModel`.

**ConstrainedMetaModel**
When I added the `options` object to the `MetaModel`, I also had to add it to the `ConstrainedMetaModel` because it was extending the `MetaModel`, and we need access to the `constValue` in renderers.

**ConstrainedObjectPropertyModel**
In `ConstrainedObjectPropertyModel`, I added methods related to the `constValue` which we can use in renderers.

**src/generators/java/renderers/ClassRenderer.ts**
In the `ClassRenderer`, I've added a `getConstValue` method which is used to render const values for Java.

**src/generators/typescript/TypeScriptObjectRenderer.ts**
In the `TypeScriptObjectRenderer`, I've added a `getConstValue` method which is used to render const values for TypeScript.

**Related issue(s)**
Resolves #946


kennethaasan:
TODO:
- add tests for Java
- fix duplicates when they have similar names (type and point_type)

jonaslagoni:
@kennethaasan you could split it up and introduce it in the meta model, and interpreter etc first, and then do the generators afterwards. 

Just to keep the pr as small as possible 🙂

kennethaasan:
@jonaslagoni, 🤔 . I can if you prefer, but it's now ready for review.

kennethaasan:
@jonaslagoni, it does. I can try to split up next time. I added some extra tests. Please have a look whenever you have time :)

jonaslagoni (changes requested):
Got a couple of things without even going into the code.

### Should `const` always use enums or values?
I.e. should the following:
```
{
    "type": "object",
    "$id": "LightMeasured",
    "additionalProperties": false,
    "properties": {
        "type": {
            "type": "string",
            "const": "test"
        }
    }
}
```
Generate 
```
private _reservedType?: ReservedType = ReservedType.TEST
// Or
private _reservedType?: string = "test"
```
Or should the user choose? 🤔 

### Weird results

#### Multiple enum values only render 1
```json
{
    "type": "object",
    "$id": "LightMeasured",
    "additionalProperties": false,
    "properties": {
        "type": {
            "const": "MyMessage",
            "allOf": [
                {
                    "$ref": "#/definitions/MyCommonEnums"
                }
            ]
        },
        "type2": {
            "allOf": [
                {
                    "$ref": "#/definitions/MyCommonEnums"
                },
                {
                    "const": "MyMessage2"
                }
            ]
        }
    },
    "definitions": {
        "MyCommonEnums": {
            "enum": ["MyMessage", "MyMessage2"]
        }
    }
}
```
I would expect the above to give me a single enum with two values, where `type` uses the first one, and `type2` uses the second one 🤔 There are two bugs there
1. They both should be referring to the same enum instead of two different ones. This is something that is unrelated to your PR, so ignore that.
2. There are a different interpretation between the `type` and `type2` properties, where `type2` actually does what I would expect, and `type` does not. They are mutually exclusive the two representations. 

Problem number 2 is related to this PR as far as I can tell 🤔 Especially when it comes to your cloudevent example, where the messages uses `DogType` and `CatType` instead of referring to `Pet` 🤔 

#### oneOf/anyOf splits them out 
```json
{
    "type": "object",
    "$id": "LightMeasured",
    "additionalProperties": false,
    "properties": {
        "type": {
            "oneOf": [
                {
                    "$ref": "#/definitions/MyCommonEnums"
                },
                {
                    "const": "MyMessage2"
                }
            ]
        }
    },
    "definitions": {
        "MyCommonEnums": {
            "title": "MyCommonEnums",
            "enum": ["MyMessage", "MyMessage2"]
        }
    }
}
```

I would expect that only `MyCommonEnums` would be used and technically the const ignored because they overlap 🤔

kennethaasan:
Regarding this first one:

if enum is **not** set, and const is set, it **can** generate:

`private _reservedType?: string = "test"`
This is not great for any programming language, because it’s impossible to separate models with a switch if they’re used in a oneOf

`private _reservedType?: "test" = "test"`
This could work, but I’m not sure if many programming languages support this.

`private _reservedType?: ReservedType = ReservedType.TEST`
This will work well with oneOf

---------------------------------------------------------------------------------------------------
if enum is set, and const is set, it should generate:
`private _reservedType?: ReservedType = ReservedType.TEST`

---------------------------------------------------------------------------------------------------
if enum is set, and default is set, it should generate: (out of scope for this PR, but should be supported)
`private _reservedType?: ReservedType = ReservedType.TEST`

jonaslagoni:
>if enum is not set, and const is set, it can generate:
private _reservedType?: string = "test"
This is not great for any programming language, because it’s impossible to separate models with a switch if they’re used in a oneOf. But, perhaps we can render it this way if discriminator is not set?

I think we have to keep in mind that we have 2 scenarios.

1. User uses const as a specific value of an enum (this is your primary use-case as far as I can tell, where oneOf is mixed into the equation) 
2. User uses const together with oneOf or anyOf to form a single enum (`{oneOf: [{const: 'test'}, {const: 'test2'}]}`)
4. User uses const to specify the property can only hold one value, i.e. not part of a larger enum as scenario 1 or 2.

To support all three scenarios, I think we have to move the constant information to be part of the ObjectPropertyModel (same as default in the future) instead of being part of the EnumModel 🤔 

>if enum is set, and const is set, it should generate: 
private _reservedType?: ReservedType = ReservedType.TEST

Agree.

>if enum is set, and default is set, it should generate: (out of scope for this PR, but should be supported)
private _reservedType?: ReservedType = ReservedType.TEST

Agree, but changeable at runtime of course.

kennethaasan:
@jonaslagoni , please check again. I've moved constValue to the MetaModel so that we can also use it in the union.

jonaslagoni:
@kennethaasan I have no idea where to even begin here 😄

Do you mind listing the changes you are proposing, so its easy to follow along in the code?

kennethaasan:
@jonaslagoni, understandable 😄 I tried to make it more clear in the PR description. The change in the `MetaModel` caused a big diff, because it's used in a lot of tests. I can also walk you through it over zoom if you want. Feel free to dm me on Slack.

src/generators/java/renderers/ClassRenderer.ts
```diff
@@ -80,13 +80,35 @@ ${this.indent(this.renderBlock(content, 2))}
...
+  }
 }
 
 export const JAVA_DEFAULT_CLASS_PRESET: ClassPresetType<JavaOptions> = {
   self({ renderer }) {
     return renderer.defaultSelf();
   },
-  property({ property }) {
+  property({ renderer, property }) {
+    const constValue = renderer.getConstValue(property);
```
jonaslagoni, line 106: What I refer to in my review is that `property.const` would be much better and follow how we currently handle the meta models.

jonaslagoni (commented):
Sorry, it took so long to review... Been a bit swamped lately 🙃 

> This caused a major diff in this PR because the MetaModel was used in a lot of places

Yea, we might want to use an object argument sometime in the future instead of regular arguments. But yea, down the road.

Yea, after reviewing the code I don't really like you have to remember where information is within the MetaModel, i.e. whether it's under `.options` or `.`. But on the other hand it is future proof for adding extra information as you said.

Now we are just mixing object arguments (options) and regular arguments, which should not be the case IMO. But if we get an issue for it I think its fine for now 👍 

>At first, I tried to add it to the ObjectPropertyModel, but that didn't work because I couldn't read the constValue in the UnionModel.

Ahh okay!

> In ConstrainedObjectPropertyModel, I added methods related to the constValue which we can use in renderers.

Is there a reason it's now functions, instead of raw values? 🤔 Because functions were one of the reasons the last big refactor took place, it was a horrible user experience when writing presets. I.e. you had to call a function to get a rendered property name instead of just saying `model.propertyName`.

I get the renderer's helper functions, but for the models I am a bit unsure of the benefit 🤔

kennethaasan:
> But if we get an issue for it I think its fine for now
Yes, def agree. I'll create an issue for that once this PR is merged.

kennethaasan:
As discussed on Slack, I've added a constant constrainer which is called in https://github.com/kennethaasan/modelina/blob/const/src/helpers/ConstrainHelpers.ts#L133 and https://github.com/kennethaasan/modelina/blob/const/src/helpers/ConstrainHelpers.ts#L603.

The TS constant constrainer looks like this: https://github.com/kennethaasan/modelina/blob/const/src/generators/typescript/constrainer/ConstantConstrainer.ts

The Java constrainer looks very similar.

jonaslagoni (approved):
I dont think I can find anything specific about the changes, other than there is too many things going on in this PR 😄 For the future try to keep one PR = one feature or change 😅 It really helps on the reviewing part and the speed it can happen 👍 

The only things I see we need to handle after this PR are the following:
- Add related docs for each language for the new constraints (I see we are missing for `enumValue`): https://github.com/asyncapi/modelina/blob/master/docs/constraints/Java.md
- Add issues for the other generators that they should implement the new constrain logic, and discriminator and adapt serialization presets where applicable, this can be done either now or once we release v2.

How does that sound? I'll help you create the issues if you want 🙂

kennethaasan:
@jonaslagoni, thanks! Will do. When working on this PR I didn't realize that this PR would end up this big. An experience to learn from.

PR #1181, opened 2023-03-22
feat: add python json serialization and deserialization preset

### LINKED ISSUE #1130: Add Python JSON serialization preset
> **fair warning**! As a good first issue this is a bit beefy, but it's "well" described and we have quite a comprehensive documentation and reference implementations, and we are always ready to help!

#### Reason/Context
Modelina often dont render the properties how they are expected to be rendered in JSON. Therefore, it needs a preset that adds support for it in, some libraries, or custom implementations, there is no preference here.

For example, I am looking for something similar to a `to_json` function, similar to this:

```python
    def to_json(self):
        to_return = {"version": self.version, "type": self.type}
        image_boxes = {}
        for key, boxes in self.bboxes.items():
            jboxes = []
            for box in boxes:
                jboxes.append(box.__dict__)
            image_boxes[key] = jboxes

        to_return["boundingBoxes"] = image_boxes
        return to_return
```

> Method 5, from https://changsin.medium.com/how-to-serialize-a-class-object-to-json-in-python-849697a0cd3

That will map the class properties to its JSON representation. The same is needed for a `from_json` function. Although I don't have much information here.

#### Description

You can find information about how to add presets here: https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#adding-a-new-preset

And you can read more about presets here: https://github.com/asyncapi/modelina/blob/master/docs/presets.md

See other examples here:
- C# Newtonsoft: https://github.com/asyncapi/modelina/blob/master/docs/languages/Csharp.md#using-newtonsoftjsonnet
- Java Jackson: https://github.com/asyncapi/modelina/blob/master/docs/languages/Java.md#jackson-annotation
- Java custom serializer: https://github.com/asyncapi/modelina/blob/master/docs/languages/Java.md#jackson-annotation
- TypeScript custom serializer: https://github.com/asyncapi/modelina/blob/master/docs/languages/TypeScript.md#generate-marshalling-and-unmarshalling-functions

prayutsu:
**Description**

- Added preset to generate JSON serialization and deserialization methods in python
- Added an example for the same
- Modified documentation for python
- Added test case

Reference for actual serialization and deserialization in python - https://yuchen52.medium.com/serialize-and-deserialize-complex-json-in-python-205ecc636caa

**Related issue(s)**

Resolves #1130


prayutsu:
@jonaslagoni PTAL

jonaslagoni (changes requested):
Looking good overall, just have a few comments 👍 
- Remember to add your example to the main example readme: https://github.com/asyncapi/modelina/blob/master/examples/README.md

There are a few things that the current implementation does not take into account (I don't think I have explained myself well enough in the issue):

Any properties that are of type `ConstrainedDictionaryModel` with a serializationType = `unwrap`, need to be unwrapped into the enclosing JSON object (this applies to `additionalProperties` in this case). This means that any properties added to the map of additionalProperties (I don't know python syntax that well, so might be some mistakes in the code):
```python
 p1 = Root({
  email: 'test',
  additionalProperties: { email2: 'test2'}
})
print(p1.serializeToJson) 
//Print {email: "test", email2: "test2"} and not { email: 'test', additionalProperties: { email2: 'test2'}}
```

Furthermore, properties names are constrained to whatever the language supports, this means that while the representation is `email` it might have been a completely different name that should be used for serialization ([read more about it here](https://github.com/asyncapi/modelina/blob/master/docs/constraints.md)). instead of the original property name, you need to use the property `unconstrainedPropertyName` when serializing. 

Some JSON libraries support this through annotation, or maybe you need to write more custom code in the serialization method itself. Also, keep in mind that if one of the properties is a ConstrainedReferenceModel it needs to call the same function again on the other class.

It might be that these features are not supported by the native `json` library in python, if they are not, we can still use this as long as we document those restrictions in the documentation 🙂

jonaslagoni:
@prayutsu no, in your example it would be:
```
{
  unconstrainedProperty1: 'test',
  unconstrainedProperty2: '12323',
  name: {
    first_name: 'John',
    last_name: 'Doe'
  }
}
```

The thought process here is the following:

The class property `additionalProperties` is in this case of type ConstrainedDictionaryModel, with serializationType = unwrap. This means any values within that property should be unwrapped when serializing the class. 

If it was serializationType = normal, it should be serialized as (if the class property `additionalProperties`'s `unconstrainedPropertyName` is `unconstrainedProperty2`):
```
{
  unconstrainedProperty1: 'test',
  unconstrainedProperty2: '12323',
  unconstrainedProperty3: {
    name: {
      first_name: 'John',
      last_name: 'Doe'
    }
  }
}
```
In any way, properties defined within a dictionary will always be as is 🙂 

It might also be that `email` should be rendered as `email`, but regardless that value still be accessible through `unconstrainedPropertyName`.

Dont hesitate to ask further clarifications, it's not an easy topic 😅

prayutsu:
> @prayutsu no, in your example it would be:
> 
> ```
> {
>   unconstrainedProperty1: 'test',
>   unconstrainedProperty2: '12323',
>   name: {
>     first_name: 'John',
>     last_name: 'Doe'
>   }
> }
> ```
> 
> The thought process here is the following:
> 
> The class property `additionalProperties` is in this case of type ConstrainedDictionaryModel, with serializationType = unwrap. This means any values within that property should be unwrapped when serializing the class.
> 
> If it was serializationType = normal, it should be serialized as (if the class property `additionalProperties`'s `unconstrainedPropertyName` is `unconstrainedProperty2`):
> 
> ```
> {
>   unconstrainedProperty1: 'test',
>   unconstrainedProperty2: '12323',
>   unconstrainedProperty3: {
>     name: {
>       first_name: 'John',
>       last_name: 'Doe'
>     }
>   }
> }
> ```
> 
> In any way, properties defined within a dictionary will always be as is 🙂
> 
> It might also be that `email` should be rendered as `email`, but regardless that value still be accessible through `unconstrainedPropertyName`.
> 
> Dont hesitate to ask further clarifications, it's not an easy topic 😅

I couldn't find any reference to implement this with `json` library (or any other library).

leigh-johnson:
Hey @prayutsu, thank you for all the hard work on this PR. :bowing_woman: It's awesome to see more Pythonistas contributing to Modelina!

I tried out this PR against some of my real-world AsyncAPI schemas since those always help me grok what's going on inside Modelina's generators. I noticed that enum serialization raises an AttributeError:

```
o = mappingproxy({'_generate_next_value_': <function Enum._generate_next_value_ at 0x7f1358ab9090>, '__module__': 'printna...tup'>, 'SHUTDOWN': <OctoPrintServerStatus.SHUTDOWN: 'Shutdown'>, '__new__': <function Enum.__new__ at 0x7f1358ab9000>})

>   return json.dumps(self.__dict__, default=lambda o: o.__dict__, indent=2)
E   AttributeError: 'mappingproxy' object has no attribute '__dict__'
```

Here's the relevant schema object ([full schema here](https://github.com/bitsy-ai/printnanny-asyncapi-schema/blob/main/2.4.0/printnanny-octoprint.yml#L116), if you'd like more context)
```
OctoPrintServerStatusChanged:
  name: OctoPrintServerStatusChanged
  summary: "OctoPrint server status changed"
  payload:
    x-parser-schema-id: OctoPrintServerStatusChanged
    additionalProperties: false
    properties:
      status:
        $ref: "#/components/schemas/OctoPrintServerStatus"
    required:
      - status
```

Generated model code:
```python
from enum import Enum
class OctoPrintServerStatus(Enum): 
  STARTUP = "Startup"
  SHUTDOWN = "Shutdown"
```

I'll follow up with some ideas on how to handle this in my next post.

leigh-johnson:
IMO, the most Pythonic way to handle this is to extend `json.JSONEncder`

1. For the pydanic preset, pydantic implements a `model.json(...)` method

2. Without the pydanic preset, `json.dumps()` accepts a `cls=<subclass of json.JSONEncder>` keyword-argument. 

Here's a super-exhaustive example of #2, with my approach for coercing various numpy types, enums to string, and datetime to an ISO 8601 strings in another Python project. 
```
import json
from enum import Enum
import numpy as np
import datetime
from io import BytesIO
import base64
import PIL

class NumpyEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, np.integer):
            return int(obj)
        elif isinstance(obj, np.floating):
            return float(obj)
        elif isinstance(obj, np.ndarray):
            return obj.tolist()
        elif isinstance(obj, (datetime.date, datetime.datetime)):
            return obj.isoformat()
        elif isinstance(obj, BytesIO):
            obj.seek(0)
            obj = obj.read()
            return base64.b64encode(obj).decode()
        elif isinstance(obj, bytes):
            return obj.decode()
        elif isinstance(obj, PIL.Image.Image):
            buffered = BytesIO()
            obj.save(buffered, format="JPEG")
            return base64.b64encode(buffered.getvalue()).decode()
        elif isinstance(obj, Enum):
            return obj.value
        return json.JSONEncoder.default(self, obj)
```

How about something like this for Modelina?
```
import json
from enum import Enum

class ModelinaEncoder(json.JSONEncoder):
    """
        A JSON encoder intended for use with Modelina-generated models
    """
    def default(self, obj):
        if hasattr(obj,'toDict'):
            return obj.toDict()
        elif isinstance(obj, Enum):
            return obj.value
        return json.JSONEncoder.default(self, obj)

class ModelinaBaseClass:
    """
        A base class to be used in src/generator/python/renderers/ClassRenderer.ts
    """

    def toDict(self):
        return dict(<render field=self.field here>)
    
    def toJson(self):
        return json.dumps(self.toDict(), cls=ModelinaEncoder)
```

prayutsu:
@leigh-johnson @jonaslagoni Just an FYI, I'll get back to this PR after a few days as I am currently focusing on writing a GSoC proposal.

jonaslagoni:
@prayutsu any update? 🙂 Do you have any questions?

prayutsu:
@jonaslagoni will update docs to include the limitation

examples/python-generate-json-serializer-and-deserializer/index.spec.ts
```diff
@@ -0,0 +1,15 @@
+const spy = jest.spyOn(global.console, 'log').mockImplementation(() => {
+  return;
+});
+import { generate } from './index';
+
+describe('Should be able to render TEMPLATE', () => {
```
jonaslagoni, line 6:
```suggestion
describe('Should be able to render JSON serialization and deserialization functions', () => {
```

jonaslagoni (changes requested):
Sorry for the long wait @prayutsu! 

Didn't catch that it was done, remember to use the `re-request review` button that lets me know I need to review it again 🙂  
image

Only have one comment other than that it's ready to be merged 👍

prayutsu:
@jonaslagoni Made the changes. Will keep in mind to re-request a review in the future.

jonaslagoni (changes requested):
You will need to update the snapshots as well 🙂 `npm run test:examples:update`

jonaslagoni (approved):
LGTM 👍

jonaslagoni:
@all-contributors please add @prayutsu for code, example, test, docs

PR #1191, opened 2023-03-26
feat: add option to Csharp generator allowing user to generate records instead of classes

### LINKED ISSUE #1153: Add support for records with the C# generator through an option
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
  - Records is a popular way of defining DTO's in C#
- How will this change help?
  - It gives more flexibility for C# users
- What is the motivation?
  - Records are a great match for DTO's in C#:
    - Reference based records are immutable by default using short declaration, ex: `public record Car(int Doors, string Manufacturer);`
    - Made for containing data, not for logic
    - They use value based comparison

#### Description
Please try answering few of those questions
- What changes have to be introduced?
- Will this be a breaking change?
- How could it be implemented/designed?

henrikjon:
**Description**
This PR adds an option to the C# generator allowing it to render records instead of classes. This is done using a new renderer that implements `self`, `property` and `additionalContent` hooks. It does not support using the `autoimplementedProperties` as option as this is instead a default when generating records.

**Related issue(s)**
Resolves #1153

### DIFF
--- a/docs/languages/Csharp.md
+++ b/docs/languages/Csharp.md
@@ -85,7 +85,12 @@ Check out this [example for a live demonstration](../../examples/csharp-overwrit
 
 If you want the generated models to inherit from a custom class, you can overwrite the existing rendering behavior and create your own class setup.
 
-Check out this [example for a live demonstration](../../examples/csharp-use-inheritance).
+## Generate models as records 
+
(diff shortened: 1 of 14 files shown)


jonaslagoni (approved):
LGTM 🔥

jonaslagoni (approved):

jonaslagoni:
@all-contributors please add @henrikjon for code, test, doc, example

PR #1194, opened 2023-03-27
chore: Improve experience when using docker

Ferror:
Hi! I noticed a couple of things that made me crazy when I was trying to build a PHP generator. Each time I changed something I had to rebuild the whole docker image, where in reality we don't need that. By `COPY ./ /` docker image was not optimal.

So I added the docker-compose with the mounted whole repository as volume and tweaked the npm scripts to use the docker-compose config. Super simple and it improves performance. Now the Dockerfile is a base image.

Another thing is the docker pipeline. I noticed we don't test if changes made by users are not breaking the Docker Image, and open source has free pipelines... so profit and profit 😄


jonaslagoni:
Gonna fix it in https://github.com/asyncapi/modelina/pull/1198

You are gonna have the same problems in https://github.com/asyncapi/modelina/pull/1196 until the PR is merged.

jonaslagoni (approved):
Dont use docker much for local development, but looks like a solid suggestion 👍

jonaslagoni:
Gonna need to wait for release workflow to be done before I can merge this.

derberg (commented):
this new docker image. I can't figure out from the workflow, is the idea to push it somewhere?

also, what "~" refers too?

Ferror:
> this new docker image. I can't figure out from the workflow, is the idea to push it somewhere?
> 
> also, what "~" refers too?

You need to be more specific. What file do you refer to? If that's the pipeline the `~` refers to nothing. With each push the pipeline will be executed

jonaslagoni:
> this new docker image. I can't figure out from the workflow, is the idea to push it somewhere?

Just to ensure that it can be compiled 🙂

Contributors use this docker image to run the black-box tests, in order to not install all the dependencies themself.

jonaslagoni:
@all-contributors please add @Ferror for infra

Issue #1195, opened 2023-03-27
PHP Generator Scoped


Ferror:
## Milestones
- [x] Base PHP Generator from the official template (https://github.com/asyncapi/modelina/pull/1196)
- [x] Define PHP reserved words/keywords (https://github.com/asyncapi/modelina/pull/1211)
- [x] Ability to generate classes
- [x] Ability to generate Enums
- [x] Mixed & Nullable Types (https://github.com/asyncapi/modelina/pull/1216)
- [x] Documentation
- [x] Examples (https://github.com/asyncapi/modelina/pull/1218)
- [x] PHP Namespace as a parameter (https://github.com/asyncapi/modelina/pull/1258)
- [x] Render strict types (https://github.com/asyncapi/modelina/pull/1253)
- [ ] Final classes by default and parameter

### Additionals
* How about Backed Enums? https://www.php.net/manual/en/language.enumerations.backed.php

## Thoughts, Ideas, PHP Specifics 

> Let's support only PHP 8.0 and up. PHP 7.4 is already not supported: https://www.php.net/supported-versions.php

### PHP 8.0
* private properties when readonly -> getters
* public properties instead of getters + setters

### PHP 8.1
* public readonly properties
* public properties instead of getters + setters

### PHP 8.2
* public readonly class
* public properties

## Options
* PHP Version (8.0, 8.1, 8.2) default (8.2)
* Readonly properties (true/false) default (true)
* Namespace

PR #1196, opened 2023-03-27
feat: php generator from official template

Ferror:
The first step of scoped PHP generator. Creating empty generator from template.

https://github.com/asyncapi/modelina/issues/1195


jonaslagoni (changes requested):
You have some linter errors @Ferror.

Do you want the PHP generator released in its current form or is it still a bit WIP? 

Remember, once merged into master and part of v1, you must uphold https://github.com/asyncapi/modelina#versioning-and-maintenance 🙂 

If you want to continue the development a bit more, we could create a temporary branch for you to work on for a bit of time. It will mean it won't be released, but then you make any change you wish while it's work in progress.

jonaslagoni:
> @jonaslagoni I want to iterate over the generator, but I don't want to keep it in one PR. However, you want to make it right. Long-living PRs do not suite me well

I think we should keep this in a `php_generator` branch in Modelinas side, and progress there 🙂 That means you can make any type of change without having to worry about breaking anything.

And then we can slowly progress the generator until you feel like it's ready to be released 🙂 

How does that sound?

Ferror:
>
> How does that sound?

Awsome. Let's go!

jonaslagoni (approved):

PR #1197, opened 2023-03-27
fix: typescript unions not being serialized if referenced model

jonaslagoni:
**Description**
This PR fixes a bug where a union, array of union, and arrays is not correctly serialized i.e. the `marshal` function of a referenced class is not called, and the instance is instead serialized without respecting the underlying `unconstrainedPropertyName`.

There is also a problem with the unmarshalling... Because at the moment the classes are not correctly created, but I will have to fix that later 🙃


kennethaasan (dismissed):

jonaslagoni:
Just realized that I am not taking multiple types into account such as arrays with unions, regular arrays, regular unions, etc... Gonna add some more 😓

jonaslagoni:
The sonarcloud reporting has nothing to do with the changes introduced here. Dont know why it all of the sudden reports those 🤨

Samridhi-98 (approved):
LGTM 🎉

jonaslagoni:
Thanks @Samridhi-98 🙇 

/rtm

PR #1211, opened 2023-04-02
feat: add PHP reserved keywords

Ferror:
Define a list of reserved php keywords. Source https://www.php.net/manual/en/reserved.keywords.php


jonaslagoni (approved):
👍

Ferror:
> 👍

That was fast 🦸

PR #1205, opened 2023-04-01
feat: add PHP generator

jonaslagoni:
**Description**
This PR is a placeholder for when to release the PHP generator.

Resolves #1195 
Resolves https://github.com/asyncapi/modelina/issues/744

PR #1214, opened 2023-04-02
feat: add correct PHP enum and class rendering

Ferror:
* Define PHP types
* Define Enum and Class present renderers
* Publish snapshots -> Not ideal, but seems to be looking decent


jonaslagoni (approved):
👍 

I am skipping over smaller things, because you either are gonna change it in the future or we do it in the final review #1205 🙂

PR #1216, opened 2023-04-03
feat: php generator - nullable && mixed types

Ferror:
PHP does not have a dictionary or any other nice types. So the `mixed` should be applied. The nullable parameters are specified with `?` mark. ex. `?string`


jonaslagoni (approved):
👍

PR #1253, opened 2023-04-23
feat: php generator - descriptions, rendering model vs rendering file, example of complete model

Ferror:
1. Added the Generate Complete Models Example
2. Moved the rendering `<?php` tag to rendering complete model
3. Rendering class descriptions
4. Rendering Strict Types declaration as default


jonaslagoni (approved):
LGTM 👍

PR #1258, opened 2023-04-25
feat: php generator - rename packageName to namespace

Ferror:
PHP has the concept of namespaces (like C#) instead of packages (like Java) so to have it coherent I renamed it


jonaslagoni (approved):
👍

PR #1269, opened 2023-05-04
refactor: adds discriminator in CommonModel and interpreters

kennethaasan:
This is a preparation PR to support union for objects in Java (https://github.com/asyncapi/modelina/pull/1264). It adds `discriminator` in the CommonModel, and the interpreters set that value when they should.

**Description**

- adds `discriminator` in the CommonModel
- sets `discriminator` in the interpreters

**Related issue(s)**

PR #1264, opened 2023-05-02
feat!: java interfaced union generation for oneOf

dlkj:
This is a first end to end draft.  The first priority is to agree a way forward for representing `allOf` in Java. Any feedback on the direction of the implementation, coding style and generated code would be appreciated.

**Description**

- Generate a Java interface for unions
  - Common property accessors are defined as part of the interface
- All classes that are part of unions implement the union interface
- Add union support to the Jackson annotation preset
  - Support using a `discriminator` property or deducing union members via property names

**Example output - java-generate-union-discriminator-jackson**

Schema:
```json
{
   "$schema":"http://json-schema.org/draft-07/schema#",
   "title":"Vehicle",
   "type":"object",
   "discriminator":"vehicleType",
   "oneOf":[
      {
         "$ref":"#/definitions/Car"
      },
      {
         "$ref":"#/definitions/Truck"
      }
   ],
   "definitions":{
      "Car":{
         "title":"Car",
         "type":"object",
         "properties":{
            "vehicleType":{
               "type":"string"
            },
            "name":{
               "type":"string"
            }
         }
      },
      "Truck":{
         "title":"Truck",
         "type":"object",
         "properties":{
            "vehicleType":{
               "type":"string"
            },
            "name":{
               "type":"string"
            }
         }
      }
   }
}
```

Java output:
```java
@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.PROPERTY, property="vehicleType")
@JsonSubTypes({
  @JsonSubTypes.Type(value = Car.class, name = "Car"),
  @JsonSubTypes.Type(value = Truck.class, name = "Truck")
})
/**
 * Vehicle represents a union of types: Car, Truck
 */
public interface Vehicle {
  String getVehicleType();
  void setVehicleType(String vehicleType);
  String getName();
  void setName(String name);
  Map<String, Object> getAdditionalProperties();
  void setAdditionalProperties(Map<String, Object> additionalProperties);
}

public class Car implements Vehicle {
  @JsonProperty("vehicleType")
  private String vehicleType = "Car";
  @JsonProperty("name")
  private String name;
  private Map<String, Object> additionalProperties;

  public String getVehicleType() { return this.vehicleType; }
  public void setVehicleType(String vehicleType) { this.vehicleType = vehicleType; }

  public String getName() { return this.name; }
  public void setName(String name) { this.name = name; }

  public Map<String, Object> getAdditionalProperties() { return this.additionalProperties; }
[...]

PR #1270, opened 2023-05-04
refactor: adds discriminator in MetaModel and ConstrainedMetaModel

kennethaasan:
This is a preparation PR to support union for objects in Java (https://github.com/asyncapi/modelina/pull/1264) and continuation of https://github.com/asyncapi/modelina/pull/1269.

**Description**

- adds `discriminator` in the `MetaModel` and the `ConstrainedMetaModel `
- sets discriminator type in `constrainUnionModel`. Do we need a constrainer for every language for this, or can we keep it like this?

**Related issue(s)**

### DIFF
--- a/src/models/ConstrainedMetaModel.ts
+++ b/src/models/ConstrainedMetaModel.ts
@@ -40,6 +40,12 @@ export abstract class ConstrainedMetaModel extends MetaModel {
         ...options.const
       };
     }
+
+    if (options.discriminator) {
+      this.options.discriminator = {
+        ...options.discriminator
+      };
+    }
   }
 
   /**
--- a/src/models/ConstrainedMetaModel.ts
+++ b/src/models/ConstrainedMetaModel.ts
@@ -2,7 +2,8 @@ import { makeUnique } from '../helpers/DependencyHelpers';
 import {
   MetaModel,
   MetaModelOptions,
-  MetaModelOptionsConst
+  MetaModelOptionsConst,
+  MetaModelOptionsDiscriminator
 } from './MetaModel';
 
 export interface ConstrainedMetaModelOptionsConst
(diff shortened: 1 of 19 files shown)


src/models/ConstrainedMetaModel.ts
```diff
@@ -2,16 +2,23 @@ import { makeUnique } from '../helpers/DependencyHelpers';
...
 }
 
+export interface ConstrainedMetaModelOptionsDiscriminator
+  extends MetaModelOptionsDiscriminator {
+  type?: string;
+}
+
 export interface ConstrainedMetaModelOptions extends MetaModelOptions {
   const?: ConstrainedMetaModelOptionsConst;
+  discriminator?: ConstrainedMetaModelOptionsDiscriminator;
```
jonaslagoni, line 21: Dont think we need to constrain discriminator, as the value will always be the same. I.e. it does not change from language to language.

src/models/MetaModel.ts
```diff
@@ -2,8 +2,13 @@ export interface MetaModelOptionsConst {
   originalInput: unknown;
 }
 
+export interface MetaModelOptionsDiscriminator {
+  originalInput: string;
```
jonaslagoni, line 6: What does the `originalInput` refer to here? 🤔 Just the value from the input or?

src/models/ConstrainedMetaModel.ts
kennethaasan: As you can see in `addDiscriminatorTypeToUnionModel`, I want to calculate if the discriminator property has a common type. I will use this property in the Union Model (interface) for Java if it exists. I can move this logic to the Java Union Renderer, but this seems language-agnostic to me.

src/models/MetaModel.ts
kennethaasan: Yeah, just the value from the input. Perhaps there is a better name for it. Just call it discriminator or value?

jonaslagoni: I think that is better yea 🤔

src/models/ConstrainedMetaModel.ts
jonaslagoni: Hmmm, I dont think I fully understand how you are gonna use it, because you are taking a different direction then I would 😄 

I would have added a discriminator value to each model in the union, that the serializer can iterate to check which one is the appropriate model to pick.

But lets just go forward with your approach, I am sure it falls into place when the generators change 😄

src/models/MetaModel.ts
kennethaasan: Changed it to `discriminator`

PR #1271, opened 2023-05-04
feat!: adds interface for oneOf objects for Java

kennethaasan:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
+++ b/test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
@@ -1,9 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`JavaGenerator CloudEvent handle allOf with const in CloudEvent type 1`] = `
+exports[`JavaGenerator oneOf/discriminator with jackson preset handle allOf with const in CloudEvent type 1`] = `
 Array [
-  "",
-  "public class Dog {
+  "@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.PROPERTY, property=\\"type\\")
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = Dog.class, name = \\"Dog\\"),
+  @JsonSubTypes.Type(value = Cat.class, name = \\"Cat\\")
+})
+/**
+ * Pet represents a union of types: Dog, Cat
+ */
+public interface Pet {
+  CloudEventType getType();
+}",
+  "public class Dog implements Pet {
   @NotNull
   @JsonProperty(\\"id\\")
   private String id;
@@ -118,7 +128,7 @@ Array [
     return String.valueOf(value);
   }
 }",
-  "public class Cat {
+  "public class Cat implements Pet {
   @NotNull
   @JsonProperty(\\"id\\")
   private String id;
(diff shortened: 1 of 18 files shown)


docs/migrations/version-1-to-2.md
```diff
@@ -161,3 +161,102 @@ data class Response(
...
+        type: object
+        discriminator: vehicleType
+        properties:
+          vehicleType:
+            title: VehicleType
+            type: string
+        required:
+          - vehicleType
+        oneOf:
+          - $ref: '#/components/schemas/Cat'
```
dlkj, line 184:
```suggestion
          - $ref: '#/components/schemas/Car'
```

dlkj:
#1264 had some examples, can we bring some of those across as well? Happy to split into a separate PR.

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
```diff
@@ -298,6 +308,178 @@ Array [
...
+public interface Pet {
+  
+}",
+  "public class Dog implements Pet {
+  @NotNull
+  @JsonProperty(\\"type\\")
+  private final DogType type = DogType.DOG;
+  private Map<String, Object> additionalProperties;
+
+  public DogType getType() { return this.type; }
```
dlkj, line 330: nit: this should have an `@Override` annotation to be consistent with other generated code bellow. Not functionally required, and we should consider not doing this as the implementation is likely to be a bit fiddly.

dlkj: ...oh, this isn't exposed via the interface, so it isn't overridden.

I would expect this to be an enum `PetType` then exposed via a getter in the interface. This is likely to interact interestingly with the const code.

dlkj: the test "JavaGenerator oneOf/discriminator with jackson preset handle allOf with const in CloudEvent type" uses a common type for the disriminator, so I think we have a mild conflict between the `union` and `const` behavour

dlkj: Idiomatically in java you would make a constant by declaring the underlying field as `final`

kennethaasan:
> I cant find any tests where you expect union without Jackson preset, is that looking like expected?

@jonaslagoni, I added a test for that in test/generators/java/JavaGenerator.spec.ts

jonaslagoni (commented):
LGTM 👍 

@dlkj wanna give a final review?

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
dlkj: OK, after discussion, I'm happy that this is something the user is in full control of. It is possible to write a schema that breaks things in the java code gen, but the users isn't guided down that path.

dlkj:
LGMT. Great work!

kennethaasan:
Appreciate the quick review! @dlkj deserves most of the credit for making https://github.com/asyncapi/modelina/pull/1264. Feel free to close that one. If you want the common properties available in the interface, please submit a new PR.

jonaslagoni:
@all-contributors please add @dlkj for code, test, example, review

Issue #1263, opened 2023-04-28
C# Enum - cannot convert from 'Newtonsoft.Json.Linq.JToken' to 'string?'


swordmaster2k:
#### Describe the bug
Attempting to use the preset `CSHARP_NEWTONSOFT_SERIALIZER_PRESET` with the `CSharpGenerator` and an AsyncAPI Spec that contains an object that references a string enum results in broken code inside of the deserializer `ReadJson`.

It generates code with the error `cannot convert from 'Newtonsoft.Json.Linq.JToken' to 'string?'`.

Generating CSharp code from a valid spec along the lines of:
```yaml
  schemas:
    userSignedUpPayload:
      type: object
      properties:
        firstName:
          type: string
          description: "foo"
        lastName:
          type: string
          description: "bar"
        email:
          type: string
          format: email
          description: "baz"
        createdAt:
          type: string
          format: date-time
        type:
          $ref : '#/components/schemas/userType'
    
    userType:
      type: string
      enum:
        - Guest
        - Admin
      default: Admin
```

Results in a model that will not compile due to the error `error CS1503: Argument 1: cannot convert from 'Newtonsoft.Json.Linq.JToken' to 'string?'` at the code line:
```csharp
  if(jo["type"] != null) {
    value.Type = UserTypeExtensions.ToUserType(jo["type"]);
  }
```

This is because it is trying to implicitly convert from 'Newtonsoft.Json.Linq.JToken' to 'string?' when calling the generated `ToUserType` on the enum:
```csharp
    public static UserType? ToUserType(string? value)
    {
      switch (value)
      {
        case "Guest": return UserType.GUEST;
        case "Admin": return UserType.ADMIN;
      }
      return null;
    }
```

#### How to Reproduce

1. Create a `CSharpGenerator` or `CSharpFileGenerator` configured with the `CSHARP_NEWTONSOFT_SERIALIZER_PRESET`.
2. Create an AsyncAPI that contains a string enum.
3. Generate the resulting models with `ts-node index.ts`.
4. Observe the broken code in `ReadJson` of `UserSignedUpPayloadConverter` in the `UserSignedUpPayload.cs` file.

**index.ts**
```typescript
import {
    CSharpFileGenerator,
    CSHARP_NEWTONSOFT_SERIALIZER_PRESET
} from '@asyncapi/modelina';

const generator = new CSharpFileGenerator({
    presets: [CSHARP_NEWTONSOFT_SERIALIZER_PRESET]
});

[...]

PR #1272, opened 2023-05-05
chore: add initial Java runtime testing

jonaslagoni:
**Description**
This is the initial runtime setup (for Java only) I was thinking about that can be replicated across all generators to ensure that the generated models are semantically correct by adding tests in Java, using the generated models.

What this PR includes:
- Workflow that runs the Java runtime tests only when related files changes.
- A self-contained Java project with a simple model generation and tests to see it working.

What this enables down the line:
- When we are happy with the manually created tests, we could choose to include them in the generation phase: https://github.com/asyncapi/modelina/pull/839
- This avoids us from releasing new versions that break the promise of the models (i.e. issues such as https://github.com/asyncapi/modelina/issues/1263)
- Reducing the number of generator snapshot tests to focus on the runtime tests. 

**Related issue(s)**
The first step for https://github.com/asyncapi/modelina/issues/612


jonaslagoni:
@kennethaasan would love to get your initial feedback and suggestions for this setup, I think this is the way to go instead of all those snapshot tests in TypeScript.

What do you think?

test/runtime/runtime-java/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+          <!-- It will execute command "npm build" inside "/src/main/javascript" 
+            directory to clean and create "/dist" directory -->
+          <execution>
+            <id>npm build</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <configuration>
+              <arguments>run build</arguments>
```
jonaslagoni, line 109: Actually pretty cool that this plugin could be used to install the AsyncAPI CLI and then generate the models with Modelina for Java projects 😄

kennethaasan:
Awesome stuff. I'll have a look at it next week.

dlkj:
The java stuff looks good to me. Great stuff.

jonaslagoni:
Thanks for the reviews! 🔥

Issue #1273, opened 2023-05-05
Kotlin generator makes all properties required


micro-jumbo:
#### Describe the bug
Kotlin generator makes all properties required, even those which are not in `required` list in Async API definition.

#### How to Reproduce
For given component:
```json
      "Success": {
        "type": "object",
        "title": "Success",
        "properties": {
          "result": {
            "title": "SuccessMessageResult",
            "type": "string",
            "const": "SUCCESS"
          },
          "message": {
            "type": "string",
            "title": "SuccessMessageMessage"
          }
        },
        "required": [
          "result"
        ],
        "additionalProperties": false
      },
```

generated Kotlin code is following:
```kotlin
data class SuccessMessage(
  val result: SuccessMessageResult,
  val message: String
)
```

which makes `message` a required, non-nullable field.

#### Expected behavior
Generate kotlin class with nullable property if propety is not required in AsyncAPI definition

LouisXhaferi:
Thanks for the issue. Probably a good idea to change this.

LouisXhaferi:
One note though: Since this fix is going to change the output of the renderer, it is considered a breaking change. Breaking changes are handled differently from "normal" changes.

@jonaslagoni could you pls elaborate on how and where to commit this fix?

jonaslagoni:
I commented on the PR 🙂

PR #1274, opened 2023-05-05
fix: update default Kotlin renderer to make non-required properties nullable

micro-jumbo:
**Description**

- Update default Kotlin renderer to make non-required properties nullable and `null` by default.
- Updated snapshots
-  
**Related issue(s)**
Fixes#1273


LouisXhaferi:
This should fix #1273

LouisXhaferi:
lgtm apart from the failing check.

I'd still want to wait for feedback from @jonaslagoni on how to handle breaking changes.

jonaslagoni:
I would definitely target `next`, cause if a consumer of a model uses them with the expectation of properties being required it would break their code (even though this is the output🙂)

micro-jumbo:
@LouisXhaferi @jonaslagoni 
Thanks guys for your review. I pushed additional change to fix the linter issue. Is there anything more I should do to get it merged?

jonaslagoni:
You need to retarget the PR towards `next`, I don't think you will get any issues, but not entirely sure, and then mark the PR as breaking i.e. `fix!:` instead and then add an entry in https://github.com/asyncapi/modelina/blob/next/docs/migrations/version-1-to-2.md that it breaks for Kotlin 🙂

micro-jumbo:
Due to conflicting files after retargeting to `next` closing this one in favour of #1277

PR #1277, opened 2023-05-08
fix!: update default Kotlin renderer to make non-required properties nullable

### LINKED ISSUE #1273: Kotlin generator makes all properties required

micro-jumbo:
**Description**

- Update default Kotlin renderer to make non-required properties nullable and `null` by default.
- Updated snapshots

**Related issue(s)**
Fixes #1273
Retarget of #1274

### DIFF
--- a/docs/migrations/version-1-to-2.md
+++ b/docs/migrations/version-1-to-2.md
@@ -135,3 +135,29 @@ schemas:
 ```
 
 This example will generate a model for `Cat` and `Dog` where `PetType` is a shared enum that contains two values (`Cat` and `Dog`). `PetType` will not have a setter, and will automatically be initialized.
+
+## Optional properties in Kotlin
+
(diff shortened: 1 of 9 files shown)


jonaslagoni (approved):
LGTM 👍

PR #1290, opened 2023-05-15
feat!: use EXISTING_PROPERTY as JsonTypeInfo.As annotation to avoid duplicates when serializing

kennethaasan:
**Description**

- Use EXISTING_PROPERTY as JsonTypeInfo.As annotation to avoid duplicates when serializing (ref https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonTypeInfo.As.html)

**Related issue(s)**

### DIFF
--- a/docs/migrations/version-1-to-2.md
+++ b/docs/migrations/version-1-to-2.md
@@ -199,7 +199,7 @@ components:
 will generate
 
 ```java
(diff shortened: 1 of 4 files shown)


kennethaasan:
@dlkj, please review.

jonaslagoni (approved):

jonaslagoni:
@kennethaasan just realized we need an entry in the migration Document 😄

PR #1292, opened 2023-05-15
chore(website): expose examples on the website

jonaslagoni:
**Description**
This PR exposes (nearly) all the examples we have on the modelina website.

It works by reading all the directories in the examples folder, and each example has some Modelina code which is read from the `index.ts` file, and then the output code is read from the jest snapshots.

The overview in `/examples` is created from https://github.com/asyncapi/modelina/blob/master/examples/README.md with a few modifications such as changing links etc, but otherwise it's the same.

Here is what this feature enables:
- Show simple examples that contain modelina code and then the output.
- Each example link is shareable through the query parameter `/examples?selectedExample=adapting-input-and-output`
- Automatically updated on each website build (handles adding new examples, changes, etc.) with minimum maintenance as possible.

image

image


modelina-website/scripts/build-examples.js
```diff
@@ -0,0 +1,130 @@
...
+  if(exampleName.includes('dart')) {
+    return 'dart';
+  }
+
+  if(exampleName.includes('go')) {
+    return 'go';
+  }
+
+  return 'typescript';
+}
```
magicmatatjahu, line 56: Make some mapping 😆

magicmatatjahu (commented):
Design/UX review:

- each links on the navigation should have `cursor-pointer` class to have "pointer" cursor icon when someone will hover it
- it will be nice if code and output will be rendered in this same line, not like current, which has some "margin" gap

  image

- readme should be rendered at center :)

  

I can accept but it will be good if you can propagate my suggestions :)

modelina-website/scripts/build-examples.js
jonaslagoni: Fine xD

jonaslagoni:
>each links on the navigation should have cursor-pointer class to have "pointer" cursor icon when someone will hover it

Makes sense 👍 

> it will be nice if code and output will be rendered in this same line, not like current, which has some "margin" gap

Let me try 👍 

> readme should be rendered at center :)

Good point 😄

jonaslagoni:
@magicmatatjahu adapted everything 🙂

magicmatatjahu (approved):
LGTM! :)

PR #1296, opened 2023-05-18
feat!: add java oneOf union support for properties of models 

dlkj:
**Description**

- add type safe support for the properties of Java models to be represented as an interface with types that implement that interface
- if the union includes a primitive type or collection the union will be represented as an Object, as is the case for all union properties at the moment

**Related issue(s)**

### DIFF
--- a/src/generators/java/JavaConstrainer.ts
+++ b/src/generators/java/JavaConstrainer.ts
@@ -79,7 +79,8 @@ export const JavaDefaultTypeMapping: JavaTypeMapping = {
     return constrainedModel.name;
   },
   Reference({ constrainedModel }): string {
-    return constrainedModel.name;
+    //propagate the type mapping of the referenced model
+    return constrainedModel.ref.type;
   },
   Any(): string {
     return 'Object';
@@ -170,9 +171,35 @@ export const JavaDefaultTypeMapping: JavaTypeMapping = {
 
     return uniqueTypes[0];
   },
-  Union(): string {
-    //Because Java have no notion of unions (and no custom implementation), we have to render it as any value.
-    return 'Object';
+  Union({ constrainedModel }): string {
+    //We don't have a good solution for unions including primitive types and
+    //collection types, so they are represented by the 'Object' type
+    if (
+      constrainedModel.union.find(
+        (u) =>
+          //Primitive types and boxed versions
+          u.type.toLowerCase() === 'byte' ||
+          u.type.toLowerCase() === 'short' ||
+          u.type === 'int' ||
+          u.type === 'Integer' ||
+          u.type.toLowerCase() === 'long' ||
+          u.type.toLowerCase() === 'float' ||
+          u.type.toLowerCase() === 'double' ||
+          u.type === 'char' ||
+          u.type === 'Character' ||
+          u.type.toLowerCase() === 'boolean' ||
+          u.type.toLowerCase() === 'int' ||
+          //Collections
+          u.type.startsWith('Map<') ||
+          u.type.startsWith('List<') ||
+          u.type.startsWith('Set<') ||
+          u.type === 'String' ||
+          u.type.endsWith('[]')
+      )
+    ) {
+      return 'Object';
+    }
+    return constrainedModel.type;
   },
   Dictionary({ constrainedModel }): string {
     //Limitations to Java is that maps cannot have specific value types...
--- a/src/generators/java/JavaConstrainer.ts
+++ b/src/generators/java/JavaConstrainer.ts
@@ -1,5 +1,5 @@
 import { Constraints } from '../../helpers';
-import { ConstrainedEnumValueModel } from '../../models';
+import { ConstrainedEnumModel, ConstrainedEnumValueModel } from '../../models';
 import {
   defaultEnumKeyConstraints,
   defaultEnumValueConstraints
@@ -80,7 +80,12 @@ export const JavaDefaultTypeMapping: JavaTypeMapping = {
   },
   Reference({ constrainedModel }): string {
     //propagate the type mapping of the referenced model
-    return constrainedModel.ref.type;
+    if (constrainedModel.ref instanceof ConstrainedEnumModel) {
+      //use the enum name rather than the underlying type of the Enum
+      return constrainedModel.ref.name;
+    } else {
+      return constrainedModel.ref.type;
+    }
   },
   Any(): string {
     return 'Object';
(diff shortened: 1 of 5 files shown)


dlkj:
Found the issues. My type mapping code assumed that references wouldn't reference a schema that included the currently being mapped schema. This isn't the case for the `JsonSchema` hyperschema examples. It's a bit unusual, but valid.

test/generators/java/JavaConstrainer.spec.ts
```diff
@@ -33,14 +33,22 @@ describe('JavaConstrainer', () => {
     });
   });
   describe('Reference', () => {
-    test('should render the constrained name as type', () => {
```
dlkj, line 36: Whoops! Didn't intended to erase this test. Needs adding back in.

dlkj: Reinstated by aa73f35bae5c86fcb70322b45393c9341baa71fb

dlkj:
@jonaslagoni - somehow this PR didn't get any reviewers added to it. Can you add yourself and @kennethaasan (and anyone else you see fit) as reviewers?

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
```diff
@@ -583,7 +842,26 @@ exports[`JavaGenerator should render \`class\` type 1`] = `
...
   public Map<String, Object> getAdditionalProperties() { return this.additionalProperties; }
   public void setAdditionalProperties(Map<String, Object> additionalProperties) { this.additionalProperties = additionalProperties; }
-}"
+}",
+  "/**
+ * Members represents a union of types: String, Double, Boolean
+ */
+public interface Members {
+  
+}",
```
dlkj, line 851: These "stub" types representing unions that we aren't using aren't great. I can't find an idiomatic way to suppress them and they would be used if we extended the `oneOf` support to include built-in `Java` types in the future. They don't do much harm but they represent a bit of a loose end for this feature.

src/generators/java/JavaConstrainer.ts
```diff
@@ -74,11 +74,48 @@ const interpretUnionValueType = (types: string[]): string => {
...
+      u.type === 'Object' ||
+      u.type === 'String' ||
+      //Collections
+      u.type.startsWith('Map<') ||
+      u.type.startsWith('List<') ||
+      u.type.startsWith('Set<') ||
+      u.type.endsWith('[]')
+  );
+  return builtInType !== undefined;
+}
```
kennethaasan, line 104: what if you change this to check for only object models instead? You can use this function in `JavaGenerator.renderUnion` to not render the interface to avoid rendering useless interfaces. 

```ts
export function unionIncludesBuiltInTypes(
  model: ConstrainedUnionModel
): boolean {
  return !model.union.every(
    (union) =>
      union instanceof ConstrainedObjectModel ||
      (union instanceof ConstrainedReferenceModel &&
        union.ref instanceof ConstrainedObjectModel)
  );
}
```

test/generators/java/JavaGenerator.spec.ts
```diff
@@ -342,6 +341,74 @@ describe('JavaGenerator', () => {
       expect(models.map((model) => model.result)).toMatchSnapshot();
     });
 
+    test('should create an interface for child models', async () => {
```
kennethaasan, line 344: perhaps it's enough to only have a test with Jackson preset?

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
jonaslagoni: I am guessing that the limits to the Java generator at the moment is simple types cannot be represented in unions i.e. string + boolean model, right? i.e. only one or more classes

src/generators/java/JavaConstrainer.ts
dlkj: Thanks. Hopefully will find the time to test out these changes sometime today/this week.

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
dlkj: Java can't include built in types in a new interface. We could wrap these types in a new class but that isn't great either. My plan is to get this all working for user defined classes and fall back to `Object` for built ins. If we want to have the option of using built in types, this could be a follow on PR.

test/generators/java/JavaGenerator.spec.ts
dlkj: Sure. Probably overkill on the testing scenarios.

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
jonaslagoni: Just remember that even though unions are a non-thing in Java, it's still something that users can define within inputs, and because of that we must represent it in the best possible way.

> We could wrap these types in a new class but that isn't great either

I have a feeling this is exactly the best solution that encompasses all use-cases, and is rather simple to achieve. But lets see 😄

dlkj: I'm prioritising composable smaller PRs over one giant PR. I don't have an issues with doing the type wrapping, but I'm trying to keep each PR to a manageable size.

test/generators/java/JavaGenerator.spec.ts
dlkj: Removed

src/generators/java/JavaConstrainer.ts
dlkj: Swapped out `unionIncludesBuiltInTypes` implementation to use modelina types. Will look at supressing the interface generation next.

dlkj: Interfaces for non `ConstrainedObjectModel` unions are now supressed.

dlkj:
@jonaslagoni, @kennethaasan - I think most of your comments have been addressed.

"Runtime testing Java models" is failing in CI, but this failure is also present on `next`.

I'd recommend we open an issue to do a follow-up feature to add wrapper classes for built-in type so that we can have full generalised union support in Java.

FYI I'm going to be on holiday from 2023-06-02 to the 2023-06-12 so I'm unlikely to pick up any further review comments until then.

kennethaasan (approved):

kennethaasan:
Great! Let's merge then. Enjoy your holidays!

PR #1302, opened 2023-05-20
feat: php generator - description preset example

Ferror:
PS. Let me know if that was what we were missing :)


jonaslagoni (changes requested):
Remember to add this example to the list of examples here: https://github.com/asyncapi/modelina/blob/master/examples/README.md

Other then that it looks nice 👍

jonaslagoni (approved):

PR #1310, opened 2023-05-22
fix: add missing exports for Rust

jonaslagoni:
**Description**
It's impossible to correctly overwrite the constraints because the helper function and types are not accessible. This PR exports them as all the other generators does.

Blocks #1309


leigh-johnson (approved):
👍

PR #1356, opened 2023-05-31
feat: add nullable mode for C#

JFCote:
This is my first PR for this project. I must say that I'm not a Typescript pro and I don't have a working setup on my desktop.
I hope one of the maintainer can help me finish and complete this PR! I will help as much as I can to have it merged and released as soon as possible :)

**Description**

- Added via a `handleNullable` param to the C# generator that is defaulted to "false" to not create breaking changes
- Add code to prevent warning when compiling the C# classes when the project is "nullable" in C# .NET

For autoImplementedProperties, it will look like that when a property is not required:

`public string Id { get; set; } = null!;` instead of `public string Id { get; set; }`

Without autoImplementedProperties, it will look like this:

`private string _id = null!;` instead of `private string _id;`

**Related issue(s)**

No issues were created for this, I went strictly to code because it was very simple to implement.

### DIFF
--- a/src/generators/csharp/renderers/ClassRenderer.ts
+++ b/src/generators/csharp/renderers/ClassRenderer.ts
@@ -103,7 +103,7 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
 
     let nullablePropertyEnding = '';
     if (options?.handleNullable && !property.required) {
-      nullablePropertyEnding = '= null!';
+      nullablePropertyEnding = ' = null!';
     }
 
     if (options?.autoImplementedProperties) {
@@ -113,9 +113,9 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
       const semiColon = nullablePropertyEnding !== '' ? ';' : '';
       return `public ${property.property.type} ${pascalCase(
         property.propertyName
-      )} { ${getter} ${setter} } ${nullablePropertyEnding}${semiColon}`;
+      )} { ${getter} ${setter} }${nullablePropertyEnding}${semiColon}`;
     }
-    return `private ${property.property.type} ${property.propertyName} ${nullablePropertyEnding};`;
+    return `private ${property.property.type} ${property.propertyName}${nullablePropertyEnding};`;
   },
   async accessor({ renderer, options, property }) {
     const formattedAccessorName = pascalCase(property.propertyName);
--- a/src/generators/csharp/renderers/ClassRenderer.ts
+++ b/src/generators/csharp/renderers/ClassRenderer.ts
@@ -110,9 +110,10 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
       const getter = await renderer.runGetterPreset(property);
       const setter = await renderer.runSetterPreset(property);
 
+      const semiColon = nullablePropertyEnding !== '' ? ';' : '';
       return `public ${property.property.type} ${pascalCase(
         property.propertyName
-      )} { ${getter} ${setter} } ${nullablePropertyEnding}`;
+      )} { ${getter} ${setter} } ${nullablePropertyEnding}${semiColon}`;
     }
     return `private ${property.property.type} ${property.propertyName} ${nullablePropertyEnding};`;
   },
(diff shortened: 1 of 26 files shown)


src/generators/csharp/renderers/ClassRenderer.ts
```diff
@@ -100,14 +100,22 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
     return renderer.defaultSelf();
   },
   async property({ renderer, property, options }) {
+
+    let nullablePropertyEnding = '';
+    if (options?.handleNullable && !property.required) {
+      nullablePropertyEnding = '= null!';
```
jonaslagoni, line 106: I would change this to
```suggestion
      nullablePropertyEnding = ' = null!';
```

src/generators/csharp/renderers/ClassRenderer.ts
```diff
@@ -100,14 +100,22 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
...
+
     if (options?.autoImplementedProperties) {
       const getter = await renderer.runGetterPreset(property);
       const setter = await renderer.runSetterPreset(property);
+
+      const semiColon = nullablePropertyEnding !== '' ? ';' : '';
       return `public ${property.property.type} ${pascalCase(
         property.propertyName
-      )} { ${getter} ${setter} }`;
+      )} { ${getter} ${setter} } ${nullablePropertyEnding}${semiColon}`;
```
jonaslagoni, line 116: I would change this to
```suggestion
      return `public ${property.property.type} ${pascalCase(
        property.propertyName
      )} { ${getter} ${setter} }${nullablePropertyEnding};`;
```

src/generators/csharp/renderers/ClassRenderer.ts
```diff
@@ -100,14 +100,22 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
...
     }
-    return `private ${property.property.type} ${property.propertyName};`;
+    return `private ${property.property.type} ${property.propertyName} ${nullablePropertyEnding};`;
```
jonaslagoni, line 118: And then you can do this 😄 
```suggestion
    return `private ${property.property.type} ${property.propertyName}${nullablePropertyEnding};`;
```

jonaslagoni (changes requested):
Thanks for taking the time to fix this 🙏 

Do you want this to be part of v2 instead (just create the change on the `next` branch)? then you don't have to introduce the option `handleNullable` if this should always be the preferred output 🙂 

If not, then you are still missing some tests that show it works with the new option 😄

JFCote:
> Do you want this to be part of v2 instead (just create the change on the `next` branch)? then you don't have to introduce the option `handleNullable` if this should always be the preferred output 🙂

Well it's good that it is an option because in C# it is an option in each project. In .NET 6.0 it default to nullable = true but anyone can put it to false. So I think it's good that it is an option for the users :)

As for the tests, I completely forgot this part, I'm not very used to typescript. I will try to add them. Thanks!

src/generators/csharp/renderers/ClassRenderer.ts
JFCote: Fixed in latest commit

src/generators/csharp/renderers/ClassRenderer.ts
JFCote: fixed in latest commit

src/generators/csharp/renderers/ClassRenderer.ts
JFCote: fixed in latest commit

jonaslagoni:
Forgot to mention we actually adding features pretty well documented here: https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#new-features 😄 

Remember to also add the following:
- Please add a section in the C# readme about this feature: https://github.com/asyncapi/modelina/blob/master/docs/languages/Csharp.md
- Please add an example show-casing this new feature: https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#adding-examples

JFCote:
@jonaslagoni I just added everything that was asked for (I think!). Let me know if something is missing and I will add it!

jonaslagoni (commented):
Awesome, nicely done for your first contribution 🔥 

Interested in helping out with other stuff? 😄 

Giving @ron-debajyoti a day to respond and then we merge 👍

JFCote:
@jonaslagoni My team and I will start using asyncapi seriously in a very big project so we will probably have other contributions coming your way. It might not be all in modelina because we are also creating custom template for asyncapi generator so it might me all over the place in the asyncapi community! :)
Thanks for the very good support and ownership of the project!

jonaslagoni (approved):

jonaslagoni:
@JFCote ahh, missed that one 😅 

No worries, I'll fix it 🙂

PR #1364, opened 2023-06-02
test: add runtime testing for C++

### LINKED ISSUE #1335: Add runtime testing for C++
#### Reason/Context
In https://github.com/asyncapi/modelina/pull/1272 we added the initial runtime setup for Java, now it's time to replicate it for C++.

The reason we add these is to ensure that the code that is generated is semantically and syntactically correct, and the only way to achieve this is by testing the models at runtime.

This initial setup should be as simple as possible, just a C++ project, with a single test and a single generated model, and that's it.

Here is the general task list:

- [ ] Create a C++ project under; `test/runtime/runtime-cplusplus`
- [ ] Add a simple test code generation script here: `test/runtime/runtime-cplusplus.ts`. Check out the Java variant as an example https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-java.ts
- [ ] Add the generation script to the package file: https://github.com/asyncapi/modelina/blob/ba0bd41e9803394f56872c5ec92f8e335a4d687a/package.json#L98
- [ ] Run the generation script
- [ ] Manually add a simple test for the generated models, could be as simple as just creating an instance of it.
- [ ] Add a simple Jest script, which just execute the C++ test: https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-java.spec.ts
- [ ] Add a simple script that will run the test: https://github.com/asyncapi/modelina/blob/ba0bd41e9803394f56872c5ec92f8e335a4d687a/package.json#L97
- [ ] Create a CI workflow that runs the runtime tests: https://github.com/asyncapi/modelina/blob/master/.github/workflows/runtime-java-testing.yml

And that's it 🎉

codingmickey:
**Description**

- Added the runtime testing for C++ of the models that are generated
- There are still errors and pending tasks from the issue

**Related issue(s)**

Closes #1335


jonaslagoni:
Are you stuck on anything @codingmickey or is there just some stuff that needs to be ironed out?

codingmickey:
I've completed this one @jonaslagoni whenever you get time you or anyone can review it! 😉

codingmickey:
I've completed this one @jonaslagoni whenever you get the time you or anyone can review it! 😉

test/runtime/runtime-cplusplus/src/AddressTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    json jsonData;
+    jsonData["streetName"] = address.street_name;
+    jsonData["houseNumber"] = address.house_number;
+    jsonData["marriage"] = address.marriage.value();
+    jsonData["members"] = std::get<double>(address.members.value());
+    jsonData["arrayType"] = stringVector;
+    jsonData["nestedObject"] = {{"test", address.nested_object.value().test.value()}};
+
+    std::string jsonStr = jsonData.dump();
```
jonaslagoni, line 45: This entire code should actually be part of the generated models themself, cause right now it's your own implementation of the serialization that you are testing.

This does not test the generated output in the slightest 🙂

jonaslagoni (changes requested):
Thanks for pushing through with this 💪 

Please see the comment in the review, I think a good first test for C++ (since there are no serialization support generated for the models (yet), is just to create an instance of the model and that's it, just pure and simple.

Is it impossible to add a package manager or C++ project setup for the runtime testing? I would really love to not be relying on downloaded libraries, cause it's unmaintainable in the long run.

codingmickey:
> Please see the comment in the review, I think a good first test for C++ (since there are no serialization support generated for the models (yet), is just to create an instance of the model and that's it, just pure and simple

Oh 🥲
I thought it's making like others, I read this in docs but should've asked before implementing...

> Is it impossible to add a package manager or C++ project setup for the runtime testing? I would really love to not be relying on downloaded libraries, cause it's unmaintainable in the long run.

Sure I'll check into that!

test/runtime/runtime-cplusplus/src/AddressTest.cpp
codingmickey: Now that you've mentioned it. I kind of feel dumb when rereading the issue.

So I should just create a object of that class in single test?

Also if needed I can help with serialization of c++ modules to json as I've already looked this much into it with this issue

codingmickey: @jonaslagoni any inputs?

jonaslagoni: Sorry, must have missed this question 🙇 

> So I should just create a object of that class in single test?

Yes, make it easy first test. It might feel useless, however, see this PR as the starting point and setup for C++ runtime testing, and not so much about creating a whole set of tests 🙂 

If you want, you are more then welcome to do it afterwards, just keep each PR as small as possible 🙂

codingmickey:
I've tried to include the libraries as a git submodule but will have to configure it in the GH actions too.. Also cleaned the code a bit

codingmickey:
Sorry for the confusion at first regarding the issue.. And making such a lot of commits😬😢 Hope this will lead to good runtime testing in the future for C++.

Happy to help anytime!

jonaslagoni (approved):
Not the biggest fan of git-submodules if it can be avoided (rather want a package manager), but I think this is fine for now and a great start 👍 

Awesome job 👍

Issue #780, opened 2022-06-29
Modelina config file


jonaslagoni:
#### Reason/Context
This is probably one of the most important features to this date, because this will enable better integrations across tools. Because while Modelina is highly customizable, it is almost impossible for CLIs and other tools to offer those advantages because many of the configs are callbacks where you can overwrite logic. Therefore I think it would make sense to add configuration file support where the customizations can be placed `.modelina.js`

This means that users can customize their models outside the scope of any integration, for example:
- Our AsyncAPI CLI where users could pass a config file such as `asyncapi generate models --config ./.modelina.js`
- Our AsyncAPI generator, if templates use Modelina to generate the data models you could pass alongside your own extended configuration file for the templates to use. I.e. `asyncapi generate fromTemplate asyncapi.json @asyncapi/ts-nats-template --param modelinaConfig=./.modelina.js`.

These config files should be simple JS files that expose a default js object which is taken as the options for the generator i.e.

```js
const {TS_COMMON_PRESET} = require('@asyncapi/modelina');
module.exports = {
  presets: [
    {
      preset: TS_COMMON_PRESET,
      options: {
        example: true
      }
    }
  ]
}
```

jonaslagoni:
I cant figure out if this feature should be solved in Modelina or the respectful integrations such as the generator and CLI. Made a proof of concept for the CLI to trigger the discussion in the next community meeting.

https://github.com/asyncapi/cli/pull/430

kaushik-rishi:
@jonaslagoni I would be very much interested to take this issue up.
I totally understand the requirement for this issue.

jonaslagoni:
Sure!

So I actually managed to get it working in the CLI, check out this PR: https://github.com/asyncapi/cli/pull/430

However, we need to move the logic to Modelina instead and add better error handling, tests, etc.

We have to figure out how to change the API of Modelina i.e. propose how the generator methods would look like when providing the options as a file path.

Furthermore, because we are working with files (i.e. something a website cannot handle) we have to adapt the [file generators](https://github.com/asyncapi/modelina/blob/master/src/generators/AbstractFileGenerator.ts) and not the regular generator.

I would suggest you start small, create a single method that takes the configuration as a file path, load the configuration (i.e. import the file), and call the other generator functions with the right options.

Let me know if this makes sense and give you a starting point 👍

kaushik-rishi:
[...]

PR #1367, opened 2023-06-03
feat: load modelina configuration options from a config file

kaushik-rishi:
**Description**
- WIP PR
**Related issue(s)**
- #780


kaushik-rishi:
@jonaslagoni I've used the `require` function to import the object exported by the config file.
Post build I've observed that the `loadConfigFile` function is only present in `/lib/cjs` and not `lib/esm`.

Why is it that way ? :thinking: 

Or, do you want me to import the module like this ? 
```js
function readJSFile(filePath) {
  try {
    const fileContent = fs.readFileSync(filePath, 'utf8');
    const moduleExports = {};
    const moduleWrapper = eval('(function(exports) {' + fileContent + '})');
    moduleWrapper(moduleExports);
    return moduleExports;
  } catch (error) {
    console.error('Error reading JS file:', error);
    return null;
  }
}
```

jonaslagoni:
> Post build I've observed that the loadConfigFile function is only present in /lib/cjs and not lib/esm.

Which build command did you run? https://github.com/asyncapi/modelina/blob/c0e0268996126ca65caeaa6d4c3d4fbcef1767bc/package.json#L78

kaushik-rishi:
> Which build command did you run?

On running `build:esm`, i do find the function `loadConfigFile`.
But previously i think i ran `build` command, i didn't find the function `loadConfigFile` in the directory `/lib/esm`, irrespective it's working now. :slightly_smiling_face: 

Thanks. :smile: 

Do you think we should add more documentation on setting up the repository for development ?

kaushik-rishi:
@jonaslagoni After i add a function, and build the project (`build:esm`).
How do i test the imports and stuff ? 
Like let's say i create a new project somewhere else in my Desktop, how do i import the newly build version of this project :thinking:  ?

src/helpers/FileHelpers.ts
```diff
@@ -64,4 +64,18 @@ export class FileHelpers {
       }
     });
   }
+
+  static async getConfigFromFile(configPath: string): Promise<any> {
```
jonaslagoni, line 68: You have to write some unit tests here 🙂 

Make sure you test error states as well:
- What if the file exist but is not valid JS?
- What if the file exist but is TypeScript?
- What if the file exist and is correct JS, but does not expose the right config?
- What if the file exist and is correct JS, but does not expose the config file the right way?

examples/get-config-from-file/index.ts
```diff
@@ -0,0 +1,29 @@
...
+      type: 'string',
+      format: 'email'
+    }
+  }
+};
+
+async function main() {
+  try {
+    const config: { config: TypeScriptOptions } =
+      await FileHelpers.getConfigFromFile('./modelina.config.alternate.js');
```
jonaslagoni, line 18: I think we should simplify this.

```
    const config = await FileHelpers.getConfigFromFile('./modelina.config.alternate.js');
```

Instead of having to do something like `config.config`, what do you think? 🤔

src/helpers/FileHelpers.ts
jonaslagoni: Make sure you document this function as well i.e. what it does 🙂

jonaslagoni (changes requested):
Generally, you need to add documentation as well, probably something like the following:
- Add a section here: https://github.com/asyncapi/modelina/blob/master/docs/advanced.md
- Add your example to the list of examples: https://github.com/asyncapi/modelina/blob/master/examples/README.md

Hope this gives you a direction, otherwise let me know if you want anything clarified 🙂

examples/get-config-from-file/index.ts
kaushik-rishi: Yeah definitely, it looks absurd.
We can directly export the config. :slightly_smiling_face: 
Would that work ?

src/helpers/FileHelpers.ts
kaushik-rishi: @jonaslagoni 
I'll make sure to document and write the unit tests.

> What if the file exist and is correct JS, but does not expose the right config?

and 

> What if the file exist and is correct JS, but does not expose the config file the right way?

Should be the same thing right ? I mean we wouldn't know if the file is exposing the config the right way, unless we know what it's exposing ? :thinking: 

The way i put it here might be confusing, feel free to ignore this, i will let you know later more context on this. :slightly_smiling_face:

kaushik-rishi: I'll try fixing things one after other and contact you back :smile:

kaushik-rishi: @jonaslagoni 

We can restrict the users to only allow them to use js file for config right ? i hardly saw places where people used ts files for config.
What do you say ?

kaushik-rishi:
@jonaslagoni 

The config file can be placed anywhere on the users machine, which is probably not a node package and doensn't have modelina installed.

Out `getConfigFromFile` function will be called let's say from `cli`, where modelina will be installed, so the `require` should work right ?

kaushik-rishi:
I will pause working on this PR for 2-3 days, busy with some other things

src/helpers/FileHelpers.ts
jonaslagoni: Agree 👍 

It can also be an improvement down the line if ever requested 🙂

jonaslagoni:
Looks like you are nearly there @kaushik-rishi 🤔

jonaslagoni:
Hey @kaushik-rishi, want to continue or can someone else take over the progress?

jonaslagoni:
Closing as no response for over a month

PR #1369, opened 2023-06-03
chore(website): add namespace options for PHP, C++ and C#

### LINKED ISSUE #1351: Add namespace option for playground
#### Reason/Context
In some generators, it's possible to define `namespace` as an option, and it should be possible to change the value in the playground.

It applies to the following generators:
- CSharpGenerator
- CplusplusGenerator
- PhpGenerator

You can read more about how the playground works here: https://github.com/asyncapi/modelina/blob/master/modelina-website/README.md#playground

Shurtu-gal:
**Description**
- Added namespace support for PHP, C# and C++

**Related issue(s)**
Fixes #1351

### DIFF
--- a/modelina-website/src/components/playground/Playground.tsx
+++ b/modelina-website/src/components/playground/Playground.tsx
@@ -87,11 +87,14 @@ class Playground extends React.Component<
     tsIncludeJsonBinPack: false,
     csharpArrayType: 'Array',
     csharpAutoImplemented: false,
-    phpIncludeDescriptions: false,
     csharpOverwriteHashcode: false,
     csharpIncludeJson: false,
     csharpOverwriteEqual: false,
     csharpIncludeNewtonsoft: false,
+    csharpNamespace: '',
+    phpIncludeDescriptions: false,
+    phpNamespace: '',
+    cplusplusNamespace: '',
   };
   hasLoadedQuery: boolean = false;
   constructor(props: ModelinaPlaygroundProps) {
@@ -258,6 +261,9 @@ class Playground extends React.Component<
       this.config.phpIncludeDescriptions =
         query.phpIncludeDescriptions === 'true';
     }
+    if (query.phpNamespace !== undefined) {
+      this.config.phpNamespace = query.phpNamespace;
+    }
     if (query.csharpIncludeJson !== undefined) {
       this.config.csharpIncludeJson =
         query.csharpIncludeJson === 'true';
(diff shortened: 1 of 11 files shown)


modelina-website/src/components/playground/options/CSharpGeneratorOptions.tsx
```diff
@@ -63,12 +64,32 @@ class CSharpGeneratorOptions extends React.Component<
...
+            <span className="mt-1 max-w-2xl text-sm text-gray-500">
+              Namespace
+            </span>
+            <input
+              type="text"
+              className="form-input rounded-md border-gray-300 cursor-pointer font-regular text-md text-gray-700 hover:bg-gray-50 focus-within:text-gray-900"
+              name="csharpNamespace"
+              value={this.context?.csharpNamespace}
+              onChange={this.onChangeNamespace}
+            />
```
jonaslagoni, line 90: We need to do something about each value change triggers a re-generation, I think we should debounch with a delay, so it's not until they are "done" writing that the models re-generate 🤔 

Whad you do you think @Shurtu-gal?

modelina-website/src/components/playground/Playground.tsx
```diff
@@ -87,11 +87,14 @@ class Playground extends React.Component<
     tsIncludeJsonBinPack: false,
     csharpArrayType: 'Array',
     csharpAutoImplemented: false,
-    phpIncludeDescriptions: false,
     csharpOverwriteHashcode: false,
     csharpIncludeJson: false,
     csharpOverwriteEqual: false,
     csharpIncludeNewtonsoft: false,
+    csharpNamespace: '',
```
jonaslagoni, line 94: Maybe we should default the value here as well 🤔 That way in the options its the default value you edit instead of an empty field 🤔

jonaslagoni (changes requested):
A few comments ✌️ 

Looking good otherwise 👍

modelina-website/src/components/playground/Playground.tsx
Shurtu-gal: Sure

modelina-website/src/components/playground/options/CSharpGeneratorOptions.tsx
Shurtu-gal: Okay

Shurtu-gal:
@jonaslagoni Requested changes are done.

jonaslagoni (approved):
Remember to use `chore(website)` in the PR title for all website-related changes, otherwise, it would release a new version of Modelina itself, which is not the intended behavior 😄 

Nicely done!

PR #1370, opened 2023-06-04
feat: php generator - Runtime Testing

Ferror:


.github/workflows/runtime-php-testing.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Build library
+        run: npm install && npm run build:prod
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@2.25.2
+        with:
+          php-version: ${{ matrix.php }}
+
+      - name: Generate Models
+        run: npm run generate:runtime:php
```
Ferror, line 41: @jonaslagoni  isn't that part missing in Java and Rust pipelines?

test/runtime/runtime-php.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { execCommand } from '../blackbox/utils/Utils';
+import path from 'path';
+
+jest.setTimeout(50000);
+
+test('PHP runtime testing', async () => {
+  const compileCommand = `cd ${path.resolve(
+    __dirname,
+    './runtime-php'
+  )} && composer install --no-interaction --quiet && composer test`;
```
Ferror, line 10: There is one problem with this command, as PHPUNIT does always output information and its taken by execCommand as error, but it isnt

.github/workflows/runtime-php-testing.yml
jonaslagoni: Java generates it when the project is build: https://github.com/asyncapi/modelina/blob/c0e0268996126ca65caeaa6d4c3d4fbcef1767bc/test/runtime/runtime-java/pom.xml#L119

I think Rust does the same

jonaslagoni (commented):
Seems like the CI fails to run 🙂

jonaslagoni (approved):
LGTM 👍

PR #1378, opened 2023-06-05
fix: inverted condition for csharp generator handle-nullable

JFCote:
**Description**

- I fixed the condition that was completely inverted. Sorry about that!

**Related issue(s)**

N/A

@jonaslagoni

### DIFF
--- a/examples/csharp-generate-handle-nullable/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-generate-handle-nullable/__snapshots__/index.spec.ts.snap
@@ -4,8 +4,8 @@ exports[`Should be able to render a C# class with null warning removed with hand
 Array [
   "public class Root
 {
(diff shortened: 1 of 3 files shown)


JFCote:
> Is it not the optional properties thats supposed to be nulled ? 😆

This is indeed the feature I added last week and that you merged this morning. I'm very sorry about that, but I inverted the condition. Fortunately, the "cli interface" stay the same, only internal code and test are now inverted. Documentation stay the same as far as I know.
Should be quick to review and merge when you have time.

Again, sorry I think I was a bit tired last week!

jonaslagoni (approved):
Dont sweat it 👍 

Read some more reasoning here: https://stackoverflow.com/questions/54724304/what-does-null-statement-mean

All good 👍

Issue #1392, opened 2023-06-08
Add descriptions for options in playground continued..


SumantxD:
Continuing from issue #1347, where the design and implementation details were discussed 
and finalized, let's now outline the following tasks at hand:
- [X] discuss and finalize what needs to be added as info for the [general options](https://modelina.org/playground)
- [X] TypeScript Specific options
- [ ] CSharp Specific options
- [ ] PHP Specific options
- [ ] C++ Specific options
- [ ] Info about package name option for (Java, go , kotlin)

SumantxD:
> @SumantxD @jonaslagoni Can I help out some way?

sure you can take TypeScript Specific options.

jonaslagoni:
Just a "pro" tip, you can use the examples and documentation to find the description of each option.

If you are in doubt about a specific one, just use a placeholder and raise the PR, then we can help narrow it down 🙂

SumantxD:
# General Options

- Include change type mapping example
    - In code generation, a common task is to map the data types from the input model to the output code. One approach to accomplish this is by using type mapping. Type mapping is the process of mapping a data type in the input model to a corresponding data type in the output code.
    - Change type mapping refers to the process of transforming one data type to another data type. This is a common technique in programming when data types need to be converted to make them compatible with another system or process
- Change indentation type
    - The indentation type option allows you to choose between using tabs or spaces for indentation in the generated code.
- Change property naming format
    - This option allows you to customize the naming style for properties in your code. It provides several options to choose from:
        - Default: This option refers to the default naming format for properties, which may vary depending on the programming language or coding convention being used.

        - Snake case: Property names are written in lowercase letters, and words are separated by underscores. (e.g: property_name)

        - Pascal case: Property names start with an uppercase letter, and subsequent words are also capitalized. (e.g: PropertyName)

        - Camel case: Property names start with a lowercase letter, and subsequent words are capitalized. (e.g: propertyName)

        - Param case: Property names use hyphens to separate words, and all letters are in lowercase. (e.g: property-name)

        - Constant case: Property names are written in uppercase letters, and words are separated by underscores. (e.g: PROPERTY_NAME)
- Change model naming format
    - This option allows you to modify the naming style for models in your code. It includes the following options
        - Default: This option refers to the default naming format for models, which may vary depending on the programming language or coding convention being used. 

        - Snake case: Model names are written in lowercase letters, and words are separated by underscores. (e.g: model_name)

[...]

PR #1412, opened 2023-06-15
chore(website): add info to options for c++, kotlin, and go

SumantxD:
this PR tries to add info to options for the language c++, kotlin and go as mentioned in issue #1392


jonaslagoni (approved):
Nice 💪

Issue #1438, opened 2023-07-11
Java: enable nullable models doesn't support double


dlkj:
#### Describe the bug
Nullable models, implemented by #1141 doesn't support doubles correctly

#### How to Reproduce

Properties that are non-nullable and of type `float` should create a property of `double` not `Double` when rendering java models.

This can be seen on multiple tests including `test/generators/java/__snapshots__/JavaGenerator.spec.ts`

#### Expected behavior

`double` should be used for non-nullable floats

jonaslagoni:
@dlkj I guess this have been fixed right?

dlkj:
Yes, fixed by #1439

PR #1141, opened 2023-02-21
feat: make MetaModel's nullable

jonaslagoni:
**Description**
This PR adds `isNullable` property to the MetaModel and adapts the conversion of CommonModel to MetaModel.

**Related issue(s)**
Fixes https://github.com/asyncapi/modelina/issues/754
Fixes https://github.com/asyncapi/modelina/issues/749

PR #1439, opened 2023-07-11
fix: #1438 java enable nullable models doesn't support double

dlkj:
**Description**

- add missing `double` nullability support for java models to `src/generators/java/JavaConstrainer.ts`

**Related issue(s)**

* #1141


jonaslagoni:
@dlkj thanks for taking the time to do the PR!

Changed the release to breaking `fix!:` just to be cautious, could argue it's just a fix, but this breaks already using models 🙂

jonaslagoni (approved):

PR #1462, opened 2023-08-08
fix: dotnet JSON Serializer preset does not access accessors

### LINKED ISSUE #1448: Generator for dotnet System.Text.Json.JsonConverter produces code that does not match case settings and does not compile.
#### Describe the bug
When generating a model the property name setting was kept to default value (PascalCase for dotnet). An option to generate json converter was also set. The output code does not compile due to property case mismatch.

#### How to Reproduce
Reproduced in modelina playground:

The sample api documment:
```yaml
{
    "asyncapi": "2.5.0",
    "info": {
        "title": "Streetlights API",
        "version": "1.0.0"
    },
    "channels": {
        "light/measured": {
            "publish": {
                "summary": "Inform about environmental lighting conditions for a particular streetlight.",
                "operationId": "onLightMeasured",
                "message": {
                    "name": "LightMeasuredEvent",
                    "payload": {
                        "type": "object",
                        "$id": "LightMeasuredEvent",
                        "required": [ "id" ],
                        "properties": {
                            "id": {                                        // <-- camel case in naming
                                "type": "integer",
                                "minimum": 10
                            },
                            "sentAt": {
                                "type": "string",
                                "format": "date-time"
                            }
                        }
                    }
                }
            }
        }
    }
}
```

with code generator options:
```ts
const generator = new CSharpGenerator({
  indentation: {
    type: IndentationTypes.SPACES
  },
  collectionType: 'Array',
  autoImplementedProperties: true,
  namespace: 'asyncapi.models',
  nullable: true,
 presets: [
  CSHARP_JSON_SERIALIZER_PRESET
]
});
```

produces model code with expected property format:
```csharp
[JsonConverter(typeof(LightMeasuredEventConverter))]
public class LightMeasuredEvent
{
	public int Id { get; set; }                             // <-- pascal case in model
	public string? SentAt { get; set; }
	public Dictionary<string, dynamic>? AdditionalProperties { get; set; }
}
```
the conventer, however uses camel case naming to access those properties:
```csharp
// ...
if (value.id != null)
{
	// write property name and let the serializer serialize the value itself
	writer.WritePropertyName("id");
	JsonSerializer.Serialize(writer, value.id, options);   // <-- camel case in converter,  should be "value.Id"
}
if (value.sentAt != null)
{
	// write property name and let the serializer serialize the value itself
	writer.WritePropertyName("sentAt");
	JsonSerializer.Serialize(writer, value.sentAt, options);   // <-- camel case in converter, should be "value.SentAt"
}
// ..
```

#### Expected behavior
The json converter code produced by the generator should be valid and possible to compile.

jamesmoey:
**Description**
Use pascalCase in JSON Serializer preset

**Related issue(s)**
Fixes #1448 

**Test Run Output**
[test.txt](https://github.com/asyncapi/modelina/files/12286854/test.txt)

### DIFF
--- a/examples/csharp-generate-json-serializer/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-generate-json-serializer/__snapshots__/index.spec.ts.snap
@@ -47,7 +47,7 @@ internal class RootConverter : JsonConverter<Root>
       if (propertyName == \\"email\\")
         {
           var value = JsonSerializer.Deserialize<string?>(ref reader, options);
-          instance.email = value;
+          instance.Email = value;
           continue;
         }
     }
@@ -65,10 +65,10 @@ internal class RootConverter : JsonConverter<Root>
   
     writer.WriteStartObject();
 
-    if(value.email != null) { 
+    if(value.Email != null) { 
       // write property name and let the serializer serialize the value itself
       writer.WritePropertyName(\\"email\\");
-      JsonSerializer.Serialize(writer, value.email, options);
+      JsonSerializer.Serialize(writer, value.Email, options);
     }
 
 
--- a/src/generators/csharp/presets/JsonSerializerPreset.ts
+++ b/src/generators/csharp/presets/JsonSerializerPreset.ts
@@ -8,6 +8,7 @@ import {
   ConstrainedReferenceModel
 } from '../../../models';
 import { CSharpOptions } from '../CSharpGenerator';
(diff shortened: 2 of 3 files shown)


jonaslagoni (approved):
Awesome @jamesmoey 👌

jonaslagoni:
@all-contributors please add @jamesmoey for code and test

PR #1469, opened 2023-08-09
feat: make discriminator property visible to deserializer from Java Jackson preset

dlkj:
**Description**

When deserializing `oneOf` structures using non-const discriminators, the discriminator value is not visible in the deserialized model.

Add `visible=true` to `@JsonTypeInfo`

This doesn't change the behaviour when using const discriminators.


jonaslagoni:
Is this a breaking change? 🤔

dlkj:
@jonaslagoni hmm, now you ask, it isn't really breaking. The data on the wire should be the same and the current behaviour will only change for cases that are currently broken.

PR #1477, opened 2023-08-15
fix: problem with json serializer option with C# generator

JFCote:
**Description**

- Fix generation problem at 2 places when using C# generator with json serializer option
* Missing "s" for a class
* Missing "value." when deserializing.


JFCote:
@jonaslagoni Please note that this will only fix the compilation. I tried the fixed code and it crash with exception at runtime. But that is another story and I don't think I have currently the knowledge to fix it myself.

jonaslagoni:
What problems are you phasing @JFCote? or exceptions?

jonaslagoni (approved):

PR #1482, opened 2023-08-18
fix: json serialization equals not working

JFCote:
**Description**

When using serialization with Json with capital letters, it was not working. I fix it by ignoring case during the comparison.


jonaslagoni:
@JFCote what is the input where this is not working? 

Cause the `unconstrainedPropertyName` should be the expected value for the property when serialized and deserialized. In theory this logic you added should be unnecessary.

Cause if you have the following input:

```
{
  "type": "object",
  "properties": {
    "NumberPropWithCapitalCase": {
      "type": "number"
    }
}
```

It currently correctly checks for `if (propertyName == "NumberPropWithCapitalCase")`.

PR #1484, opened 2023-08-18
test: add runtime testing for Go models 

### LINKED ISSUE #1333: Add runtime testing for Go
#### Reason/Context
In https://github.com/asyncapi/modelina/pull/1272 we added the initial runtime setup for Java, now it's time to replicate it for Go.

The reason we add these is to ensure that the code that is generated is semantically and syntactically correct, and the only way to achieve this is by testing the models at runtime.

This initial setup should be as simple as possible, just a Go project, with a single test and a single generated model, and that's it.

Here is the general task list:

- [ ] Create a Go project under; `test/runtime/runtime-go`
- [ ] Add a simple test code generation script here: `test/runtime/runtime-go.ts`. Check out the Java variant as an example https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-java.ts
- [ ] Add the generation script to the package file: https://github.com/asyncapi/modelina/blob/ba0bd41e9803394f56872c5ec92f8e335a4d687a/package.json#L98
- [ ] Run the generation script
- [ ] Manually add a simple test for the generated models, could be as simple as just creating an instance of it.
- [ ] Add a simple Jest script, which just execute the Go test: https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-java.spec.ts
- [ ] Add a simple script that will run the test: https://github.com/asyncapi/modelina/blob/ba0bd41e9803394f56872c5ec92f8e335a4d687a/package.json#L97
- [ ] Create a CI workflow that runs the runtime tests: https://github.com/asyncapi/modelina/blob/master/.github/workflows/runtime-java-testing.yml

And that's it 🎉

Devansh-Bhatt:
**Description**

- Created a Go Project under `test/runtime/runtime-go`
- Added a simple test-code generations script at: `test/runtime/runtime-go`
- Added a generation script in `package.json`
- Ran the generation script and ignored the directory `test/runtime/runtime-go/target` which contained the auto generated models as a result of running the generation script
- Manually added a simple test for the generated models under `test/runtime/runtime-go/Test_Go/Test_Address.go`
- Added a jet script to execute the Go test
- Added the script to the run the test in `package.json`
- Create  CI Workflow that runs the runtime test under ` .github/workflows/runtime-go-testing.yml
`

**Related issue(s)**
Resolves #1333


test/runtime/runtime-go/Test_Go/Test_Address.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestShouldNotContainAdditionalPropertiesWhenSerialized(t *testing.T) {
+	address := runtimego.Address{
+		StreetName:           "Test Address 2",
+		HouseNumber:          2.0,
+		Marriage:             true,
+		Members:              nil,
+		ArrayType:            nil,
+		NestedObject:         nil,
+		AdditionalProperties: nil,
```
jonaslagoni, line 40: @Devansh-Bhatt if you look at one of the other runtime test, do you mind using the same values as in there? 🙂

test/runtime/runtime-go/Test_Go/Test_Address.go
```diff
@@ -0,0 +1,58 @@
+package Test_Go
+
+import (
+	"encoding/json"
+	runtimego "runtimego/target"
+	"testing"
+)
+
+func TestShouldbeAbleToSerializeModel(t *testing.T) {
```
jonaslagoni, line 9: We still need to compare the actual serialized value with something we expect, you can grab the one in the other runtime test 🙂

jonaslagoni (changes requested):
Looks great @Devansh-Bhatt!

Just got those two comments, and then please use `test:` as the PR title 🙂

test/runtime/runtime-go/Test_Go/Test_Address.go
jonaslagoni: Yep, thats about it 👍

Devansh-Bhatt:
@jonaslagoni Is there any issue with the commit / changes that i made. I quite don't seem to understand what's the issue here 😕

jonaslagoni (approved):
LGTM 👍

jonaslagoni:
@all-contributors please add @Devansh-Bhatt  for test, infra

PR #1486, opened 2023-08-22
fix: adds format in options instead of using original input

kennethaasan:
**Description**

- fix: adds format in options instead of using original input. The original input can't be trusted when merging schemas. This PR fixes that issue.

**Related issue(s)**
Please review and merge after #1485 because this PR is based on that branch.


kennethaasan:
@jonaslagoni, please review.

jonaslagoni (approved):
👍

PR #1491, opened 2023-08-24
feat: add PHP JsonSerializable Preset

markuspoerschke:
In PHP objects (and also arrays and scalars) can be serialized to JSON using the function `json_encode()`. In order to maintain the correct property names and values for Enums, normalization of values needs to be implemented manually using the `JsonSerializable` interface.

The serialization to JSON can be enabled by using the optional preset `PHP_JSON_SERIALIZABLE_PRESET`.

This way of serializing PHP objects does not require any user land package and works out of the box with each PHP version. It is also compatible with different serialization frameworks, like Symfony.

### DIFF
--- a/docs/languages/Php.md
+++ b/docs/languages/Php.md
@@ -36,7 +36,6 @@ This will ensure that, for example, enum values and property names are serialize
 To add support of serialization to JSON, apply the `PHP_JSON_SERIALIZABLE_PRESET` preset.
 Check out this [example for a live demonstration](../../examples/php-generate-json-serializable-preset).
 
-Marshalling PHP objects from JSON is currently supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
 
 ### To and from XML
 
--- a/docs/languages/Php.md
+++ b/docs/languages/Php.md
@@ -15,21 +15,33 @@ There are special use-cases that each language supports; this document pertains 
 
 ## Description Present
 
-By default, descriptions are not rendered for the model, you can change that by applying `PHP_DESCRIPTION_PRESET`.
+By default, descriptions are not rendered for the model; you can change that by applying `PHP_DESCRIPTION_PRESET`.
 
 Check out this [example for a live demonstration](../../examples/php-generate-documentation-preset).
 
 ## Generate serializer and deserializer functionality
 
-The most widely used usecase for Modelina is to generate models that include serilization and deserialization functionality to convert the models into payload data. This payload data can of course be many different kinds, JSON, XML, raw binary, you name it.
+The most widely used usecase for Modelina is to generate models that include serialization and deserialization functionality to convert the models into payload data.
+This payload data can, of course, be many different kinds, JSON, XML, raw binary, you name it.
 
-As you normally only need one library to do this, we developers can never get enough with creating new stuff, therefore there might be one specific library you need or want to integrate with. Therefore there is not one specific preset that offers everything. Below is a list of all the supported serialization presets. 
+As you normally only need one library to do this, we developers can never get enough of creating new stuff, therefore, there might be one specific library you need or want to integrate with.
+Therefore, there is not one specific preset that offers everything. Below is a list of all the supported serialization presets. 
 
 ### To and from JSON
-Currently not supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
+
+Objects in PHP can generally be serialized to JSON using the [`json_encode()` function](https://www.php.net/manual/en/function.json-encode.php).
+To ensure that the data is serialized correctly, the [`JsonSerializable` interface](https://www.php.net/manual/en/class.jsonserializable.php) needs to be implemented.
+This will ensure that, for example, enum values and property names are serialized correctly.
+
+To add support of serialization to JSON, apply the `PHP_JSON_SERIALIZABLE_PRESET` preset.
+Check out this [example for a live demonstration](../../examples/php-generate-json-serializable-preset).
+
+Marshalling PHP objects from JSON is currently supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
 
 ### To and from XML
+
 Currently not supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
 
 ### To and from binary
+
 Currently not supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
--- a/examples/php-generate-json-serializable-preset/__snapshots__/index.spec.ts.snap
+++ b/examples/php-generate-json-serializable-preset/__snapshots__/index.spec.ts.snap
@@ -8,7 +8,7 @@ declare(strict_types=1);
 namespace Asyncapi;
 
 
-use JsonSerializable;
+
 final class Root implements \\\\JsonSerializable
 {
   private ?string $email;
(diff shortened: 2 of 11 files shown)


src/generators/php/presets/JsonSerializablePreset.ts
```diff
@@ -0,0 +1,88 @@
...
+  content,
+  renderer
+}: {
+  content: string;
+  renderer: PhpRenderer<ConstrainedMetaModel>;
+}): string {
+  renderer.dependencyManager.addDependency('use JsonSerializable;');
+
+  const contentLines = content.split('\n');
+  contentLines[0] += ` implements JsonSerializable`;
```
markuspoerschke, line 15: Implementing adding interfaces to classes this may cause incompatibilities when multiple presets try to implement interfaces. Can you imagine any other solution here?

src/generators/php/presets/JsonSerializablePreset.ts
```diff
@@ -0,0 +1,88 @@
...
+                    return `'${property.unconstrainedPropertyName}' => $this->${property.propertyName},`;
+                  })
+                )
+              ),
+              '];'
+            ])
+          ),
+          '}'
+        ])
+      );
```
jonaslagoni, line 57: All these nested renderBlock and indent calls can quickly become unmaintainable. 

An alternative is to do something along the following:
```suggestion
  const serializedProperties = Object.values(model.properties).map((property) => {
    if (property.property instanceof ConstrainedDictionaryModel && property.property.serializationType === 'unwrap') {
      return `...$this->${property.propertyName}`;
    }
  
    return `'${property.unconstrainedPropertyName}' => $this->${property.propertyName}`;
  });

  return `${content}
  public function jsonSerialize(): array
  {
    return [
     ${renderer.indent(serializedProperties.join(',\n'))}
    ];
  }
```

jonaslagoni (changes requested):
@markuspoerschke thanks for taking the time to provide a PR 🙏 

I have a few comments :v:

@Ferror do you have time to do a review? 🤔

src/generators/php/presets/JsonSerializablePreset.ts
jonaslagoni: Uhh, yeah, this one is tough. 

We could later on when the problem occurs implement a support function to add interfaces.

Alternatively, all functionality that need to add a new interface would need to be part of the same preset and then we control what is rendered based on options. But that's not really maintainable long term 🤔

markuspoerschke: I am fine with postponing the problem. I see no other use-case at the moment. Likewise, I was also considering putting it somehow to the options.

I will close this conversation.

docs/languages/Php.md
```diff
@@ -15,21 +15,33 @@ There are special use-cases that each language supports; this document pertains
...
-Currently not supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
+
+Objects in PHP can generally be serialized to JSON using the [`json_encode()` function](https://www.php.net/manual/en/function.json-encode.php).
+To ensure that the data is serialized correctly, the [`JsonSerializable` interface](https://www.php.net/manual/en/class.jsonserializable.php) needs to be implemented.
+This will ensure that, for example, enum values and property names are serialized correctly.
+
+To add support of serialization to JSON, apply the `PHP_JSON_SERIALIZABLE_PRESET` preset.
+Check out this [example for a live demonstration](../../examples/php-generate-json-serializable-preset).
+
+Marshalling PHP objects from JSON is currently supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
```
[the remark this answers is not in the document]
jonaslagoni: Ahhh, so it should have been:
```suggestion
Marshalling PHP objects from JSON is currently NOT supported, [let everyone know you need it](https://github.com/asyncapi/modelina/issues/new?assignees=&labels=enhancement&template=enhancement.md)!
```

examples/php-generate-json-serializable-preset/__snapshots__/index.spec.ts.snap
```diff
@@ -0,0 +1,28 @@
...
+exports[`Should be able to render PHP and should log expected output to console 1`] = `
+Array [
+  "<?php
+declare(strict_types=1);
+
+namespace Asyncapi;
+
+
+use JsonSerializable;
+final class Root implements JsonSerializable
```
Ferror, line 12:
```suggestion
declare(strict_types=1);

namespace Asyncapi;

final class Root implements \JsonSerializable
```

How about this outcome?

test/generators/php/presets/__snapshots__/JsonSerializablePreset.spec.ts.snap
```diff
@@ -0,0 +1,46 @@
...
+  case VALUE_MINUS_A;
+  case VALUE_MINUS_B;
+
+  public function jsonSerialize(): mixed
+  {
+    return match($this) {
+      self::VALUE_MINUS_A => \\"value-A\\",
+      self::VALUE_MINUS_B => \\"value-B\\",
+    };
+  }
```
Ferror, line 43: Idk if that's the right approach, but have no other idea how to handle not backend enums.

also I guess we should add a test with backed enums. I will add an issue for that

Ferror (dismissed):
I much more prefer using Serializers form Symfony or other, but I can guess people may like that. Next would be ability to create event from array of data 😄

test/generators/php/presets/__snapshots__/JsonSerializablePreset.spec.ts.snap
markuspoerschke: How can you define a non-backed enum in JSON schema? If you can provide an example, I could test it.

I added at least a test case for mixed enums: 048c9e17d7e0431f032f707e0810dc93ffd4b3c5

examples/php-generate-json-serializable-preset/__snapshots__/index.spec.ts.snap
markuspoerschke: I changed this in 048c9e17d7e0431f032f707e0810dc93ffd4b3c5

markuspoerschke:
> I much more prefer using Serializers form Symfony or other, but I can guess people may like that. Next would be ability to create event from array of data 😄

Me too. The cool thing is, that Symfony ships with a [JsonSerializableNormalizer](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Serializer/Normalizer/JsonSerializableNormalizer.php), which allows using this kind of PHP object also with Symfony serializer.

src/generators/php/presets/JsonSerializablePreset.ts
markuspoerschke: The indention is configurable. The suggested solution will not make use of the defined size and type of indention.

I applied the suggestion only partially in f2616f9d2e741282ccee785ecefea35b9c6f5299

test/generators/php/presets/__snapshots__/JsonSerializablePreset.spec.ts.snap
Ferror: That's a good question. Ideally, we should only have backed enums because the protocol needs values.

PHP examples would be:
```php
enum Suit: string
{
    case Hearts = 'H';
    case Diamonds = 'D';
    case Clubs = 'C';
    case Spades = 'S';
}
```

```
enum Week: int
{
    case Monday = 1;
    case Tuesday = 2;
    case Wednesday = 3;
}
```

jonaslagoni (changes requested):
The good thing is that even if two competing serialization libraries, each can have their own preset 🙂 We have such a case in C# for example.

Seems like some of the test snapshots need an update 🙂

Ferror (approved):

jonaslagoni (approved):

jonaslagoni:
@all-contributors please add @markuspoerschke for code, example, test, docs

PR #1498, opened 2023-09-04
feat!: generate models for OpenAPI parameters

dlkj:
**Description**
Models that were only present in OpenAPIV3 parameters were not being generated. This was causing problems for parameters with schema like `enum` that can be reflected in the language type system.

### DIFF
--- a/src/processors/OpenAPIInputProcessor.ts
+++ b/src/processors/OpenAPIInputProcessor.ts
@@ -163,14 +163,16 @@ export class OpenAPIInputProcessor extends AbstractInputProcessor {
   }
 
   private iterateParameters(
-    parameters: (OpenAPIV3.ReferenceObject | OpenAPIV3.ParameterObject)[] | undefined,
+    parameters:
+      | (OpenAPIV3.ReferenceObject | OpenAPIV3.ParameterObject)[]
+      | undefined,
     path: string,
     inputModel: InputMetaModel,
     options?: ProcessorOptions
   ) {
     for (const parameterObject of parameters || []) {
       const parameter = parameterObject as OpenAPIV3.ParameterObject;
-      if(parameter.schema){
+      if (parameter.schema) {
         this.includeSchema(
           parameter.schema as OpenAPIV3.SchemaObject,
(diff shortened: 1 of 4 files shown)


dlkj:
Adding a more explicit unit test to `test/processors/OpenAPIInputProcessor.spec.ts` has turned out to be a fiddly as there are no models in the processed `InputMetaModel` due to the mocked `Interpreter`.

dlkj:
Converting to draft - I've now got my head round how to write a more targeted unit test defending this change. Will convert back to a full PR once I have added it.

dlkj:
Added an explicit unit test and fixed up the schema naming, ready for review now.

src/processors/OpenAPIInputProcessor.ts
```diff
@@ -160,6 +172,27 @@ export class OpenAPIInputProcessor extends AbstractInputProcessor {
...
   }
 
+  private iterateParameters(
+    parameters:
+      | (OpenAPIV3.ReferenceObject | OpenAPIV3.ParameterObject)[]
+      | undefined,
+    path: string,
+    inputModel: InputMetaModel,
+    options?: ProcessorOptions
+  ) {
```
jonaslagoni, line 182: You dont want to introduce an option so the user can specify whether they want to include parameters or not? 🤔

jonaslagoni (changes requested):

src/processors/OpenAPIInputProcessor.ts
dlkj: The parameters are just as much part of the OpenAPI spec as the requests and responses, so I don't see much of a use case in having them as an option. The extra cost is minimal. Happy to add one if you have strong views. I don't see anything similar on `ProcessorOptions` for other document formats.

jonaslagoni: Thats because in for example AsyncAPI input processor, we only generate message payloads, not parameters, even though we probably should 😄 

But no I dont have any strong feelings for the OpenAPI input, so here it's fine 😄

PR #1501, opened 2023-09-07
fix: fixes performance issues in java when calling parent unions

kennethaasan:
**Description**

Some backround: To do the java code gen for unions, when you generate each model you need to know "what unions am I a part of, and what is the ConstrainedUnionModel for all those unions. To do this naively you can walk the full document and look for unions that contain the current model. This PR moves this "walk" to the `AbstractGenerator` so you only need to do it once over the document [O(n)] rather than having to do this for each model and then walk the document [O(n^2)].

In https://github.com/asyncapi/modelina/blob/next/src/generators/java/renderers/ClassRenderer.ts#L104-L110 we convert all UnionMetaModel's to ConstrainedUnionModel's for every model. This leads to a huge performance hit. In this PR I moved that conversion to the `AbstractGenerator` and stores it in the Constrained Model. These changes improved generating Java code from 2m15s to 7s for my employer's AsyncAPI spec on my laptop.

**Related issue(s)**

### DIFF
--- a/src/generators/AbstractGenerator.ts
+++ b/src/generators/AbstractGenerator.ts
@@ -116,18 +116,10 @@ export abstract class AbstractGenerator<
     return options.dependencyManager;
   }
 
-  /**
-   * Generates the full output of a model, instead of a scattered model.
-   *
-   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
-   *
-   */
-  public async generateCompleteModels(
-    input: any | InputMetaModel,
-    completeOptions: Partial<RenderCompleteModelOptions>
-  ): Promise<OutputModel[]> {
-    const inputModel = await this.processInput(input);
-
+  private getConstrainedModels(inputModel: InputMetaModel): Array<{
+    constrainedModel: ConstrainedMetaModel;
+    dependencyManager: AbstractDependencyManager;
+  }> {
     interface ConstrainedMetaModelWithDepManager {
       constrainedModel: ConstrainedMetaModel;
       dependencyManager: AbstractDependencyManager;
@@ -156,6 +148,7 @@ export abstract class AbstractGenerator<
         unionConstrainedModelsWithDepManager.push(
           getConstrainedMetaModelWithDepManager(model)
         );
+        continue;
       }
 
       constrainedModelsWithDepManager.push(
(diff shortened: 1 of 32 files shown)


src/generators/AbstractGenerator.ts
```diff
@@ -41,6 +43,25 @@ export const defaultGeneratorOptions: CommonGeneratorOptions = {
...
+
+export interface AbstractGeneratorRenderCompleteModelArgs<
+  Options extends CommonGeneratorOptions,
+  RenderCompleteModelOptions
+> {
+  constrainedModel: ConstrainedMetaModel;
+  inputModel: InputMetaModel;
+  completeOptions: Partial<RenderCompleteModelOptions>;
+  options?: DeepPartial<Options>;
+}
```
jonaslagoni, line 63: Any reason you want to break the options, or just left over from the old solution?

jonaslagoni (changes requested):

src/generators/java/JavaGenerator.ts
```diff
@@ -118,33 +122,118 @@ export class JavaGenerator extends AbstractGenerator<
     );
   }
 
+  /**
+   * Generates the full output of a model, instead of a scattered model.
+   *
+   * OutputModels result is no longer the model itself, but including package, package dependencies and model dependencies.
+   *
+   */
+  public async generateCompleteModels(
```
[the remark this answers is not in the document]
kennethaasan: Agree, I moved it to the `AbstractGenerator`.

src/generators/AbstractGenerator.ts
kennethaasan: I was thinking about it after we agreed to change the initial PR. I decided to keep it like this because I think that this approach is better than what we have, but I can remove it in case you disagree.

jonaslagoni: If you can easily re-commit these changes to a separate branch and do a separate PR I would appreciate it 🙏 

It's cluttering the changes of what this PR actually tries to solve, including the discussions will go into multiple directions 😅

kennethaasan: Removed those changes. I'll submit another PR with those changes after this is merged.

kennethaasan:
Ready for review.

kennethaasan:
@dlkj, please reivew

src/generators/AbstractGenerator.ts
```diff
@@ -97,6 +99,74 @@ export abstract class AbstractGenerator<
     return options.dependencyManager;
   }
 
+  private getConstrainedModels(inputModel: InputMetaModel): Array<{
+    constrainedModel: ConstrainedMetaModel;
+    dependencyManager: AbstractDependencyManager;
+  }> {
```
jonaslagoni, line 105: Do you mind adding some jsdoc's for this function? Cause I have no idea why it's there for all generators and what it does 😅

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
```diff
@@ -1052,6 +1052,18 @@ exports[`JavaGenerator should render enums with translated special characters 1`
...
+
+",
+  "package test.packageName;
+
+
+",
+  "package test.packageName;
+
+
+",
```
jonaslagoni, line 1065: This seems a bit wrong 🤨

src/models/ConstrainedMetaModel.ts
```diff
@@ -19,6 +19,7 @@ export interface ConstrainedMetaModelOptionsDiscriminator
 export class ConstrainedMetaModelOptions extends MetaModelOptions {
   const?: ConstrainedMetaModelOptionsConst;
   discriminator?: ConstrainedMetaModelOptionsDiscriminator;
+  parentUnions?: ConstrainedUnionModel[];
```
dlkj, line 22: Torn between making this a more general `parents` list of all parent models and the specific `parentUnions` required for this feature. This is probably the right call and could be extended if required in future to be `parents`.

jonaslagoni (changes requested):

dlkj (approved):

src/models/ConstrainedMetaModel.ts
jonaslagoni: Didnt even notice it was called parentUnions instead of `parentUnions` 😄 

Would make sense to me yea 👍

dlkj: The other downside of moving this to `parents` would be the need to relax the type and do a bunch of reflection on access. Happy to leave this as it is.

test/generators/java/__snapshots__/JavaGenerator.spec.ts.snap
kennethaasan: Well, it's not introduced in this PR. The order of the output has changed a bit because of the changes in `AbstractGenerator.getConstrainedModels`.

src/models/ConstrainedMetaModel.ts
kennethaasan: `parents` can be handy for other use cases in the future, so I changed it to that. The access for Java's ClassRenderer only needed a check for `model instanceof ConstrainedUnionModel`, so I believe that's perfectly fine.

dlkj (approved):

jonaslagoni (approved):

PR #1502, opened 2023-09-08
fix!: fixes java jackson discriminator when property is not camel case

kennethaasan:
**Description**

- fixes Java Jackson discriminator when the property is not camel case. You get a runtime error without this fix.

**Related issue(s)**

### DIFF
--- a/test/generators/java/presets/__snapshots__/JacksonPreset.spec.ts.snap
+++ b/test/generators/java/presets/__snapshots__/JacksonPreset.spec.ts.snap
@@ -55,7 +55,7 @@ exports[`JAVA_JACKSON_PRESET should render Jackson annotations for enum 1`] = `
 
 exports[`JAVA_JACKSON_PRESET union handle oneOf with AsyncAPI discriminator with Jackson 1`] = `
 Array [
-  "@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.EXISTING_PROPERTY, property=\\"vehicleType\\", visible=true)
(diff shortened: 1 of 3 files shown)


jonaslagoni (changes requested):
Is this a breaking change? 🤔

test/generators/java/presets/__snapshots__/JacksonPreset.spec.ts.snap
```diff
@@ -64,10 +64,10 @@ Array [
  * Vehicle represents a union of types: Car, Truck
  */
 public interface Vehicle {
-  String getVehicleType();
```
dlkj, line 67: Was dropping this intentional? The current functionality exposes the discriminator if it is the same type across all the union types.

dlkj: I suspect this is caused by a similar name matching issue.

kennethaasan: Good catch. The same fix was needed in `addDiscriminatorTypeToUnionModel` which I just added.

kennethaasan:
@jonaslagoni, I thought we had to make every PR to the `next` branch a breaking change?

jonaslagoni:
> @jonaslagoni, I thought we had to make every PR to the `next` branch a breaking change?

Only those that are 😄

jonaslagoni:
@kennethaasan the only difference between `next` and `master` is that `next` already contains breaking change commits, and it will always only change 1 major version regardless of how many commits we do with breaking changes. Because it is using pre-releases.

So yea, only mark a PR as breaking if it is in fact a breaking change, otherwise we just do regular `fix` 🙂

kennethaasan:
@jonaslagoni , I see 😄 I guess I've only done breaking PRs so far.

jonaslagoni (approved):

PR #1509, opened 2023-09-20
fix: escape java string literals used for regex patterns and string constants

dlkj:
**Description**

- Java string constants were being rendered without escaping. This could generate invalid java. This was a particular issue for regex patters as these frequently include the `\` character

### DIFF
--- a/test/generators/java/presets/__snapshots__/ConstraintsPreset.spec.ts.snap
+++ b/test/generators/java/presets/__snapshots__/ConstraintsPreset.spec.ts.snap
@@ -10,7 +10,7 @@ exports[`JAVA_CONSTRAINTS_PRESET should render constraints annotations 1`] = `
   private double maxNumberProp;
   @Size(min=2, max=3)
   private Object[] arrayProp;
-  @Pattern(regexp=\\"^I_\\")
+  @Pattern(regexp=\\"^\\\\\\\\w+(\\\\\\"\\\\\\\\.\\\\\\\\w+)*$\\")
   @Size(min=3)
   private String stringProp;
   private Map<String, Object> additionalProperties;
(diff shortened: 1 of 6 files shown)


test/generators/java/presets/__snapshots__/ConstraintsPreset.spec.ts.snap
```diff
@@ -10,7 +10,7 @@ exports[`JAVA_CONSTRAINTS_PRESET should render constraints annotations 1`] = `
   private double maxNumberProp;
   @Size(min=2, max=3)
   private Object[] arrayProp;
-  @Pattern(regexp=\\"^I_\\")
+  @Pattern(regexp=\\"^\\\\\\\\w+(\\\\\\"\\\\\\\\.\\\\\\\\w+)*$\\")
```
jonaslagoni, line 13: Just double checking this looks right to you @dlkj 😄

jonaslagoni (changes requested):

jonaslagoni:
Is this not a fix rather then feature? 🤔

dlkj:
> Is this not a fix rather then feature? 🤔

Yes, sorry, this is a fix. Updated title

test/generators/java/presets/__snapshots__/ConstraintsPreset.spec.ts.snap
dlkj: this does look right, but let's step through it slowly.

The diff is part of an escaped  JS string capturing the output of the test. This has been escaped twice.

`"@Pattern(regexp=\\"^\\\\\\\\w+(\\\\\\"\\\\\\\\.\\\\\\\\w+)*$\\")"`

-->

`@Pattern(regexp=\"^\\\\w+(\\\"\\\\.\\\\w+)*$\")`

-->

`@Pattern(regexp="^\\w+(\"\\.\\w+)*$")`

jonaslagoni (approved):
👍

PR #1547, opened 2023-10-04
docs: use mermaid for class diagram

### LINKED ISSUE #1182: [📑 Docs]: Use code to represent the internal models
### What Dev Docs changes are you proposing?

Currently, it's not easy to update the [internal model images](https://github.com/asyncapi/modelina/blob/master/docs/internal-model.md) and keeping them up to date, as I have the draw.io file...

Instead, we should add them as code through something like Mermaid: https://mermaid.js.org/syntax/classDiagram.html

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Qypol342:
**Description**
Corrected sentence:
- Update documentation internal-model class diagram from PNG to [Mermaid](https://mermaid.js.org/syntax/classDiagram.html#styling) allowing for easier updates. Preview result [here](https://github.com/Qypol342/modelina/blob/master/docs/internal-model.md)

**Related issue(s)**
Resolves #1182


docs/internal-model.md
```diff
@@ -19,9 +19,125 @@ These are the meta models and their meaning:
...
+    }
+
+    class TupleModel{
+        <<class>>
+        TupleValueModel[] tupleModels
+    }
+
+    class ObjectModel{
+        <<class>>
+        String, ObjectPropertyModel properties
```
jonaslagoni, line 48: Looks like there are a few of the properties that are missing their types. In the old image, it's a dictionary with `string` key and `ObjectPropertyModel` value 🙂 

There are a few places this applies 🙂

jonaslagoni (changes requested):
Nice @Qypol342, this is really needed 👍

docs/internal-model.md
Qypol342: @jonaslagoni There you go! I've fixed it up. 🙃

jonaslagoni (commented):
image

The `property --> MetaModel` is that correct? or is it being displayed incorrectly? 🤔

Qypol342:
> image
> 
> The `property --> MetaModel` is that correct? or is it being displayed incorrectly? 🤔

I've just followed the **original** doc:
image

`property --> MetaModel` the way it is **written** is the mermaid **standard** to signify an association **without** the arrow **going** all over the schema

I can make it look the old **good** way by not **specifying** which **attribute** the association **comes** from.

jonaslagoni (approved):
Alright, let's keep it 👍

jonaslagoni:
@all-contributors please add @Qypol342 for docs

kennethaasan:
Really great stuff @Qypol342!

PR #1592, opened 2023-11-02
fix: multi line descriptions for the pydantic preset in python

### LINKED ISSUE #1058: Python Pydantic improvements
##### Field.alias
It seems that for Pydantic `Field` typing use of `property.description` as an `alias` is incorrect, and `property.name` or some other field should be used instead. Current implementation:
```typescript
const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
  property(params) {
    const type = params.property.required
      ? params.property.property.type
      : `Optional[${params.property.property.type}]`;
    const alias = params.property.property.originalInput['description']
      ? `alias='${params.property.property.originalInput['description']}'`
      : '';

    return `${params.property.propertyName}: ${type} = Field(${alias})`;
  }
};
```
Possible fix:
```typescript
    const alias = params.property.property.originalInput['name']
      ? `alias='${params.property.property.originalInput['name']}'`
      : '';
```
##### Field.description
Introduce support for setting a Field's `description` from `property.description`

##### Extra.forbid
For models declaring `additionalProperties` as false enable native validation:
```python
from pydantic import BaseModel, Extra


class Model(BaseModel, extra=Extra.forbid):
    a: str
```

#### Description
- Change from `description` to `name` when setting `Field.alias` will be a breaking change, but it's more of a misuse or a bug in the current state.

kennethaasan:
**Description**

- fixes multi-line descriptions for the Pydantic preset in Python

**Related issue(s)**

Fixes #1058

### DIFF
--- a/examples/generate-python-pydantic-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-python-pydantic-models/__snapshots__/index.spec.ts.snap
@@ -3,8 +3,8 @@
 exports[`Should be able to render python models and should log expected output to console: class-model 1`] = `
 Array [
   "class Root(BaseModel): 
-  optionalField: Optional[str] = Field(alias='this field is optional')
-  requiredField: str = Field(alias='this field is required')
+  optionalField: Optional[str] = Field(alias='''this field is optional''')
+  requiredField: str = Field(alias='''this field is required''')
   noDescription: Optional[str] = Field()
   options: Optional[Options] = Field()
 ",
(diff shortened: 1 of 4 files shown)


jonaslagoni (changes requested):
Need some test snapshots updated 🙂 

Looks good otherwise 👍

kennethaasan:
> Need some test snapshots updated 🙂
> 
> Looks good otherwise 👍

Ups, updated now 😄

jonaslagoni (approved):
🔥

Issue #108, opened 2021-02-24
Figure out how we can introduce rendering of inheritance


jonaslagoni:
#### Reason/Context
One of the strong features of the CommonModel is the possibility to use `extend` which contains a list of `$id`s of the CommonModels that the current model should extend. Therefore we need to figure out how we render this in the different languages

jonaslagoni:
Still relevant.

jonaslagoni:
Still relevant.

smoya:
Currently, supported inputs don't really support inheritance (we could discuss if JSON Schema somehow does, but it doesn't really do). 

However, if we ever support any other language as input (like PHP for example), the generators will need explicit info about both a class extending and/or implementing an interface. See the following examples:

Class Iinheritance:
```php
class Human extends Animal
```

Implementing an interface:
```php
class Human implements Animal {
```

But also

```php
class Human extends Animal implements LivingBeing
```

And also the fact a class can implement several interfaces.

So, does this issue still applies? Do we want to support at model level? is it worth to support right now?

jonaslagoni:
I don't think we should focus on it now, as it seems like a feature and not a complete rewrite that is required to fully make this work. But good points to remember for the future 👍

smoya:
> I don't think we should focus on it now, as it seems like a feature and not a complete rewrite that is required to fully make this work. But good points to remember for the future 👍

Maybe you want to pop out this from [Version 1](https://github.com/asyncapi/modelina/milestone/3) milestone?

jonaslagoni:
I have been thinking a bit about this feature, and there is one huge problem with it, that I dont know if we want to introduce - and that is complexity.

When you write presets and use the class to anything, you MUST remember that it CAN extend from something else, and handle that accordingly. Which is a HUGE effort which needs to be put into Modelina, without much benefit IMO.

I can however see a reason we would want to introduce a preset hook (for now) that enables you to more easily extend from your own custom class, in case you have some use-case where that would suit you. But it does not force preset and generator creators to handle any of the complexity because it's not something thats added to the `ConstrainedModel`.

The complexity I am referring to is for example in serialize presets that given the following generated models in C#:
```
public class Person
{
    public string FName { get; set; }
    public string LName { get; set; }

    public Person(string fname, string lname)
    {
        FName = fname;
        LName = lname;
    }
}

public class Student : Person
{
    public Student(Person person, string code)
        : base(person.FName, person.LName)
    {
        this.code = code;
    }
[...]

PR #1593, opened 2023-11-02
feat!: adds inheritance with interfaces for java

kennethaasan:
**Description**

This MR adds inheritance with interfaces for Java. Our Java users are complaining about missing interfaces for schemas that include `allOf`. Unlike TypeScript, In Java, it's problematic to use classes without interfaces if lots of classes are similar. This is an attempt to fix that. This is a difficult one to solve in a good way, and the arguments in #108 describe that, but we need to find a way to improve the generated code for Java. 

**Related issue(s)**

See also #108 and #1613


kennethaasan:
@dlkj, perhaps you have some input here as a Java user who also might need this? Keep in mind that this is still a draft, but something we're prioritizing at the moment.

kennethaasan:
@jonaslagoni, I'm also eager to hear your input on how this can be solved. Am I on the right track, or should this be solved in a different way?

jonaslagoni (commented):
Honestly looks pretty good to me 🤔 

Looking forward to seeing this in C# and TypeScript.

My only concern would be how it handles multiple inheritance when`allOf` has multiple entries, in languages that only allow a single one? 🤔

kennethaasan:
Thanks for your feedback! I'm still working through issues we've found by generating code with these changes in our own codebase. Once I'm happy with the output I'll split this draft into multiple smaller PRs.

@jonaslagoni, I think that should be the responsibility of the generator of the language. E.g. in Java, it's natural to use interfaces because it supports multiple, while extend only supports one. I think lots of other programming languages are in a similar situation as Java.

kennethaasan:
@jonaslagoni, please review

jonaslagoni (approved):
👍

PR #1601, opened 2023-11-13
feat: add file path as input

### LINKED ISSUE #1591: File path as input to the generator
#### Reason/Context
I am implementing AsyncAPI for the project that has huge set of REST APIs and Socket Events. To make things modularised and easy to read I have divided a single documentation into multiple files. Here is a reference to the file structure
```
/docs
  - asyncapi.yml
  - servers.yml
  /components
      - component1.yml
      - component2.yml
 /schemas
      - schema1.yml
      - schema2.yml 
```
These files are internally referenced so the modules are easy to read and edit.

Now when I try to create models using `@asyncapi/modelina` it accepts only parsed input. *_Even if I import a file and parse it to proved to the generated it will mess up the references in the file_*

#### Description
- What changes have to be introduced?
Let the `generateToFiles` method of the generator accept the file path as input. And process and parse the file before supplying it to the Renderer.
- Will this be a breaking change?
Yes
- How could it be implemented/designed?
I propose a new method `generateFromFile` which will accept file as input and can output models to multiple files.

nilkanth987:
**Description**

- Added `inputProcessor` option in the AbstractGenerator with `type` enum param
- This option will allow users to pass how to process input to generators
- Added `file` option in input type to allow user to pass file path as input to any generator
- Changed input type in `generateToFiles` method in `AbstractFileGenerator` to any for handling any type of input
- Looking for some input so I can improve the error logging
- Also need to know what are the different types of inputs so I can add them in `InputTypes` enum and handle them.
- **_NOTE: Documentation update is pending_**

**Related issue(s)**
Resolves #1591


src/generators/AbstractGenerator.ts
```diff
@@ -27,6 +28,7 @@ export interface CommonGeneratorOptions<
   defaultPreset?: P;
   presets?: Presets<P>;
   processorOptions?: ProcessorOptions;
+  inputProcessor?: InputProcessorOptions;
```
jonaslagoni, line 31: Please re-use the processorOptions 🙂 Just merge `InputProcessorOptions` into `ProcessorOptions`.

src/generators/AbstractGenerator.ts
```diff
@@ -63,6 +65,16 @@ export interface AbstractGeneratorRenderCompleteModelArgs<
...
 }
 
+export interface InputProcessorOptions {
+  type?: InputType
+}
+
+export enum InputType {
+  FILE = 'file',
+  DOCUMENT = 'document',
+  OBJECT = 'object'
```
jonaslagoni, line 75: I actually dont think we need this specifically 🤔 The idea is good though!

`file` can be detected https://en.wikipedia.org/wiki/File_URI_scheme
`document` can also be detected most of the times https://github.com/asyncapi/modelina/blob/e53c2af1095e020667ee2a14d4f0aa09f5546796/src/processors/AsyncAPIInputProcessor.ts#L393
`object` can also be detected with typeof and the above check.

I think it makes more sense to have `ProcessorType` where you force the input to be loaded as a specific input processor.

src/generators/AbstractGenerator.ts
```diff
@@ -274,6 +285,22 @@ export abstract class AbstractGenerator<
...
+    input: any | InputMetaModel
+  ): Promise<InputMetaModel> {
+    const inputProcessorOptions = this.options.inputProcessor
+    switch (inputProcessorOptions?.type) {
+      case InputType.FILE:
+        const parser = new Parser();
+        const { document, diagnostics } = await fromFile(parser, input).parse();
+        // TODO : Want to log the diagnostics
+        if (!document) throw new Error('Invalid file input')
+        return this.process(document as any)
```
jonaslagoni, line 298: This means that you can move this logic to the AsyncAPI processor where it belongs, because what if some of the other input processors want to support file input as well? Here we hardcode always to parse the input as AsyncAPI which is not the case 🙂

jonaslagoni (changes requested):
Thanks for taking the time to do this @nilkanth987 🙏 Got a few things that are needed to change 🙂

test/generators/testasyncapi.yml
```diff
@@ -0,0 +1,23 @@
...
+      payload:
+        type: object
+        properties:
+          displayName:
+            type: string
+            description: Name of the user
+          email:
+            type: string
+            format: email
+            description: Email of the user
```
TRohit20, line 23: Spec 3.0 is now out :)
```suggestion
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    address: user/signedup
    messages:
      subscribe.message:
        $ref: '#/components/messages/UserSignedUp'
operations:
  user/signedup.subscribe:
    action: send
    channel:
      $ref: '#/channels/user~1signedup'
    messages:
      - $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

TRohit20 (changes requested):

src/generators/AbstractGenerator.ts
nilkanth987: @jonaslagoni , I was hoping to make input as file path and not the URI. I am not aware of whole code so can you tell me if we could have other strings as input. Or can I assume that if there is a string as input then it will be a file path.
- If every string is a file path then we can easily check if the file exists and throw error if not.
- If every string is not a file path we have to check if file exists and if not we have to pass it further to the processor without throwing error.

nilkanth987: Also can you tell me why is the input type is `any`? Cant we make it dynamic to use type effectively?

test/generators/testasyncapi.yml
jonaslagoni: Good point, but we have not added support for it until recently, so let's just stick with v2 in the tests 🙂

src/generators/AbstractGenerator.ts
jonaslagoni: So this all comes down to how we want to process inputs, check out if you have not yet https://github.com/asyncapi/modelina/blob/master/docs/input-processing.md

The reason why we have any is because we support such a wide range of inputs, i.e. Stringified objects, already parsed inputs, [typescript file](https://modelina.org/examples?selectedExample=java-from-typescript-type), etc.

Enabling support for file input has to follow a similar strategy as the one we are using. The reason I suggested URI as the format, is because its super simple to check it for the resolvement strategy that should be used, i.e. for local files you check for `file://` and if it's HTTP for example, we can add support for that as well just by looking at that first part.

So, the question is, how can we integrate file inputs into our current architecture?

My suggestion is that we slightly alter https://github.com/asyncapi/modelina/blob/e53c2af1095e020667ee2a14d4f0aa09f5546796/src/processors/InputProcessor.ts#L48

To check which type of input was provided. Then based on that it will move that input down to the right input processor 🙂 I.e. if it's local file input we can pass it down to the AsyncAPI processor (at least for now) either through the same `process` function or a new one called `processFile` (maybe).

Does that make sense?

test/generators/testasyncapi.yml
TRohit20: Makes sense.

src/generators/AbstractGenerator.ts
nilkanth987: Yes. It makes sense to me. I have updated the code to handle file URI. As a future improvement I suggest we keep the input type dynamic and let Input processor decide its type. This will give a good Intelephense to the end user. 😊

jonaslagoni (changes requested):
You have some linting issues that make it hard to review the PR 🙂

nilkanth987:
@jonaslagoni , Fixed linting issues. Also do I have to use new Parser for `fromFile` method? It does not accept the newParser.

jonaslagoni:
> @jonaslagoni , Fixed linting issues. Also do I have to use new Parser for fromFile method? It does not accept the newParser.

Let me check 🙂

jonaslagoni:
@nilkanth987 when you say new parser what do you mean?

nilkanth987:
@jonaslagoni , This is the new parser added in next branch that I was referring to.

https://github.com/asyncapi/modelina/blob/ef56f3e4a08987bfe262dc37dd025826cb57d7e9/src/processors/AsyncAPIInputProcessor.ts#L18

jonaslagoni:
@nilkanth987 its not a new parser just a bundler that can convert between multiple implementations. 

Just keep what you have 👍

jonaslagoni:
@nilkanth987 its not a new parser just a bundler that can convert between multiple implementations. 

Just keep what you have 👍

nilkanth987:
@jonaslagoni , If everything looks good then should I write documentation for the changes done in PR?

jonaslagoni:
> @jonaslagoni , If everything looks good then should I write documentation for the changes done in PR?

Yes, add an example and documentation for the feature 👌 

Also make sure you fix the linting problem the CI complains about 🙂

nilkanth987:
@jonaslagoni , Added example. Let me know if any issue. Merge it if everything looks good. 🙂

src/generators/AbstractFileGenerator.ts
```diff
@@ -7,7 +7,7 @@ export type FileGenerator = (content: string, toFile: string) => Promise<void>;
  */
 export interface AbstractFileGenerator<RenderCompleteModelOptions> {
   generateToFiles(
-    input: Record<string, unknown> | InputMetaModel,
+    input: any | InputMetaModel,
```
Ferror, line 10: Specifying the type to any kills the idea of typing. Maybe we should add another type instead?

jonaslagoni: I mean we can, but it would be so generic, that even any makes it easier to read.

So to me `any` is probably the best way at the moment.

jonaslagoni:
@nilkanth987 you still have a failing check: https://github.com/asyncapi/modelina/actions/runs/6918808106/job/18831710393?pr=1601

nilkanth987:
@jonaslagoni , Fixed

examples/file-uri-input/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "config" : { "example_name" : "TEMPLATE" },
```
jonaslagoni, line 2:
```suggestion
  "config" : { "example_name" : "file-uri-input" },
```

jonaslagoni (approved):
Nice persistence @nilkanth987! 🔥

jonaslagoni:
@all-contributors please add @nilkanth987 for code, test, example, docs

PR #1602, opened 2023-11-13
fix: make number default type instead of double for float in java

kennethaasan:
**Description**

- This PR makes `Number` the default type instead of `Double` for the Float constrainer in Java. If the format is not specified when the type is `number`, it should allow for both float and double formats. 

I believe this is a fix because this is technically a bug in Modelina today.

**Related issue(s)**

### DIFF
--- a/examples/java-from-typescript-type/__snapshots__/index.spec.ts.snap
+++ b/examples/java-from-typescript-type/__snapshots__/index.spec.ts.snap
@@ -3,13 +3,13 @@
 exports[`Should be able to generate Java model from a TypeScript type file and should log expected output to console 1`] = `
 Array [
   "public class InnerData {
-  private number age;
+  private Number age;
   private String name;
   private boolean free;
   private Map<String, Object> additionalProperties;
 
-  public number getAge() { return this.age; }
-  public void setAge(number age) { this.age = age; }
+  public Number getAge() { return this.age; }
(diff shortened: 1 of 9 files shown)


kennethaasan:
@dlkj, do you agree with this change? I've received this feedback from our Java users.

dlkj:
My main concern would be verifying that `Number` interacts well with common serialization frameworks like Jackson. In Java `Number` is abstract and represents all numerical types, including integers. I understand where the feedback is coming from, but there isn't a direct one to one mapping of the JSON `number` to the Java `Number`.

I'm happy with the current functionality. https://datatracker.ietf.org/doc/html/rfc7159#section-6 acknowledges that implementation are likely to represent numbers as an IEEE double precision floating point number. In Java that is `double`.

dlkj:
We mandated that our uses specify a `format` for all number types. Either `float` or `double`. This ensures that implementations are agreed in what precision will be used in the underlying implementation as JSON `number` is deliberately ambiguous.

kennethaasan:
@dlkj, good points. I made https://github.com/Redocly/openapi-sampler/pull/154 instead. Our biggest issue right now is that generated samples from openapi-sampler generate integers instead of floats. After reading both the OpenAPI and AsyncAPI specs, number should default to a float value.

PR #1613, opened 2023-11-16
feat!: adds extend in common, meta, and constrained models

kennethaasan:
**Description**

This is a preparation for adding inheritance with interfaces for Java in https://github.com/asyncapi/modelina/pull/1593

This PR finalizes what has been started regarding the `allowInheritance` option, and adds `extend` in the CommonModel, MetaModel, and ConstrainedMetaModel. It also adds logic in the Splitter to split our extended models and set `isExtended` in the options where this is needed.

**Related issue(s)**

See also https://github.com/asyncapi/modelina/issues/108.

### DIFF
--- a/docs/migrations/version-2-to-3.md
+++ b/docs/migrations/version-2-to-3.md
@@ -2,6 +2,22 @@
 
 This document contains all the breaking changes and migrations guidelines for adapting your code to the new version.
 
+## allowInheritance set to true will enable inheritance
+
+This feature introduces a new option called `allowInheritance` in the interpreter options, which controls whether the generated models should inherit when the schema includes an `allOf`. By default, this option is set to false, which means that you'll not be affected if this property is not set. In the `MetaModel` and the `ConstrainedMetaModel` options, there is now an `extend` property (a list of models) and an `isExtended` property (boolean).
+
+Here is an example of how to use the new feature and the `allowInheritance` option in your code:
+
+```ts
+const generator = new JavaFileGenerator({
+  processorOptions: {
+    interpreter: {
+      allowInheritance: true
+    }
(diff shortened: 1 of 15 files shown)


jonaslagoni (changes requested):
Remember the migration guide 😄 

Gonna have a closer look within the next 24 hours 👍

src/interpreter/Interpreter.ts
```diff
@@ -50,6 +50,10 @@ export type InterpreterOptions = {
    * When interpreting a schema with discriminator set, this property will be set best by the individual interpreters to make sure the discriminator becomes an enum.
    */
   discriminator?: string;
+  /**
+   * Use this option to disable cache when interpreting schemas. This will affect merging of schemas.
```
jonaslagoni, line 54: Do you mind clarifying a bit more what this option does? For example what cache we are talking about and when to set it to true vs false.

jonaslagoni (changes requested):

src/interpreter/Interpreter.ts
kennethaasan: Done

docs/migrations/version-2-to-3.md
```diff
@@ -2,6 +2,22 @@
...
+```ts
+const generator = new JavaFileGenerator({
+  processorOptions: {
+    interpreter: {
+      allowInheritance: true
+    }
+  }
+});
+```
+
```
jonaslagoni, line 20: Yea, if we show in the subsequent PRs how that code will look in the different generators, that would be nice 👌

jonaslagoni (approved):
Awesome 👍

PR #1626, opened 2023-11-21
feat!: render python union in pydantic in the pre 3.10 way

kennethaasan:
**Description**

See https://www.blog.pythonlibrary.org/2021/09/11/python-3-10-simplifies-unions-in-type-annotations/

To be able to support Python pre 3.10, we need to adjust how types are rendered for union types. 

**Related issue(s)**


jonaslagoni (approved):
You are killing it with all these changes 🔥

PR #1628, opened 2023-11-22
chore(website): create new playground layout

### LINKED ISSUE #1346: Improve layout of playground
#### Reason/Context
It's starting to get a bit tight when it comes to how the playground layout is and I think it can be improved. 

I think can focus on these two areas:

1. With more and more options, it's taking up a lot of space.

image

2. You can never fully see the generated models nor the input making it a weird experience. 
 
image

devilkiller-ag:
**Description**
This is a draft PR for building a new layout of Modelina Playground. The new layout will be similar to the layout of [Async API Studio](https://studio.asyncapi.com/). The first step will be to create a new layout and then I will be working on the logic. I have created a temporary component file `NewPlayground.tsx` to build the new layout. We will replace the old components with the new ones once the layout is ready.

**Related issue(s)**

Resolves #1346


devilkiller-ag:
Hi @jonaslagoni @codingmickey, I was trying to use [React Split Pane](https://www.npmjs.com/package/react-split-pane) for the section which will show input, output, and the output navigation. But it seems that it haven't been updated since last three years and doesn't support React v18. 

I found this new [react-split-pane implementation by Jeremy Grieshop](https://github.com/JeremyGrieshop/react-split-pane)  which supports React v17 and above. It's also been used in the AsyncAPI Studio. Can I use this instead? The problem is that it isn't available as package so I will have to copy paste whole repo as a component as also done in the studio.

jonaslagoni:
@devilkiller-ag I would say no, dont use a tool that has no users 🙂 Then its better to have no split pane functionality.

devilkiller-ag:
Hi @jonaslagoni, I agree with you. For now I have used the code for the split pane from Async API Studio Codebase. But if you want I can try to implement it from scratch for our playground. What's your opinion?

devilkiller-ag:
> > Hi @jonaslagoni, I agree with you. For now I have used the code for the split pane from Async API Studio Codebase. But if you want I can try to implement it from scratch for our playground. What's your opinion?
> 
> You can select the direction you want to take, there is no requirement here 🙂 We don't have split panel right now anyway, so it can be added later if desired 👌 Then all you need to focus on is the basic layout, making it easier to achieve.

Okay Sure, I will go with the basic layout for now 🙂

devilkiller-ag:
I have added the mobile responsive layout without split-pane. Now I will add the Editor, Options, and Output UI.

jonaslagoni:
@devilkiller-ag this is starting to look great!

Only point I have is that it's impossible to see the models + options that are generated. Would it be possible to have the sidebar + input + models + output side by side somehow? 🤔

devilkiller-ag:
Hi @jonaslagoni, I did exactly that. Strangely, it's not visible on your side. Here is the [recording](https://drive.google.com/file/d/1ed2SO7HfOe6PQuNzCDCxL4zof9YY42J4/view?usp=sharing) of the new UI.

95c42a87-2a2d-466d-935c-25a4b60436b7

I have also checked the PR [deploy preview](https://deploy-preview-1628--modelina.netlify.app/), and it's working exactly the same as the local deployment. Are you talking about the sidebar, input, output, and models same as shown in the above gif or have I misunderstood something? I'm sorry for the confusion.

jonaslagoni:
Do you see that when you have the different models visible and what to change the options, the models view disappears?

Thats the part I don't like, i.e. I think the models view should always be visible and maybe right next to the output view 🙂 Similar to what we have in the current modelina playground. We can still have the button in the sidebar to toggle it to be visible and hidden, but it should not be closed just because you want the options 🙂

devilkiller-ag:
Okay Got it. So you want something like this?
Screenshot 2023-12-05 095514

jonaslagoni:
Yes 👍

jonaslagoni:
> Do we need to make the layout mobile responsive? Or Should we ask the user visiting from mobile to visit from desktop? The current layout is not mobile-responsive. Do you have any layout for mobile devices in mind?

I don't think there exists a layout that will work well with a mobile view 🤔 Even in it's current form it's almost impossible to see anything. You could just display all the views on mobile devices like it currently works. I.e. so there is no regression between the current and new version of playground 🤔

devilkiller-ag:
Did you mean, we should display the old playground layout for mobile viewers? Can't we just display something like this: "Revisit from desktop to use Modelina Playground"?

devilkiller-ag:
@jonaslagoni The work is complete! Kindly check it out.

jonaslagoni (changes requested):
Looking great! Got some minor comments 🔥 

Here is what I see and what I base the comments on:
- All the views need to be aligned in height
- The `?` symbol would look better if it blends better with the black background
- I would add more margins to the menu item, as they are too close to the panel to the right.
- The view with the `generator code` and generated models, it's hard to see that  `generator code` can be clicked
- We should make it more easier to see which output model is being selected.
- In the generator options view, for example if you select `kotlin` the panel gets scrollable 🤔 Is there a better way to show long options?

image

devilkiller-ag:
Hi @jonaslagoni,

I have incorporated all the suggestions. Kindly check it out.

modelina-website/src/components/playground/Sidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+    // Options
+    {
+      name: 'options',
+      title: 'Options',
+      isActive: false,
+      onClick: () => {
+        setShowOptions((prevShowOptions) => !prevShowOptions);
+      },
+      icon: <IoOptionsOutline className="w-5 h-5" />,
+      tooltip: 'Option'
```
jonaslagoni, line 30: Think we should add a little more explanation here 🤔 
```suggestion
      tooltip: 'Show or hide all the options'
```

Feel free to go with another text if you have one in mind!

modelina-website/src/components/playground/Sidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+    // Output Explorer
+    {
+      name: 'outputExplorer',
+      title: 'Output',
+      isActive: false,
+      onClick: () => {
+        setShowOutputNavigation((prevShowOutputNavigation) => !prevShowOutputNavigation);
+      },
+      icon: <VscListSelection className="w-5 h-5" />,
+      tooltip: 'Output'
```
jonaslagoni, line 41: Think we should add a little more explanation here 🤔 
```suggestion
      tooltip: 'Show or hide the list of output models'
```

Feel free to go with another text if you have one in mind!

modelina-website/src/components/playground/options/GeneralOptions.tsx
```diff
@@ -66,16 +66,16 @@ class GeneralOptions extends React.Component<
...
 
-  
+
 
 
   render() {
     return (
       <ul className="flex flex-col">
-        <h3 className="text-lg font-medium leading-6 text-gray-900">
+        <h3 className="py-2 w-full text-left border-b-[1px] border-gray-700 text-sm">
```
jonaslagoni, line 75: Setting text-sm and using headers cancel them out 🙂 

```suggestion
        <h3 className="py-2 w-full text-left border-b-[1px] border-gray-700">
```

This is happening in multiple places, included in each language option, I would stick with using appropriate headers and class names 🙂

jonaslagoni (changes requested):
You are fast @devilkiller-ag, only got very few remarks. I think this is the final changes needed, checked all the code and it looks great, only got those small points 💪 Cant wait to merge this 🔥  

Think we can change the width of the sidebar menu to 50: 
image

jonaslagoni:
Once we merge this I would love to offer you code ownership of the `modelina-website` code, let me know if you want to stick around for longer than a single PR 😉 

Let me know if you have any questions about it 👍

devilkiller-ag:
Hi @jonaslagoni, I have made all the changes you suggested, additionally, I have changed the Tooltip color and padding to make it visible. Thanks for offering me the code ownership of the `modelina-website`. I would love to be able to contribute forward. I had an issue in mind that I was going to open just after this PR closes: Currently, the website code uses class components, we should update the code base to use functional components and hooks at all places. This will result in cleaner, easier-to-understand components compared to class components. What's your view on this?

jonaslagoni:
> Hi @jonaslagoni, I have made all the changes you suggested, additionally, I have changed the Tooltip color and padding to make it visible

Perfect 👍 

> I had an issue in mind that I was going to open just after this PR closes: Currently, the website code uses class components, we should update the code base to use functional components and hooks at all places. This will result in cleaner, easier-to-understand components compared to class components. For the Playground, I have updated most of the class components to functional components on which I worked. What's your view on this?

Sounds good, feel free to create an issue for it to track it 🙂

jonaslagoni (approved):
🔥

jonaslagoni:
@all-contributors please add @devilkiller-ag for code, design

Issue #1475, opened 2023-08-15
It should be possible to generate TypeScript interfaces without "reserved" prefix for special keywords


trevordixon:
Instead of generating, for example `reservedStatic`, it could simply quote the property name:

```
interface PayloadFoo {
  'static': string;
  'status': string;
}
```

How hard is it to get this behavior right now?

jonaslagoni:
Super easy, There two ways, this one is probably the best approach, overwriting constraint logic (ignoring reserved keywords) and then overwriting the default rendering of interfaces and the property to follow your syntax `'static': string;`

```ts
const generator = new TypeScriptGenerator({
  constraints: {
    propertyKey: typeScriptDefaultPropertyKeyConstraints({
      // Ignore reserved keywords for properties entirely
      NO_RESERVED_KEYWORDS: (name) => { return name; }
    })
  },
  modelType: 'interface',
  defaultPreset: {
    interface: {
      property: ({property}) => {
        return `'${property.propertyName}': ${property.property.type};`
      }
    }
  }
});
```

Alternatively, a smaller but a little more hacky approach is just using the unconstrained property name as such:

```ts
const generator = new TypeScriptGenerator({
  modelType: 'interface',
  defaultPreset: {
    interface: {
      property: ({property}) => {
        return `'${property.unconstrainedPropertyName}': ${property.property.type};`
      }
    }
  }
});
```

And that's it 🙂 

I wonder if it makes sense to make this the default outcome for interfaces? 🤔 Is there any downside you see @trevordixon?

trevordixon:
Thank you so much! That helps a ton. No downside to doing this by default
as far as I know. It seems possible to do with classes too by the way.

```
class Foo {
  get 'static'() { return 1; }
}
(new Foo).static;
```

I think this would be preferable as well for classes, though not as huge a
deal with generated marshal/unmarshal. But with interfaces, it's critical
for me that the interface describes what's actually coming down the wire.

jonaslagoni:
Solved in https://github.com/asyncapi/modelina/pull/1729

PR #1667, opened 2023-12-19
feat!: add options as parameter to constraints

jonaslagoni:
**Description**
This PR adds options as part of the constraint callbacks so you have direct access to them when the models are being constrained, i.e. for conditionally changing the behavior.

This is only a breaking change if you use the current constraint types in TS integrations.

**Related issue(s)**
Part of solving https://github.com/asyncapi/modelina/issues/1475


.eslintrc
```diff
@@ -55,7 +56,7 @@
     "no-empty-character-class": 2,
     "no-self-compare": 2,
     "valid-typeof": 2,
-    "no-unused-vars": 0,
+    "unused-imports/no-unused-imports": 2,
```
jonaslagoni, line 59: Had to change lint rule because the old one did not understand `--fix` option

kennethaasan (approved):
LGTM

Samridhi-98 (approved):
LGTM 🎉

Ferror (approved):
PHP ok

jonaslagoni:
Thanks for the review folks 🙇

PR #1748, opened 2024-01-16
chore: add python runtime tests

### LINKED ISSUE #1745: Add runtime tests for Python
#### Reason/Context

To properly test that the models that are generated are semantically correct, we need to test them in their specific runtime. This means that we need to add the setup for Python.

We already have a bunch of existing runtime tests for the other languages, that you can use as reference implementation: https://github.com/asyncapi/modelina/tree/master/test/runtime

Here is the TODO to solve this issue:
- Create a simple Python project under `runtime-python` that has a simple testing setup 
- Add a simple generation script called `runtime-python.ts` that generates the models: https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-kotlin.ts
- Add a simple TS test file that is used in the ecosystem to test the runtime: https://github.com/asyncapi/modelina/blob/master/test/runtime/runtime-kotlin.spec.ts
- Add [two scripts to the package.json](https://github.com/asyncapi/modelina/blob/c4b83f2d084add0c002325bc7a955a2cdc0e8c40/package.json#L110) file called `test:runtime:python` and `generate:runtime:python`
- Add a GitHub workflow file that spins up the test: https://github.com/asyncapi/modelina/blob/master/.github/workflows/runtime-kotlin-testing.yml
- Manually add a dart test that makes sure that the generated python model can be used as expected.

That should be it to add runtime testing.

If you encounter any problems just reach out :v:

AkhilJ321:
## Description
Added runtime tests for python

## Related Issue
Fixes #1745


jonaslagoni (commented):
Looking good 👍

AkhilJ321:
@jonaslagoni , can you review this PR

test/runtime/runtime-python.ts
```diff
@@ -0,0 +1,22 @@
+import { PYTHON_DEFAULT_PRESET,PythonFileGenerator } from "../../";
+import path from "path";
+
+import input from "./generic-input.json";
+
+const generator = new PythonFileGenerator({
+    presets: [PYTHON_DEFAULT_PRESET],
+    
+});
```
jonaslagoni, line 9: Remember to generate the models with serialization support: https://modelina.org/docs/languages/python#generate-models-with-json-serializer-and-deserializer-methods

.github/workflows/runtime-python-testing.yml
```diff
@@ -0,0 +1,36 @@
+name: Runtime testing Python Models
+
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'src/generators/python/**'
+      - 'test/runtime/runtime-python/**'
+      - 'test/runtime/**kotlin**'
```
jonaslagoni, line 10:
```suggestion
```

jonaslagoni (changes requested):
Let's focus on making the workflow succeed 🙂

AkhilJ321:
@jonaslagoni , thank you for the reviews, added new changes

test/runtime/runtime-python/test/main.py
```diff
@@ -0,0 +1,37 @@
...
+        self.members = members
+        self.array_type = array_type
+        self.nested_object = nested_object
+
+    def to_json(self):
+        return json.dumps(self.__dict__)    
+    
+class NestedObject:
+    def __init__(self, test):
+        self.test = test
```
jonaslagoni, line 18: You need to remove this definition.

It should be generated models used here instead. I.e. what you generate with https://github.com/asyncapi/modelina/blob/57787cb77068c3da92f03f5bb558c5c5d74b3d1a/test/runtime/runtime-python.ts

jonaslagoni (changes requested):
Almost there it seems 👍

test/runtime/runtime-python/test/main.py
AkhilJ321: @jonaslagoni Oh, sorry my bad. Fixed this now

test/runtime/runtime-python/test/main.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+import json
+from main.Address import Address
+from main.NestedObject import NestedObject
```
jonaslagoni, line 4: Seems that the CI cannot access the models. Are you able to run the commands locally? You can check the error here: https://github.com/asyncapi/modelina/actions/runs/7635660479/job/20805641740?pr=1748

jonaslagoni (changes requested):

test/runtime/runtime-python/test/main.py
AkhilJ321: ` npm run test:runtime:python` this command is causing error : 
` > cross-env CI=true jest ./test/runtime/runtime-python.spec.ts`
`FAIL test/runtime/runtime-python.spec.ts`
`  ● Test suite failed to run`
`Bindings not found`

jonaslagoni: You need to be able to run the scripts locally before the CI can run it as well 🙂

AkhilJ321: @jonaslagoni , there were some issues in the testing script ,I resolved them , now the python testing script returns output  :
..
----------------------------------------------------------------------
Ran 2 tests in 0.000s

OK

But the exec command gets this output in stderr . I have been stuck on this part.

jonaslagoni: Did you push the changes? 🤔

AkhilJ321: pushed now

test/runtime/runtime-python.ts
```diff
@@ -0,0 +1,21 @@
...
+const generator = new PythonFileGenerator({
+    presets: [PYTHON_JSON_SERIALIZER_PRESET],
+});
+
+generator.generateToFiles(
+    input,
+    path.resolve(
+        // eslint-disable-next-line no-undef
+        __dirname,
+        "./runtime-python/test/main/"
```
jonaslagoni, line 15: Is this not supposed to be for the src code not the test?
```
        "./runtime-python/src/main/"
```

test/runtime/runtime-python/test/main.py
AkhilJ321: @jonaslagoni  
There are certain problems :
1.The generated code which is added by generator contain two files, 
Address.py and NestedObject.py  
Screenshot from 2024-02-07 16-23-07

So when this file runs ,I get this error:
Screenshot from 2024-02-07 16-25-21

So when running locally I removed this import manually. 

2.I added a __init__.py file manually in the generated folder to make that as module so it can be imported.This thing is failing in the CI also .

AkhilJ321: @jonaslagoni  
There are certain problems :
1.The generated code which is added by generator contain two files, 
Address.py and NestedObject.py  
Screenshot from 2024-02-07 16-23-07

So when this file runs ,I get this error:
Screenshot from 2024-02-07 16-25-21

So when running locally I removed this import manually. 

2.I added a __init__.py file manually in the generated folder to make that as module so it can be imported.This thing is failing in the CI also .

jonaslagoni: > So when running locally I removed this import manually.

You are not allowed to manually make changes to the generated models (cause the manual changes will be overwritten next time you run the command). 

You should setup the project so that the test files and other project files i.e. init.py etc are added manually, but the generated models are used as is with no other manual changes needed.

Is this not possible with the current generated models?

> 2.I added a init.py file manually in the generated folder to make that as module so it can be imported.This thing is failing in the CI also .

This has not been committed?

AkhilJ321: @jonaslagoni  The whole main folder is getting generated by the script so how can I add init.py manually?
The first screenshot which I attached is totally associates with the generated code and it give import errors,So I am not getting what can I do for that

jonaslagoni: You can manually add the main folder and the init script as you see fit 👍 The script just generates the models next to it.

AkhilJ321: @jonaslagoni 
I have added some fixes, so the script should run now . 
1.Python script is returning OK ran 2 tests but the script is getting it as error.

test/runtime/runtime-python.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+jest.setTimeout(50000);
+
+test("Python runtime testing", async () => {
+//   The 'python ' command here
+
+    const compileCommand = `cd ${path.resolve(
+        __dirname,
+        "./runtime-python/test/"
+    )} && python3 main.py`;
+    await execCommand(compileCommand);
```
jonaslagoni, line 14: Seems like this test fails because of the execCommand fails or exits with status code 1 i.e. indicating a failure 🤔

jonaslagoni (changes requested):

jonaslagoni:
@AkhilJ321 have you had time to take a look at the problem above?

AkhilJ321:
> @AkhilJ321 have you had time to take a look at the problem above?

yes, tried debugging it in execCommand code also, guess its some issue with the way Python tests give output.
will try to get a proper solution for this soon.

jonaslagoni (changes requested):
After manually running your test command, you should be able to get the exit code from the command using `echo $?` to check whether it is a problem i.e. that it returns 1 when tests are successful.

jonaslagoni:
It also slightly looks like that there is an exception thrown that the test ignores (based on the CI logs)

jonaslagoni:
@AkhilJ321 have you giving up on the effort here?

AkhilJ321:
> @AkhilJ321 have you giving up on the effort here?

Sorry for not reaching out to you.Earlier when I made changes, and then fetched the latest results from the repository. From then I wasn't able to run the tests command 
image

Tried resolving this issue but nothing helped.

AkhilJ321:
> > After manually running your test command, you should be able to get the exit code from the command using `echo $?` to check whether it is a problem i.e. that it returns 1 when tests are successful.
> 
> It returns 0 on successful execution.
But python tests also return some information even on successful execution which is coming in stderr. so I made allowStdError =true.
So now this thing is resolved.

jonaslagoni (approved):

jonaslagoni:
Awesome @AkhilJ321 :v:

jonaslagoni:
@all-contributors please add @AkhilJ321  for code, test

PR #1782, opened 2024-02-06
fix: partial and playground

jonaslagoni:
## Description
this PR fixes a problem with playground and local running build where the implementation does not work 🤨 Dont think I figured out the root cause, but this at least fixes the problem.


devilkiller-ag (approved):
Looks good to me!

jonaslagoni:
@devilkiller-ag when its a draft feel free to ignore it as its still in progress ✌️

Thanks for the hastely review though!

devilkiller-ag:
Okay Sure 😄

jonaslagoni:
Split out the PR into two

PR #1788, opened 2024-02-12
fix: partial and playground

jonaslagoni:
## Description
This PR fixes the implementation for partial that incorrectly handles array values, and thereby fixing the playground and others that dependent on Modelina.

The problem arose when the array values contained instances of classes that would be re-created incorrectly. We only intend to merge the default and target together.


jonaslagoni:
Because of the nature of the bug, I am gonna force merge this @kennethaasan @magicmatatjahu, if anything comes up after or you review it with comments I will of course provide the necessary changes :v:

PR #1799, opened 2024-02-13
fix: dont render unmarshal for property with const

kennethaasan:
## Description

There seems to be a bug introduced in https://github.com/asyncapi/modelina/pull/1717/files. Properties that have `const` set are read-only, and can't be set when unmarshaling. If this is not fixed, then the TS compiler fails on the generated code.

## Checklist


Samridhi-98 (approved):
LGTM :)

PR #1800, opened 2024-02-13
feat: export and import interfaces as types when using named imports and ESM Module Syntax

### LINKED ISSUE #1793: Support verbatimModuleSyntax in TypeScript Generator
#### Reason/Context
Currently the typescript generator does not explicitly export/import interfaces as types. This is an issue when the typescript option 'verbatimModuleSyntax' is enabled and can potentially cause some other pitfalls.

#### Description
Add a new flag (or make this the default behaviour when using interfaces, afaik that should not cause any issues) that changes the behaviour of 'src/generators/typescript/TypeScriptDependencyManager.ts' so it imports and exports interfaces as types.

Before:
```typescript
import { PublicUser } from './PublicUser'
interface NewStatePayload {
  members: PublicUser[]
}
export { NewStatePayload }
```
After:
```typescript
import type { PublicUser } from './PublicUser'
interface NewStatePayload {
  members: PublicUser[]
}
export type { NewStatePayload }
```

moritzkalwa:
## Description
When using modelType 'interface', exportType 'named' and moduleSystem 'ESM' export and import interfaces as types to support verbatimModuleSyntax.
This should not be a breaking change, interfaces are types even when not using verbatimModuleSyntax, explicitly marking them as types is optional in that case though.

## Related Issue
fixes #1793 

## Checklist

## Additional Notes
We might want to document the fact that one needs to use named exports when using interfaces and verbatimModuleSyntax. Im not sure where this documentation should go though, please point me in the right direction here :D


test/generators/typescript/TypeScriptGenerator.spec.ts
```diff
@@ -128,8 +133,10 @@ ${content}`;
     };
 
     generator = new TypeScriptGenerator({ modelType: 'interface' });
-    const models = await generator.generate(doc);
-    expect(models).toHaveLength(1);
+    const models = await generator.generateCompleteModels(doc, {
+      exportType: 'named'
+    });
+    expect(models).toHaveLength(2);
```
jonaslagoni, line 139: If you want to test the generated complete models, please create a test just for this 🙂 As it's two different renderings and what they test :v:

jonaslagoni (changes requested):

moritzkalwa:
@jonaslagoni Do you think it would be okay to change [this example](https://github.com/asyncapi/modelina/blob/master/examples/typescript-interface/index.ts) to use exportType 'named' and moduleSystem 'ESM' to show these changes?

jonaslagoni:
I would duplicate it and create one that shows it with those settings 🙂 You can follow this: https://github.com/asyncapi/modelina/blob/master/docs/contributing.md#adding-examples

examples/typescript-interfaces-as-types/README.md
```diff
@@ -0,0 +1,17 @@
+# TypeScript interface
```
jonaslagoni, line 1:
```suggestion
# TypeScript interface as types
```

jonaslagoni (changes requested):
Just a small change

jonaslagoni (approved):

mikaellanger:
This makes enum values unusable when using enumType 'enum' since it also exports enums as types not just interfaces, leading to
`MyEnum' cannot be used as a value because it was exported using 'export type'.`
when doing e.g. MyEnum.AnEnumValue

jonaslagoni:
Thanks for raising the issue @mikaellanger

Reverting commit in master.

Taking a second look, this should be targeting `next` because of the breaking changes in the dependency manager: https://github.com/asyncapi/modelina/pull/1800/files#diff-95c504c5267dbd4e90a8ee687814eabaf191637b94093293b968648836d91d39R21

PR #1801, opened 2024-02-13
feat!: properly implement constants in c#

### LINKED ISSUE #1471: Constants are compiled as enums
#### Describe the bug
If I have a value that's a constant on an object, it will get compiled into an anonymous schema. This overcomplicates things by a lot.

Input:
```yaml
    TradeMessageClient:
      name: Trade Message (client)
      title: Trade Message (client)
      description: A message containing trade data
      payload:
        title: TradePayloadClient
        type: object
        properties:
          operation:
            type: string
            const: Trade
          data:
            $ref: '#/components/schemas/TradeClient'
        additionalProperties: false
        required:
          - operation
          - data
```

Expected TypeScript output:
```typescript
interface TradePayloadClient {
  operation: "Trade"
  data: TradeClient;
}
```

Actual Typescript Output:
```typescript
interface TradePayloadClient {
  operation: AnonymousSchema_11;
  data: TradeClient;
}
enum AnonymousSchema_11 {
  TRADE = "Trade",
}
```

Expected C# output:
```c#
public class TradePayloadClient
{
    public const string Operation = "Trade";
    public TradeClient? Data { get; set; }
}
```

Actual C# output:
```c#

public class TradePayloadClient
  {
    public AnonymousSchema_11? Operation { get; set; }
    public TradeClient? Data { get; set; }
  }

public enum AnonymousSchema_11
  {
    TRADE
  }

  public static class AnonymousSchema_11Extensions
  {
    public static string? GetValue(this AnonymousSchema_11 enumValue)
    {
      switch (enumValue)
      {
        case AnonymousSchema_11.TRADE: return "Trade";
      }
      return null;
    }

    public static AnonymousSchema_11? ToAnonymousSchema_11(string? value)
    {
      switch (value)
      {
        case "Trade": return AnonymousSchema_11.TRADE;
      }
      return null;
    }
  }
```

moritzkalwa:
## Description
Enable the C# Class and Record Generators to properly handle constant values.

## Related Issue
fixes #1471 

## Checklist

### DIFF
--- a/docs/migrations/version-3-to-4.md
+++ b/docs/migrations/version-3-to-4.md
@@ -29,3 +29,33 @@ interface AnonymousSchema_1 {
   aa_00_testAttribute?: string;
 }
 ```
+
+## C#
+
+### Constant values are now properly rendered as const properties
+
+This example used to generate a `string` with a getter and setter, but will now generate a const string that is initialized to the const value provided. 
+
+```yaml
+type: object
+properties:
+  property:
+    type: string
+    const: 'abc'
+```
+
+will generate
+
+```csharp
+public class TestClass {
+  private const string property = "test";  
+  
+  public string Property 
+  {
+    get { return property; }
+  }
+  ...
+}
+```
+
+Notice that `Property` no longer has a `set` method. This might break existing models.
--- a/src/generators/csharp/constrainer/ConstantConstrainer.ts
+++ b/src/generators/csharp/constrainer/ConstantConstrainer.ts
@@ -1,7 +1,53 @@
+import {
+  ConstrainedEnumModel,
+  ConstrainedMetaModel,
+  ConstrainedMetaModelOptionsConst,
+  ConstrainedReferenceModel,
+  ConstrainedStringModel
+} from '../../../models';
 import { CSharpConstantConstraint } from '../CSharpGenerator';
 
+const getConstrainedEnumModelConstant = (args: {
+  constrainedMetaModel: ConstrainedMetaModel;
+  constrainedEnumModel: ConstrainedEnumModel;
+  constOptions: ConstrainedMetaModelOptionsConst;
+}) => {
+  const constrainedEnumValueModel = args.constrainedEnumModel.values.find(
+    (value) => value.originalInput === args.constOptions.originalInput
+  );
+
+  if (constrainedEnumValueModel) {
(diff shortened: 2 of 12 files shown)


jonaslagoni:
I think we have to do this change on the `next` branch, as the serializer currently in place will break as they will try assign values to the constants 🤔

jonaslagoni:
Similar to https://github.com/asyncapi/modelina/pull/1799

moritzkalwa:
@jonaslagoni No, I can definitely add a migration guide for this. I probably won't have the time for it until Friday though.

jonaslagoni:
> @jonaslagoni No, I can definitely add a migration guide for this. I probably won't have the time for it until Friday though.

No worries, no rush :v:

jonaslagoni:
Changed the target branch to `next` :v:

moritzkalwa:
> I think we have to do this change on the `next` branch, as the serializer currently in place will break as they will try assign values to the constants 🤔 I think that either we add an issue after this PR is merged, or we adapt the presets where needed.

If you point me to the correct place I might be able to adapt the serializers right in this PR / add a second PR before we merge this one?

jonaslagoni:
You have the presets here: https://github.com/asyncapi/modelina/blob/master/src/generators/csharp/presets/JsonSerializerPreset and https://github.com/asyncapi/modelina/blob/master/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts

I think you only need to make changes to renderSerializeProperty and renderDeserializeProperty and check whether the model has a constant value, if so, dont serialize what you get, just set it 🙂

moritzkalwa:
> You have the presets here: https://github.com/asyncapi/modelina/blob/master/src/generators/csharp/presets/JsonSerializerPreset and https://github.com/asyncapi/modelina/blob/master/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts
> 
> I think you only need to make changes to renderSerializeProperty and renderDeserializeProperty and check whether the model has a constant value, if so, dont serialize what you get, just set it 🙂

Im still trying to understand the serializer so forgive me if this is a dumb question, but do we really need to change the `renderDeserializeProperty` function? Writing to the constant values would be a problem for sure, but deserialising shouldn't be a problem, because it only reads that property, right?

jonaslagoni:
Most likely correct @moritzkalwa 👍 you can use the runtime test to ensure it works 😄

PR #1821, opened 2024-02-17
fix: python generator type safety

### LINKED ISSUE #1810: Python generator does not have type safetype on dependency manager
#### Describe the bug
When trying to overwrite the dependency manager for Python it does not have the correct types, meaning its hard to write a custom one.

#### How to Reproduce
While trying to use the following generator code:
```
const generator = new PythonFileGenerator({ 
    importsStyle: 'explicit',
    dependencyManager: () => {
      return {
        addDependency: () => {},
        dependencies: []
      }
    }
  });
```

It only suggest the abstract dependency manager type and not the python one.

moritzkalwa:
## Description
Correctly type the dependencyManager option for the PythonGenerator

## Related Issue
Fixes #1810 

## Checklist

### DIFF
--- a/src/generators/python/PythonGenerator.ts
+++ b/src/generators/python/PythonGenerator.ts
@@ -35,11 +35,11 @@ import {
 import { DeepPartial, mergePartialAndDefault } from '../../utils/Partials';
 import { PythonDependencyManager } from './PythonDependencyManager';
 
-export interface PythonOptions extends CommonGeneratorOptions<PythonPreset> {
+export interface PythonOptions
+  extends CommonGeneratorOptions<PythonPreset, PythonDependencyManager> {
   typeMapping: TypeMapping<PythonOptions, PythonDependencyManager>;
   constraints: Constraints<PythonOptions>;
   importsStyle: 'explicit' | 'implicit';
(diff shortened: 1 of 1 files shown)


jonaslagoni (changes requested):
Because this option is part of the AbstractGenerator, https://github.com/asyncapi/modelina/blob/6338a6b8ab1d2cabf246bdbf4fed3a1751682854/src/generators/python/PythonGenerator.ts#L38C40-L38C62 

```ts
export interface PythonOptions extends CommonGeneratorOptions<PythonPreset, PythonDependencyManager> {
```

I think this is all that's needed 🤔

moritzkalwa:
> Because this option is part of the AbstractGenerator, https://github.com/asyncapi/modelina/blob/6338a6b8ab1d2cabf246bdbf4fed3a1751682854/src/generators/python/PythonGenerator.ts#L38C40-L38C62
> 
> ```ts
> export interface PythonOptions extends CommonGeneratorOptions<PythonPreset, PythonDependencyManager> {
> ```
> 
> I think this is all that's needed 🤔

Oh my, you're right. Sorry, fixed it :D

jonaslagoni (approved):
This should be enough yea 👍

PR #1834, opened 2024-02-26
fix!: mapping for C# object date

JFCote:
## Description

There was a problem in the recent addition of mapping for `date` and `date-time` field to C#.

Following AsyncAPI documentation:
image

And the RFC documentation:
image

My simple change rectify this situation.

## Checklist
- [x] Documentation has been updated to reflect the changes. (**not required**)

### DIFF
--- a/docs/migrations/version-3-to-4.md
+++ b/docs/migrations/version-3-to-4.md
@@ -59,3 +59,29 @@ public class TestClass {
 ```
 
 Notice that `Property` no longer has a `set` method. This might break existing models.
+
+### DateTime and DateTimeOffset are now properly rendered based on specification format
+
+In the previous version, `date-time` and `date` formats were rendered as `DateTime` and `DateTimeOffset` respectively. 
+This has been changed to render `DateTimeOffset` for `date-time` and `DateTime` for `date` formats.
+
+This might break existing implementation and require manual changes.
+
+The best thing to do is to fix your specification and use what you really need. If you don't care about the time and time zone, use `date` instead of `date-time`.
+Otherwise, keep the `date-time` format and update your code to use `DateTimeOffset` instead of `DateTime`.
+That usually means doing this:
+
+```csharp
+var dateTime = new DateTime(2008, 6, 19, 7, 0, 0);
+
+// Set the DateTime property of the ModelinaModel
+var modelinaModel = new ModelinaModel();
+modelinaModel.DateTime = dateTime;
+Console.WriteLine(modelinaModel.DateTime);
+
+// Get the DateTime property from the ModelinaModel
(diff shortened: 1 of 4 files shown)


jonaslagoni (commented):
@JFCote dont you think it would be better to target `next` branch, as this will break usage of models (even though it is a bug)? 🤔

JFCote:
@jonaslagoni I honestly think that the previous change was breaking (making `date-time` à DateTime in C#) because it can't contains the time zone / offset information. All our microservices using modelina in the code generator need to pass the offset thus making the new version unusable.

Anyway, DateTimeOffset is a "bigger" object than "DateTime" as it contains more information. People who currently use Modelina with DateTime will be able to cast a DateTime to DateTimeOffset and vice-versa as you just have to ignore the offset. For people in my situation, right now there is no possible solution than to roll back or wait for the next version.

I'm not the owner of the project but if it was mine, I would fix it directly in the main branch as soon as possible. :)

I'll let you decide!

P.S: Another solution for user of DateTime only is to change their spec to `date` instead of `date-time`.

jonaslagoni:
My only concern is that people currently using Modelina to generate models and are using them, should always be able to upgrade minor and patch version of the tool. 

So as long as this code is valid after this PR is merged, then we can do it on the `master` branch 👍 

```c#
DateTime dateTime = new DateTime(2008, 6, 19, 7, 0, 0);
ModelinaModel model = new ModelinaModel();
model.dateTime = dateTime;
```

But if this change, breaks their implementation, i.e. having to now cast the type, then it's a breaking change and we should target `next` instead (it's released immediately, so it should not change anything for you). You can read more about it here: https://github.com/asyncapi/modelina?tab=readme-ov-file#versioning-and-maintenance

JFCote:
@jonaslagoni 
I did the test and your use case do not need casting as it is implicit but on the other side, it needs casting. Depending if their implementation is publishing or subscribing, it will need a change.

image

So I guess the better thing to do it to create the PR against the next major version.
Updating the PR !

jonaslagoni:
Ignore the extra commits, gonna be fixed by https://github.com/asyncapi/modelina/pull/1837 in a sec

jonaslagoni (changes requested):
Can you add this change to the migration guide about the breaking change? We already have one for v3->v4

JFCote:
@jonaslagoni I can't find a way to rebase on the `next` branch... My fork of the main project do not contain the `next` branch even if I sync it completely. Any idea how I can achieve this?

jonaslagoni:
Because we squash commits, the only problem is that your branch is not up to date with the latest `next` changes, so should be as simple as merging them into your branch if I recall correctly 🤔

jonaslagoni (approved):
Thanks for taking the time to fix this! 💯

Issue #1817, opened 2024-02-15
Python generate models with inconsistant indentation 


jonaslagoni:
#### Describe the bug
When generating models with the Python generator it complains about `Inconsistent use of tabs and spaces in indentation` throughout the models.

PR #1856, opened 2024-03-04
fix: inconsistent indentation for Python models

jonaslagoni:
## Description
This PR fixes the inconsistent indentation for the Python models where explicit `\t` characters where used.

## Related Issue
Fixes https://github.com/asyncapi/modelina/issues/1817

### DIFF
--- a/examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
@@ -16,14 +16,14 @@ class ObjProperty:
       return self._number
     @number.setter
     def number(self, number):
-    	self._number = number
+      self._number = number
 
     @property
     def additionalProperties(self):
       return self._additionalProperties
     @additionalProperties.setter
     def additionalProperties(self, additionalProperties):
-    	self._additionalProperties = additionalProperties
+      self._additionalProperties = additionalProperties
 ",
 ]
 `;
@@ -44,14 +44,14 @@ class ObjProperty:
       return self._number
     @number.setter
     def number(self, number):
-    	self._number = number
+      self._number = number
 
     @property
     def additionalProperties(self):
       return self._additionalProperties
     @additionalProperties.setter
     def additionalProperties(self, additionalProperties):
-    	self._additionalProperties = additionalProperties
+      self._additionalProperties = additionalProperties
 ",
 ]
 `;
@@ -72,14 +72,14 @@ class Root:
       return self._email
     @email.setter
     def email(self, email):
-    	self._email = email
+      self._email = email
 
     @property
     def objProperty(self):
       return self._objProperty
     @objProperty.setter
     def objProperty(self, objProperty):
-    	self._objProperty = objProperty
+      self._objProperty = objProperty
 ",
 ]
 `;
@@ -100,14 +100,14 @@ class Root:
       return self._email
     @email.setter
     def email(self, email):
-    	self._email = email
+      self._email = email
 
     @property
     def objProperty(self):
       return self._objProperty
     @objProperty.setter
     def objProperty(self, objProperty):
-    	self._objProperty = objProperty
+      self._objProperty = objProperty
 ",
 ]
 `;
--- a/examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
@@ -8,8 +8,8 @@ class ObjProperty:
   def __init__(self, input):
     if hasattr(input, 'number'):
       self._number = input.number
-    if hasattr(input, 'additionalProperties'):
-      self._additional_properties = input.additionalProperties
+    if hasattr(input, 'additional_properties'):
+      self._additional_properties = input.additional_properties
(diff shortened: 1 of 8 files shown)


jonaslagoni:
@dimitern tagging you as @kennethaasan said you are the Python maestro and did not mind helping out reviewing these :v:

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
```diff
@@ -7,23 +7,23 @@ Array [
...
+    @number.setter
+    def number(self, number):
+      self._number = number
+
+    @property
+    def additionalProperties(self):
+      return self._additionalProperties
+    @additionalProperties.setter
+    def additionalProperties(self, additionalProperties):
+      self._additionalProperties = additionalProperties
```
dimitern, line 26: This does not look right - you want to define a property `number` inside the `ObjProperty` class, right? If so, it needs to be at the same identation as the `def __init__` is, otherwise it would be considered as part of the constructor (nested functions inside `__init__()`).

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
```diff
@@ -35,23 +35,23 @@ Array [
...
```
dimitern, line 54: Same as above

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
```diff
@@ -63,23 +63,23 @@ Array [
...
+    @email.setter
+    def email(self, email):
+      self._email = email
+
+    @property
+    def objProperty(self):
+      return self._objProperty
+    @objProperty.setter
+    def objProperty(self, objProperty):
+      self._objProperty = objProperty
```
dimitern, line 82: Same as above, it needs to look something like this:

```python
class Root:
  def __init__(self, input):
    if hasattr(input, 'email'):
      self._email = input.email
    if hasattr(input, 'objProperty'):
      self._objProperty = input.objProperty

  @property
  def email(self):
    return self._email

  @email.setter
  def email(self, email):
    self._email = email
    
  @property
  def objProperty(self):
    return self._objProperty

  @objProperty.setter
  def objProperty(self, objProperty):
    self._objProperty = objProperty
```

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
```diff
@@ -91,23 +91,23 @@ Array [
...
```
dimitern, line 110: Same here

examples/generate-python-models/__snapshots__/index.spec.ts.snap
```diff
@@ -5,14 +5,14 @@ Array [
...
-  	return self._email
-  @email.setter
-  def email(self, email):
-  	self._email = email
+    @property
+    def email(self):
+      return self._email
+    @email.setter
+    def email(self, email):
+      self._email = email
```
dimitern, line 15: And here

examples/python-generate-json-serializer-and-deserializer/__snapshots__/index.spec.ts.snap
```diff
@@ -5,23 +5,23 @@ Array [
...
-  	return self._additionalProperties
-  @additionalProperties.setter
-  def additionalProperties(self, additionalProperties):
-  	self._additionalProperties = additionalProperties
+    @property
+    def additionalProperties(self):
+      return self._additionalProperties
+    @additionalProperties.setter
+    def additionalProperties(self, additionalProperties):
+      self._additionalProperties = additionalProperties
```
dimitern, line 24: and here

src/generators/python/renderers/ClassRenderer.ts
```diff
@@ -83,7 +83,8 @@ export const PYTHON_DEFAULT_CLASS_PRESET: ClassPresetType<PythonOptions> = {
     if (Object.keys(properties).length > 0) {
       const assigments = Object.values(properties).map((property) => {
         if (!property.required) {
-          return `if hasattr(input, '${property.propertyName}'):\n\tself._${property.propertyName} = input.${property.propertyName}`;
+          return `if hasattr(input, '${property.propertyName}'):
+  self._${property.propertyName} = input.${property.propertyName}`;
```
dimitern, line 87: Adding the newline was correct, but you need to use 2 spaces after it instead of tab. Python can take any indents with spaces, but need to be consistent within the same file. It's not quite clear here if there are spaces before the `self._...` concatenation.

dimitern (changes requested):
All the ".snap" files show incorrect identation, that will cause incorrect code to be generated for the class. I suggest using explicit leading spaces for indent in the code you concatenate generated bits of code

jonaslagoni:
> All the ".snap" files show incorrect identation, that will cause incorrect code to be generated for the class. I suggest using explicit leading spaces for indent in the code you concatenate generated bits of code

Yea I see it 🧐 Looking at it now, good eyes :v:

jonaslagoni:
@dimitern should be all fixed now :v:

PR #1858, opened 2024-03-04
fix!: add python types, contructor and constants

jonaslagoni:
## Description
This PR introduces 3 changes to the Python generator.
1. Type hints are now added to the properties and arguments.
2. The constructor now correctly initializes properties with the right models
3. Constants are now rendered for Python


jonaslagoni:
@dimitern another one.

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
```diff
@@ -5,24 +5,24 @@ Array [
   "
 
 class ObjProperty: 
-  def __init__(self, input):
-    if hasattr(input, 'number'):
-    	self._number = input.number
-    if hasattr(input, 'additionalProperties'):
-    	self._additionalProperties = input.additionalProperties
+  def __init__(self, input: dict):
```
dimitern, line 8: Ideally, for backwards compatibility we should use `Dict` instead of `dict` here, and same goes for `List` vs `list` in type annotations. Both of those need `from typing import Dict` or `List`

src/generators/python/renderers/ClassRenderer.ts
```diff
@@ -81,27 +82,38 @@ export const PYTHON_DEFAULT_CLASS_PRESET: ClassPresetType<PythonOptions> = {
...
+      const assignments = Object.values(properties).map((property) => {
+        let assignment: string;
+        if (property.property instanceof ConstrainedReferenceModel) {
+          assignment = `self._${property.propertyName}: ${property.property.type} = ${property.property.type}(input["${property.propertyName}"])`;
+        } else {
+          assignment = `self._${property.propertyName}: ${property.property.type} = input["${property.propertyName}"]`;
+        }
         if (!property.required) {
-          return `if hasattr(input, '${property.propertyName}'):\n\tself._${property.propertyName} = input.${property.propertyName}`;
+          return `if hasattr(input, "${property.propertyName}"):\n\t${assignment}`;
```
dimitern, line 93: Tabs are not recommended for indentation, I'd suggest to use 2 or 4 spaces instead, but consistently.

dimitern (changes requested):
A few minor comments, I hope those can be made for better compatibility

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
jonaslagoni: Do you think we should include it in this PR, or another? 🤔

src/generators/python/renderers/ClassRenderer.ts
jonaslagoni: https://github.com/asyncapi/modelina/pull/1856 clears this up :v:

examples/generate-python-complete-models/__snapshots__/index.spec.ts.snap
jonaslagoni: Changed it :v:

jonaslagoni:
@dimitern should be all fixed now :v:

Had to work a bit with the dependencies to not include duplicate stuff.

PR #1865, opened 2024-03-06
fix: switch to Object.Equals for C# Equals override

### LINKED ISSUE #1850: Change C# Equals() overload to allow reference type properties to be properly compared
We need this improvement because the current Equals() overload is functionally pointless. It only works if the properties of a given model are all value types or if the model and the object it is being compared with reference the same object in memory.

With the current behavior, the moment a JSON object is deserialized and typecasted to its Modelina model, it will always fail its comparison with an identical object if it contains a reference type which does not overload the "==" operator (which includes other models generated by Modelina).

This change would allow the Equals() function to be applicable in more situations.

This does not fix Equals() completely, as both the current implementation and this implementation proposed below would still fail to return the proper result in the case of Enumerable properties such as Array or List.

By instead using the [Equals(Object, Object) function](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-8.0#system-object-equals(system-object-system-object)), we can:

- Implicitly perform a ReferenceEquals() on the compared properties.
- Return true if both properties are null (which may be the case if the properties are nullable).
- Fallback on the Equals() function of the compared properties.

### 🚧 Breaking changes
Yes

 I checked and didn't find a similar issue.
 I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)
Yes, I can work on this issue.

Louis-PhilippeGentile:
## Description

- Replaces `==` with `Object.Equals()` when overriding `Equals` for C# model generation
- Updates test related to the `Equals` override
- Adds a note regarding the limitations of the `Equals` override

## Related Issue

Fixes #1850 

## Checklist


jonaslagoni:
cc @RowlandBanks @JFCote wanna review?

docs/languages/Csharp.md
```diff
@@ -61,10 +61,12 @@ Currently not supported, [let everyone know you need it](https://github.com/asyn
 
 ## Generate models with equals and GetHashCode methods
 
-To overwrite the `Equal` and `GetHashCode` methods, use the preset `CSHARP_COMMON_PRESET` and provide the options `equal: true` and `hashCode: true`
+To overwrite the `Equals` and `GetHashCode` methods, use the preset `CSHARP_COMMON_PRESET` and provide the options `equal: true` and `hashCode: true`
 
 Check out this [example for a live demonstration](../../examples/csharp-generate-equals-and-hashcode).
 
+Note that, the `Equals` method uses `Object.Equals(objA, objB)`. This means that [reference types properties](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types) that do not override their default `Equals` method will perform a reference equality check to determine equality, returning `false` if they do not reference the same object.
```
RowlandBanks, line 68: nit: `Note that, the` → `Note that the`.

docs/languages/Csharp.md
```diff
@@ -61,10 +61,12 @@ Currently not supported, [let everyone know you need it](https://github.com/asyn
...
+
 ## Generate models with auto-implemented properties
 
 To generate auto-implemented properties (the ones with with `{ get; set; }` accessors), use the preset `CSHARP_COMMON_PRESET` and provide the option `autoImplementedProperties: true`
```
RowlandBanks, line 72: I realize this wasn't written by you, but can you fix the `with with` in this line 🙂? Thanks!

RowlandBanks (changes requested):

RowlandBanks (approved):

jonaslagoni (approved):
Thanks for taking the time @Louis-PhilippeGentile ✌️

jonaslagoni:
@all-contributors please add @Louis-PhilippeGentile for code, test, docs, bug

PR #1867, opened 2024-03-07
fix: python using the wrong key type for dictionaries

jonaslagoni:
## Description
This PR fixes that the Python constrainer is using an incorrect value for the key for dictionaries.


jonaslagoni:
If you want to have a look @dimitern :v:

PR #1873, opened 2024-03-09
feat!: export and import interfaces as types when using named imports and ESM Module Syntax

### LINKED ISSUE #1793: Support verbatimModuleSyntax in TypeScript Generator

marakalwa:
## Description
When using modelType 'interface', exportType 'named' and moduleSystem 'ESM' export and import interfaces as types to support verbatimModuleSyntax.
This should not be a breaking change, interfaces are types even when not using verbatimModuleSyntax, explicitly marking them as types is optional in that case though.

## Related Issue
fixes #1793 

## Checklist

## Additional Notes
This PR also includes a change that makes sure only interfaces are imported/exported as types, as the previous version of this PR also exported enums as types.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -135,6 +135,7 @@ These are all specific examples only relevant to the TypeScript generator:
 - [typescript-generate-example](./typescript-generate-example) - A basic example of how to use Modelina and output a TypeScript class with an example function.
 - [typescript-generate-marshalling](./typescript-generate-marshalling) - A basic example of how to use the un/marshalling functionality of the typescript class.
 - [typescript-generate-comments](./typescript-generate-comments) - A basic example of how to generate TypeScript interfaces with comments from description and example fields.
+- [typescript-interfaces-as-types](./typescript-interfaces-as-types) - A basic example that generate the models as interfaces and uses ESM module system syntax to import and export them as types.
 - [typescript-use-esm](./typescript-use-esm) - A basic example that generate the models to use ESM module system.
 - [typescript-use-cjs](./typescript-use-cjs) - A basic example that generate the models to use CJS module system.
 - [typescript-generate-jsonbinpack](./typescript-generate-jsonbinpack) - A basic example showing how to generate models that include [jsonbinpack](https://github.com/sourcemeta/jsonbinpack) functionality.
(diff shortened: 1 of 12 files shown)


jonaslagoni (changes requested):
Gonna have to re-target to `next` because of the breaking changes to the dependency manager (it's exposed to users). Just need to make a quick update https://github.com/asyncapi/modelina/pull/1875 before I can do it.

Do you mind adding those changes to the migration docs: https://github.com/asyncapi/modelina/blob/next/docs/migrations/version-3-to-4.md

You should be able to merge `next` into your branch with no issues

Samridhi-98 (approved):

moritzkalwa:
> Gonna have to re-target to `next` because of the breaking changes to the dependency manager (it's exposed to users). Just need to make a quick update #1876 before I can do it.
> 
> Do you mind adding those changes to the migration docs: https://github.com/asyncapi/modelina/blob/next/docs/migrations/version-3-to-4.md
> 
> You should be able to easily merge `next` into your branch 🙂

Are you sure this needs to go into the migration guide? This should not break anything afaik, environments that haven't used verbatim module Syntax before should still work with interfaces now exported/imported as types and environments that do will now work without causing errors. 🤔 I could certainly add this into the migration guide, it feels wrong there though as there is no action required.

jonaslagoni:
> Are you sure this needs to go into the migration guide? This should not break anything afaik, environments that haven't used verbatim module Syntax before should still work with interfaces now exported/imported as types and environments that do will now work without causing errors.

Its not the new feature that is breaking, but the syntax of the dependency manager functions, which are public interface for people to use 🙂

PR #1874, opened 2024-03-10
fix: use this. for property setter in c# class renderer

### LINKED ISSUE #1862: [BUG] C# setter for property named `value` prevents assignment.
### Describe the bug.

A C# setter has a "magic" variable called `value` which contains the value being assigned.

If a model has a property called `Value`, then the generated setter is broken, as it will attempt to assign the magic variable, rather than the backing class field.

```c#
public partial class Duration
{
    private readonly decimal value;

    public decimal Value
    {
        get { return value; }
        // This setter is broken - it is assigning the setter-level value, not the class-level value.
        set { value = value; }
    }
}

### Expected behavior

The setter could be generated as `this.value = value`;

```c#
public decimal Value
{
    get { return value; }
    set { this.value = value; }
}
```

### Screenshots

N/A

### How to Reproduce

1. I opened the [online playground](https://modelina.org/playground?language=csharp).
2. I changed one of the field names to `value`.
3. I noted the C# setter was generated incorrectly, as described above.

### 🥦 Browser

Google Chrome

- [X] I checked and didn't find similar issue

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

moritzkalwa:
## Description
The setter for c# class properties should use this.{propertyName} to assign values to avoid an edge case where there is a property named 'value'

## Related Issue
Fixes #1862 

## Checklist

### DIFF
--- a/examples/csharp-change-collection-type/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-change-collection-type/__snapshots__/index.spec.ts.snap
@@ -9,7 +9,7 @@ Array [
   public IEnumerable<string>? Email 
   {
     get { return email; }
-    set { email = value; }
+    set { this.email = value; }
   }
 }",
 ]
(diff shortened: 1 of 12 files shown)


jonaslagoni (changes requested):
Remember to update the example snapshots `npm run test:examples:update`

jonaslagoni (approved):
Awesome 👍

PR #1955, opened 2024-04-23
fix: support dash in property names (Pydantic generator)

ahem:
## Description
This enables the generated models for Pydantic to work for properties with dashes in the name.

The generated code now puts descriptions in the "description" argument of the Pydantic Field decorator, and puts the original, unconstrained, property name in the "alias" field. This is correct usage as per the [Pydantic documentation](https://docs.pydantic.dev/latest/concepts/fields/#field-aliases)

Also, dashes are allowed through in property key constrainer, so that a property name like "content-type" is converted into "contentType" instead of "contentMinusType".

## Related Issue
I believe this fixes https://github.com/asyncapi/modelina/issues/1058 but apparently that has been closed with a fix to an mostly unrelated problem.

## Checklist

## Additional Notes
With these changes the generated Pydantic model class can correctly deserialise and serialise the JSON object described by the schema in the updated [generate-python-pydantic-models](https://github.com/asyncapi/modelina/blob/9610ee2e66b9de045543590bc627585b5f7955ef/examples/generate-python-pydantic-models/index.ts) test:

```json
{
   "requiredField": "test@example.com",
   "noDescription": "test",
   "options": 123,
   "content-type": "application/json"
}
```

Without this change Pydantic would ignore the `"content-type"` field since it would not know to map that to the models `contentType` property.

Note that this is technically a breaking change, since before users could use the "description" field in the JSON schema to set the "alias" field on the Pydantic model. This would be a misuse of the description field, though, since it is to decorate a user interface with information about the data produced by this user interface.

### DIFF
--- a/examples/generate-python-pydantic-models/__snapshots__/index.spec.ts.snap
+++ b/examples/generate-python-pydantic-models/__snapshots__/index.spec.ts.snap
@@ -3,10 +3,11 @@
 exports[`Should be able to render python models and should log expected output to console: class-model 1`] = `
 Array [
   "class Root(BaseModel): 
-  optionalField: Optional[str] = Field(alias='''this field is optional''', default=None)
-  requiredField: str = Field(alias='''this field is required''')
+  optionalField: Optional[str] = Field(description='''this field is optional''', default=None)
+  requiredField: str = Field(description='''this field is required''')
   noDescription: Optional[str] = Field(default=None)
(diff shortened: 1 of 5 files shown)


jonaslagoni (approved):
Awesome, thanks @ahem 🙏

jonaslagoni:
@all-contributors please add @ahem for code, test, bug

PR #1964, opened 2024-04-25
chore: make the npm script portable in Maven example

rquinio1A:
## Description

* Make the npm script portable in Maven example. npm guideline is to not prefix with relative path, as executables are added to the $PATH. Tested on Windows and macOS.
* Add missing "npm install" excution to Maven frontend plugin.

## Checklist

I don't think it's applicable, as the examples are not run in the CI ?


jonaslagoni:
@all-contributors please add @rquinio1A for code, examples

PR #2004, opened 2024-05-20
fix: add go omitempty json tag

Souvikns:
## Description

This PR is adding a new option for `GO_COMMON_PRESET` for generating `omitempty` parameter. 

## Checklist

### DIFF
--- a/examples/go-json-tags/__snapshots__/index.spec.ts.snap
+++ b/examples/go-json-tags/__snapshots__/index.spec.ts.snap
@@ -4,8 +4,8 @@ exports[`Should be able to render json-tags in struct and should log expected ou
 Array [
   "// Root represents a Root model.
 type Root struct {
-  Cities *Cities \`json:\\"cities\\"\`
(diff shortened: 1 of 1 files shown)


jonaslagoni (changes requested):
I dont think this should be having it's own option, it's required for the json tags to correctly work IMO 🤔 

Why would you like to have it as an option @Souvikns?

Souvikns:
It needs `addJsonTag` to work, I added an option thinking that if anyone does not need it, if anyone wants to have empty data while Marshaling, they can do it. Just didn't wanted to force the `omitempty` tag on the user.

jonaslagoni:
The problem as I see it with adding it as an option is that it currently (by default) creates incorrect JSON Data as soon as you have an optional property or null type. To me this is a bug rather then a missing feature 🤔

jonaslagoni (changes requested):
Got some snapshots that are not happy :v:

jonaslagoni (approved):
👍

PR #2068, opened 2024-07-11
feat!: no setters for enums in interfaces for java when allowInheritance is used

kennethaasan:
## Description

In Java, when a class implements an interface, it must implement all the methods of that interface. Because of that, Modelina now doesn't render the setter for enums in interfaces when `allowInheritance` is true because the classes that implement the interface might use a constant value.

## Checklist

### DIFF
--- a/src/generators/java/renderers/ClassRenderer.ts
+++ b/src/generators/java/renderers/ClassRenderer.ts
@@ -197,10 +197,25 @@ export const JAVA_DEFAULT_CLASS_PRESET: ClassPresetType<JavaOptions> = {
     } ${getterName}() { return this.${property.propertyName}; }`;
   },
   setter({ property, model }) {
+    const setterName = FormatHelpers.toPascalCase(property.propertyName);
+
     if (property.property.options.const?.value) {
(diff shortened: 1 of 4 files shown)


kennethaasan:
@jonaslagoni, please check. Can we instead check if the classes that implement interfaces have const set in them, or is this the only way?

jonaslagoni (approved):

src/generators/java/renderers/ClassRenderer.ts
```diff
@@ -10,6 +10,7 @@ import { FormatHelpers } from '../../../helpers';
 import { JavaOptions } from '../JavaGenerator';
 import { ClassPresetType } from '../JavaPreset';
 import { unionIncludesBuiltInTypes } from '../JavaConstrainer';
+import { isEnum } from '../../csharp/Constants';
```
jonaslagoni, line 13: Think it's time to move this to general :laughing:

kennethaasan: Yes 😄

PR #2103, opened 2024-09-24
feat!: add jackson annotations additionalProperties

memdal:
* Adds @JsonAnySetter and @JsonAnyGetter annotations to additionalProperties

## Description
Adds these annotations so that Jackson can marshall and unmarshall additional properties.

## Checklist

### DIFF
--- a/src/generators/java/presets/JacksonPreset.ts
+++ b/src/generators/java/presets/JacksonPreset.ts
@@ -74,10 +74,15 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
       if (model.options.isExtended) {
         return content;
       }
-      const isAdditionalProperties =
-        property.propertyName === 'additionalProperties';
+      //Properties that are dictionaries with unwrapped options, cannot get the annotation because it cannot be accurately unwrapped by the jackson library.
+      const isDictionary =
+        property.property instanceof ConstrainedDictionaryModel;
+      const hasUnwrappedOptions =
+        isDictionary &&
+        (property.property as ConstrainedDictionaryModel).serializationType ===
+          'unwrap';
       const blocks: string[] = [];
-      if (isAdditionalProperties) {
+      if (hasUnwrappedOptions) {
         blocks.push(renderer.renderAnnotation('JsonAnyGetter'));
       }
       blocks.push(content);
--- a/src/generators/java/presets/JacksonPreset.ts
+++ b/src/generators/java/presets/JacksonPreset.ts
@@ -35,6 +35,7 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
       const blocks: string[] = [];
 
       if (hasUnwrappedOptions) {
+        blocks.push(renderer.renderAnnotation('JsonAnySetter'));
         if (!property.required) {
           blocks.push(
             renderer.renderAnnotation(
@@ -67,6 +68,19 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
 
       blocks.push(content);
 
+      return renderer.renderBlock(blocks);
+    },
+    getter({ renderer, property, content, model }) {
+      if (model.options.isExtended) {
+        return content;
+      }
+      const isAdditionalProperties =
+        property.propertyName === 'additionalProperties';
+      const blocks: string[] = [];
+      if (isAdditionalProperties) {
+        blocks.push(renderer.renderAnnotation('JsonAnyGetter'));
+      }
+      blocks.push(content);
       return renderer.renderBlock(blocks);
     }
   },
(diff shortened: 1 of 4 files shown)


memdal:
Hi, thanks for taking a look. What would unwrapped mean in this context?

In my tests any unknown key/value pair was given its own entry in the map, and any object value would become a map of its own.

jonaslagoni:
`{additionalProperties: {test: '123'}}` should become `{test: '123'}`.

if it's not possible we should just ignore the property and add it to the documentation

memdal:
In that case it does some wrapping/unwrapping. What happens is now (pseudocode):

Json to deserialize:
```json
{
  "defined_property": "defined property value",
  "undefined_property": "undefined property value"
}
```

Java class to unmarshal the JSON into:
```java
class JavaClassRepresentation {
// getters/setters/annotations and other methods omitted for brevity
  private String definedProperty = "defined property value";
  private Map<String, Object> additionalProperties = Map.of(
    Entry.of("undefined_property", "undefined property value")
  );
}
```
    
Json serialized from the Java class:
```json
{
  "defined_property": "defined property value",
  "undefined_property": "undefined property value"
}
```

The one edge case I am uncertain about is if an additional property is actually named `additionalProperties` and has an object value. In that case it can either merge the maps and lose the wrapping/unwrapping, or it will have to lose data.

memdal:
@jonaslagoni The @JsonAnyGetter annotation ensures that the properties from the additionalProperties map are unwrapped on serialization. Do you think this change could be added?

jonaslagoni:
@memdal thats perfect yea: https://www.tutorialspoint.com/jackson_annotations/jackson_annotations_jsonanygetter.htm

My bad for not realizing it before :v:

src/generators/java/presets/JacksonPreset.ts
```diff
@@ -27,13 +27,19 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
       //Properties that are dictionaries with unwrapped options, cannot get the annotation because it cannot be accurately unwrapped by the jackson library.
       const isDictionary =
         property.property instanceof ConstrainedDictionaryModel;
+      const isAdditionalProperties =
+        property.propertyName === 'additionalProperties';
```
jonaslagoni, line 31:
```suggestion
```

src/generators/java/presets/JacksonPreset.ts
```diff
@@ -27,13 +27,19 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
...
       const hasUnwrappedOptions =
         isDictionary &&
         (property.property as ConstrainedDictionaryModel).serializationType ===
           'unwrap';
 
       const blocks: string[] = [];
 
+      if (isAdditionalProperties) {
```
jonaslagoni, line 39: You can use `hasUnwrappedOptions` instead.

```suggestion
      if (hasUnwrappedOptions) {
```

jonaslagoni (changes requested):

src/generators/java/presets/JacksonPreset.ts
memdal: removed this const

src/generators/java/presets/JacksonPreset.ts
memdal: moved the push into the existing if block for `hasUnwrappedOptions`

src/generators/java/presets/JacksonPreset.ts
```diff
@@ -67,6 +68,19 @@ export const JAVA_JACKSON_PRESET: JavaPreset = {
...
+    getter({ renderer, property, content, model }) {
+      if (model.options.isExtended) {
+        return content;
+      }
+      const isAdditionalProperties =
+        property.propertyName === 'additionalProperties';
+      const blocks: string[] = [];
+      if (isAdditionalProperties) {
+        blocks.push(renderer.renderAnnotation('JsonAnyGetter'));
+      }
```
jonaslagoni, line 82: Where did the check for constrained dictionary model with unwrap serialization type go? 🤔

jonaslagoni (changes requested):

src/generators/java/presets/JacksonPreset.ts
memdal: I forgot to update this block of code as well. I am guessing I should replace this:
```ts
      const isAdditionalProperties =
        property.propertyName === 'additionalProperties';
      const blocks: string[] = [];
      if (isAdditionalProperties) {
```

with this:
```
      //Properties that are dictionaries with unwrapped options, cannot get the annotation because it cannot be accurately unwrapped by the jackson library.
      const isDictionary =
        property.property instanceof ConstrainedDictionaryModel;
      const hasUnwrappedOptions =
        isDictionary &&
        (property.property as ConstrainedDictionaryModel).serializationType ===
          'unwrap';
      const blocks: string[] = [];
      if (hasUnwrappedOptions) {
```

jonaslagoni: Yep, looks right

jonaslagoni (approved):
Nice 👍

jonaslagoni:
@all-contributors please add @memdal for code, test

PR #2111, opened 2024-10-23
fix: python pydantic unwrap properties

memdal:
## Description
1. Pydantic code generation now checks if data is a dict or an object before getting keys in the unwrap functions.
2. Alias has been removed for unconstrained properties so that Pydantic objects can follow idiomatic casing for the additional_properties property:
```python
MyPydanticObject(
    defined_property=1,
    additional_properties={
        "undefined_property": "2"
    }
)
```

## Checklist

### DIFF
--- a/src/generators/python/presets/Pydantic.ts
+++ b/src/generators/python/presets/Pydantic.ts
@@ -57,9 +57,6 @@ const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
     if (!property.required) {
       decoratorArgs.push('default=None');
     }
-    if (property.propertyName !== property.unconstrainedPropertyName) {
-      decoratorArgs.push(`alias='''${property.unconstrainedPropertyName}'''`);
-    }
 
     return `${propertyName}: ${type} = Field(${decoratorArgs.join(', ')})`;
   },
@@ -101,6 +98,8 @@ def custom_serializer(self, handler):
 @model_validator(mode='before')
 @classmethod
 def unwrap_${dictionaryModel?.propertyName}(cls, data):
+  if not isinstance(data, dict):
+    data = data.model_dump()
   json_properties = list(data.keys())
   known_object_properties = [${allProperties
     .map((value) => `'${value}'`)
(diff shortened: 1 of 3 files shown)


src/generators/python/presets/Pydantic.ts
```diff
@@ -101,7 +101,7 @@ def custom_serializer(self, handler):
 @model_validator(mode='before')
 @classmethod
 def unwrap_${dictionaryModel?.propertyName}(cls, data):
-  json_properties = list(data.keys())
+  json_properties = list(data.model_dump().keys())
```
maxplatov, line 104: [Before model validators are passed the raw input which is often a dict[str, Any] but could also be an instance of the model itself (e.g. if UserModel.model_validate(UserModel.construct(...)) is called) ](https://docs.pydantic.dev/2.9/concepts/validators/#model-validators)

your code have to support dict(how it was before) and pydantic model

memdal: I've replaced this with a function that I believe will work for both the model and a dict now. Can you check if this looks useable?

maxplatov: I guess you need to change
`list_keys(data) `to `cls.list_keys(data)`

memdal: Updated it to be `cls.list_keys(data)` now

src/generators/python/presets/Pydantic.ts
```diff
@@ -114,7 +117,7 @@ def unwrap_${dictionaryModel?.propertyName}(cls, data):
     .map((value) => `'${value.unconstrainedPropertyName}'`)
     .join(', ')}]
   ${dictionaryModel?.propertyName} = {}
-  for obj_key in list(data.keys()):
+  for obj_key in cls.list_keys(data):
```
maxplatov, line 120: I still believe that this code will never called due to condition above
`if len(unknown_object_properties) == 0:`

jonaslagoni (commented):
@maxplatov want to have another look?

jonaslagoni (approved):

jonaslagoni:
@all-contributors please add @memdal  for code, test

jonaslagoni:
@all-contributors please add @maxplatov for review

PR #2131, opened 2024-11-29
feat: add support for typed array unmarshalling in TypeScript

jonjomckay:
## Description

This PR adds support for properly unmarshalling array types, when using the TypeScript generator. Previously, the JSON array was being passed straight into the resulting object, which causes issues when using the value in another schema object and marshalling that object. The other object expects the array to contain full objects with a callable `marshal` function, not a plain object, giving an error like:

```
err: {
      "type": "TypeError",
      "message": "unionItem.marshal is not a function",
      "stack":
          TypeError: unionItem.marshal is not a function
              at KeywordSearchRequest.marshal
...
```
### Example

Given a type like this:

```yaml
arrayProperty:
  type: array
  items:
    $ref: "#/definitions/SomeObjectType"
```

The generator will now generate unmarshalling code like this:

```typescript
...
instance.arrayProperty = obj["arrayProperty"] == null
  ? null
  : obj["arrayProperty"].map((item: any) => SomeObjectType.unmarshal(item));
```

## Checklist
- [] All tests pass successfully locally.(`npm run test`).

## Additional Notes

The tests don't seem to pass even before my changes. I have updated the existing snapshot to cater for this change, and that test runs successfully.

### DIFF
--- a/src/generators/typescript/presets/utils/UnmarshalFunction.ts
+++ b/src/generators/typescript/presets/utils/UnmarshalFunction.ts
@@ -1,12 +1,14 @@
 import { ClassRenderer } from '../../renderers/ClassRenderer';
 import { getDictionary, getNormalProperties } from '../../../../helpers';
 import {
+  ConstrainedArrayModel,
   ConstrainedDictionaryModel,
   ConstrainedEnumModel,
   ConstrainedMetaModel,
   ConstrainedObjectModel,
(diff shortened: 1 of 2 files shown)


jonaslagoni (approved):

jonaslagoni:
@all-contributors please add @jonjomckay for code, bug, test

Issue #2190, opened 2025-02-21
[BUG] TypeScript generator for Date property type


roman-supy-io:
### Describe the bug.

The Typescript generator does not respect the
```json
{ "type": "string", "format": "date-time" }
```

The output is expected to be
```ts
deliveryDate: Date;
```
Instead it is generated as:
```ts
deliveryDate: string;
```

The issue looks similar to [https://github.com/asyncapi/modelina/issues/1607](This Issue) but is Typescript-related

### Expected behavior

The `Date` type should generated and not a `string`

### Screenshots

![Image](https://github.com/user-attachments/assets/e2830a17-7194-4b4d-98d7-a181dfbd71c0)

![Image](https://github.com/user-attachments/assets/1e0d8a1d-da02-4491-83a9-85080624ec69)

### How to Reproduce

1. Use a schema with a property defined as { "type": "string", "format": "date-time" }
2. run `npx modelina generate typescript ...`
3. inspect the generated output's class/interface

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

jonaslagoni:
Should be pretty easy fix, here is how we do it for Java: https://github.com/asyncapi/modelina/blob/b38dd6c771753b42eefba94b37ed313be4e4e164/src/generators/java/JavaConstrainer.ts#L172

Should be applied here: https://github.com/asyncapi/modelina/blob/master/src/generators/typescript/TypeScriptConstrainer.ts#L34

jonaslagoni:
@roman-supy-io if you make the change and provide a PR it can be released right after 🙂

kunal-595:
Hi, I’d like to work on this issue if it’s available.
Could you please assign it to me?

AarishMansur:
Hey @roman-supy-io  I have pushed my PR 
lemme know if some changes are needed 
Thanks !!

PR #2192, opened 2025-02-21
fix(typescript-generator)!: handle 'date' and 'date-time' for 'string'…

roman-supy-io:
… data type

## Description
Date type was not properly handled and resulted into string generated type output
This fixes the issue

## Related Issue
[](https://github.com/asyncapi/modelina/issues/2190)

## Checklist

### DIFF
--- a/modelina-cli/README.md
+++ b/modelina-cli/README.md
@@ -162,7 +162,7 @@ $ npm install -g @asyncapi/modelina-cli
 $ modelina COMMAND
 running command...
 $ modelina (--version)
-@asyncapi/modelina-cli/4.0.3 linux-x64 node-v18.20.6
+@asyncapi/modelina-cli/4.0.4-next.1 linux-x64 node-v18.20.7
 $ modelina --help [COMMAND]
 USAGE
   $ modelina COMMAND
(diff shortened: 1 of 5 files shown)


jonaslagoni:
@roman-supy-io remember to update all the test etc `npm run test:update`

jonaslagoni (changes requested):
Besides updating snapshots can you add two test to https://github.com/asyncapi/modelina/blob/039dab0580e740005692f6b2770f364ed501e1e9/test/generators/typescript/TypeScriptConstrainer.spec.ts#L82 ?

You can use the Java tests as guideline: https://github.com/asyncapi/modelina/blob/master/test/generators/java/JavaConstrainer.spec.ts#L329

jonaslagoni:
@roman-supy-io did you run `npm run test:library -- -u` or `npm run test:update`? Looks like there is still some snapshots that has not been updated 🙂

jonaslagoni:
Regardless, because this is a breaking change, I dont think any automatic updates will happen.

jonaslagoni:
@roman-supy-io you can read more about it here: https://github.com/asyncapi/modelina?tab=readme-ov-file#versioning-and-maintenance

Alternatively you can add a flag specifying an enum of types for date formats :thinking: i.e. `string` (default) and `date` so the output does not automatically break for the consumers of the models.

Issue #2076, opened 2024-07-23
[BUG] from . import AnonymousSchemaX breaks types (?)


black-snow:
### Describe the bug.

(ref. https://github.com/asyncapi/cli/issues/1485)

So with `--pyDantic`, the CLI generates models like:

```py
from . import AnonymousSchema15

class AnonymousSchema13(BaseModel): 
  command: Optional[str] = Field(default=None)
  meta: Optional[AnonymousSchema15.AnonymousSchema15] = Field(default=None)
```

with 

```py
class AnonymousSchema15(BaseModel): 
  whatever: str = Field()
```

But when I try to use the models I end up with the unhelpful message:
> ...
> /opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/typing.py:415: in _eval_type
    return t._evaluate(globalns, localns, type_params, recursive_guard=recursive_guard)
/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/typing.py:938: in _evaluate
    eval(self.__forward_code__, globalns, locals_to_pass),
<string>:1: in <module>
    ???
../.venv/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py:237: in __getattr__
    raise AttributeError(item)
E   AttributeError: AnonymousSchema15

The import *seems* to break the types for pydantic. It works fine when I go with `from .AnonymousSchema15 import AnonymousSchema15` and then `Optional[AnonymousSchema15]`.
This seems to be some deep crappy typehinting black magic eff-up.

So the fix might just be to adjust the import + the usage.

### Expected behavior

Nested models should work out of the box.

### Screenshots

-

### How to Reproduce

* Have a message with `allOf`:
```yaml
      payload:
        allOf:
          - $ref: "#/components/schemas/Base"
          - $ref: "#/components/schemas/SomethingOnTop"
```
2. `asyncapi generate models python asyncapi.yaml --pyDantic`
3. Try to use the generated model(s)

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

black-snow:
[...]

PR #2201, opened 2025-03-06
fix: import breaks types (?) #2076

Prthmsh7:
## Description
I tried fixing this issue #2076 by correcting how nested anonymous schemas are imported when using the --pyDantic flag. Previously, the generator produced code like: 
```python
from . import AnonymousSchema15

class AnonymousSchema13(BaseModel): 
  command: Optional[str] = Field(default=None)
  meta: Optional[AnonymousSchema15.AnonymousSchema15] = Field(default=None)
```
which led to runtime type errors. With this fix, the import is now explicit:

```python
from .AnonymousSchema15 import AnonymousSchema15

class AnonymousSchema13(BaseModel): 
  command: Optional[str] = Field(default=None)
  meta: Optional[AnonymousSchema15] = Field(default=None)
```

This change ensures that Pydantic can properly evaluate the type hints.

## Related Issue
(ref. https://github.com/asyncapi/cli/issues/1485)

## Checklist
- [✅] The code follows the project's coding standards and is properly linted (`npm run lint`).
- [✅] Tests have been added or updated to cover the changes.
- [✅] Documentation has been updated to reflect the changes.
- [✅] All tests pass successfully locally.(`npm run test`).
bug-solved

## Additional Notes
**This is my first time working with the tests and on such a large scale, so please review the changes thoroughly**


jonaslagoni (changes requested):
Please revert your package changes as this fix should not require it 🙂

PR #2233, opened 2025-04-16
fix(csharp-json-generator): remove CanConvert override for Json converters

rmaced0:
## Description
I ran into errors where the generated class's custom JsonConverter implementation was applying to objects that it should not be applying to. It applied the same JsonConverter to the properties of the object. For example, lets say we have a class called `Robot` and two properties (`name: string`, and `id: System.Guid`). If `CanConvert` always returns `true`, then when parsing a json string to the `Robot` class, it will use the `RobotConverter` on the `System.Guid` causing it to fail.

According to [Microsoft Documentation](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to):
```
Override the CanConvert method only if necessary.
The default implementation returns true when the type to convert is of type T.
Therefore, converters that support only type T don't need to override this method.
```

## Checklist

### DIFF
--- a/examples/csharp-generate-json-serializer/__snapshots__/index.spec.ts.snap
+++ b/examples/csharp-generate-json-serializer/__snapshots__/index.spec.ts.snap
@@ -7,7 +7,7 @@ public partial class Root
 {
   private string? email;
 
-  public string? Email 
+  public string? Email
   {
     get { return email; }
     set { this.email = value; }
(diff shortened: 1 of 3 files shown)


jonaslagoni (approved):
Thanks @rmaced0 :v:

jonaslagoni:
@all-contributors please add @rmaced0  for code, test, bug

Issue #2227, opened 2025-04-10
[BUG] Missing Common Properties in Generated Interface for Union Types in Java Generator


ThomasVerhoeven1998:
### Describe the bug.

Let's assume the following example using Modelina 4.2.0 with the java generator:

```
asyncapi: 3.0.0
info:
  title: Event Bus
  version: 1.0.0
servers:
  pulsar:
    host: 'localhost:6650'
    protocol: pulsar
channels:
  workers-created:
    address: workers-created
    messages:
      WorkersCreatedPublicEvent:
        $ref: '#/components/messages/WorkersCreatedPublicEvent'
operations:
  workers-created.subscribe:
    action: send
    channel:
      $ref: '#/channels/workers-created'
    messages:
      - $ref: '#/channels/workers-created/messages/WorkersCreatedPublicEvent'
components:
  schemas:
    WorkersCreatedEvent:
      title: WorkersCreatedEvent
      type: object
      properties:
        id:
          type: string
          format: uuid
        workers:
          type: array
          items:
            $ref: '#/components/schemas/Worker'
    Worker:
      type: object
      properties:
        name:
          type: string
      discriminator: "@type"
      required:
        - name
      oneOf:
        - $ref: '#/components/schemas/Employer'
        - $ref: '#/components/schemas/Employee'
    Employer:
      title: Employer
      type: object
      properties:
        hasFullAccess:
          type: boolean
      required:
        - hasFullAccess
    Employee:
      title: Employee
      type: object
      properties:
        registrationDate:
          type: string
      required:
        - registrationDate
  messages:
    WorkersCreatedPublicEvent:
      title: WorkersCreatedPublicEvent
      name: WorkersCreatedPublicEvent
      contentType: application/schema+json;version=draft-07
      payload:
        schemaFormat: application/vnd.aai.asyncapi+json;version=3.0.0
        schema:
          $ref: "#/components/schemas/WorkersCreatedEvent"
```

This will generate the following code: 

```
public class WorkersCreatedEvent {
  @JsonProperty("id")
  @JsonInclude(JsonInclude.Include.NON_NULL)
  private String id;
  @JsonProperty("workers")
  @JsonInclude(JsonInclude.Include.NON_NULL)
  private List<Worker> workers;

  public String getId() { return this.id; }
  public void setId(String id) { this.id = id; }

  public List<Worker> getWorkers() { return this.workers; }
  public void setWorkers(List<Worker> workers) { this.workers = workers; }
}
```

```
[...]

PR #2247, opened 2025-04-24
feat: support common properties in unions for java generator

ThomasVerhoeven1998:
## Description
feat: support common properties in unions for java generator (fix for issue https://github.com/asyncapi/modelina/issues/2227)
fix: records implementing union interface now compile
fix: remove infinite loop when allowInheritance: true in combination with oneOf and allOf
fix: do not generate setter for discriminator fields
fix: apply `@override` when method is from parent or extend correctly

## Related Issue
https://github.com/asyncapi/modelina/issues/2227

## Checklist

### DIFF
--- a/docs/internal-model.md
+++ b/docs/internal-model.md
@@ -62,6 +62,7 @@ classDiagram
     class UnionModel{
         <<class>>
         MetaModel[] unionModels
+        &lt;String, ObjectPropertyModel&gt; properties
     }
 
     class EnumValueModel{
@@ -180,7 +181,7 @@ classDiagram
 
     class ConstrainedObjectModel{
         <<class>>
-        &lt;String, ObjectPropertyModel&gt; properties
+        &lt;String, ConstrainedObjectPropertyModel&gt; properties
     }
 
     class ConstrainedObjectPropertyModel{
@@ -198,6 +199,7 @@ classDiagram
     class ConstrainedUnionModel{
         <<class>>
         ConstrainedMetaModel[] unionModels
+        &lt;String, ConstrainedObjectPropertyModel&gt; properties
     }
 
     class ConstrainedEnumValueModel{
--- a/src/generators/cplusplus/constrainer/PropertyKeyConstrainer.ts
+++ b/src/generators/cplusplus/constrainer/PropertyKeyConstrainer.ts
@@ -1,4 +1,9 @@
-import { ConstrainedObjectModel, ObjectModel } from '../../../models';
+import {
+  ConstrainedObjectModel,
+  ConstrainedUnionModel,
+  ObjectModel,
+  UnionModel
+} from '../../../models';
 import {
   NO_NUMBER_START_CHAR,
   NO_DUPLICATE_PROPERTIES,
@@ -12,8 +17,8 @@ export type PropertyKeyConstraintOptions = {
   NO_SPECIAL_CHAR: (value: string) => string;
   NO_NUMBER_START_CHAR: (value: string) => string;
   NO_DUPLICATE_PROPERTIES: (
-    constrainedObjectModel: ConstrainedObjectModel,
-    objectModel: ObjectModel,
+    constrainedObjectModel: ConstrainedObjectModel | ConstrainedUnionModel,
+    objectModel: ObjectModel | UnionModel,
     propertyName: string,
     namingFormatter: (value: string) => string
   ) => string;
--- a/src/generators/csharp/constrainer/PropertyKeyConstrainer.ts
+++ b/src/generators/csharp/constrainer/PropertyKeyConstrainer.ts
@@ -1,4 +1,9 @@
-import { ConstrainedObjectModel, ObjectModel } from '../../../models';
+import {
+  ConstrainedObjectModel,
+  ConstrainedUnionModel,
+  ObjectModel,
+  UnionModel
+} from '../../../models';
 import {
   NO_NUMBER_START_CHAR,
   NO_DUPLICATE_PROPERTIES,
@@ -13,8 +18,8 @@ export type PropertyKeyConstraintOptions = {
   NO_SPECIAL_CHAR: (value: string) => string;
   NO_NUMBER_START_CHAR: (value: string) => string;
   NO_DUPLICATE_PROPERTIES: (
-    constrainedObjectModel: ConstrainedObjectModel,
-    objectModel: ObjectModel,
+    constrainedObjectModel: ConstrainedObjectModel | ConstrainedUnionModel,
+    objectModel: ObjectModel | UnionModel,
     propertyName: string,
     namingFormatter: (value: string) => string
   ) => string;
@@ -22,7 +27,7 @@ export type PropertyKeyConstraintOptions = {
   NAMING_FORMATTER: (value: string) => string;
   NO_RESERVED_KEYWORDS: (value: string) => string;
   NO_ENCLOSING_NAMES: (
-    constrainedObjectModel: ConstrainedObjectModel,
+    constrainedObjectModel: ConstrainedObjectModel | ConstrainedUnionModel,
     value: string
   ) => string;
 };
--- a/src/generators/dart/constrainer/PropertyKeyConstrainer.ts
+++ b/src/generators/dart/constrainer/PropertyKeyConstrainer.ts
@@ -1,4 +1,9 @@
-import { ConstrainedObjectModel, ObjectModel } from '../../../models';
+import {
+  ConstrainedObjectModel,
+  ConstrainedUnionModel,
+  ObjectModel,
+  UnionModel
+} from '../../../models';
 import {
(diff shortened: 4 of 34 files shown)


ThomasVerhoeven1998:
Some runtime tests seem to fail, but it has nothing to do with our changes. Any advice on how to fix?

jonaslagoni:
Gonna review it tonight when I have more time :)

docs/internal-model.md
```diff
@@ -62,6 +62,7 @@ classDiagram
     class UnionModel{
         <<class>>
         MetaModel[] unionModels
+        &lt;String, ObjectPropertyModel&gt; properties
```
jonaslagoni, line 65: Why should a union model all of the sudden have properties?

jonaslagoni: Ahh, you want to represent the inheritance through union model...

I dont think this is the best idea from an abstracted point of view, Do you? In most languages union just means that it can be one or more types.

Not really an inheritance ObjectModel. Would it not make more sense to include a `inheritFrom: ObjectModel` property in the ObjectModel or something similar to represent the inheritance?

ThomasVerhoeven1998: Subclasses implement the union which has common properties indeed. 

Lets say you have a pet with a name and color. A Pet can be a dog or a cat. The union should have name and color, than the override is correct. Now the union is pretty useless

ThomasVerhoeven1998: See the last example in the following issue https://github.com/asyncapi/spec/issues/1098

ThomasVerhoeven1998: If you say it can't be properties, what is the usage of union than?

ThomasVerhoeven1998: Let's say we have the following schema: 

All workers have a name. `Employer` and `employee` are `workers`. Employer also has a field `hasFullAccess` and Employee has a `registrationTimestamp`

```
components:
  schemas:
    WorkersCreatedEvent:
      title: WorkersCreatedEvent
      type: object
      properties:
        id:
          type: string
          format: uuid
        workers:
          type: array
          items:
            $ref: '#/components/schemas/Worker'
    Worker:
      type: object
      properties:
        "@type":
          type: string
        name:
          type: string
      discriminator: "@type"
      required:
        - "@type"
        - name
      oneOf:
        - $ref: '#/components/schemas/Employer'
        - $ref: '#/components/schemas/Employee'
    Employer:
      title: Employer
      type: object
      properties:
        hasFullAccess:
          type: boolean
      allOf:
        - $ref: '#/components/schemas/Worker'
      required:
        - hasFullAccess
    Employee:
      title: Employee
      type: object
      properties:
        registrationTimestamp:
          type: string
          format: date-time
      allOf:
        - $ref: '#/components/schemas/Worker'
      required:
        - registrationTimestamp
  messages:
    WorkersCreatedPublicEvent:
      title: WorkersCreatedPublicEvent
      name: WorkersCreatedPublicEvent
      contentType: application/schema+json;version=draft-07
      payload:
        schemaFormat: application/vnd.aai.asyncapi+json;version=3.0.0
        schema:
          $ref: "#/components/schemas/WorkersCreatedEvent"
```

Before this change modelina would have generated: 

```
@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.EXISTING_PROPERTY, property="@type", visible=true)
@JsonSubTypes({
  @JsonSubTypes.Type(value = Employer.class, name = "Employer"),
  @JsonSubTypes.Type(value = Employee.class, name = "Employee")
})
/**
 * Worker represents a union of types: Employer, Employee
 */
public interface Worker {
  String getAtType();
}

public class Employer implements Worker {
  @NotNull
  @JsonProperty("@type")
  private String atType;
  @NotNull
  @JsonProperty("name")
  private name;
  @NotNull
  @JsonProperty("hasFullAccess")
  private boolean hasFullAccess;
  @JsonAnySetter
  @JsonInclude(JsonInclude.Include.NON_NULL)

  @Override
  public java.lang.String getAtType() { return this.atType; }

  public String getName() { return this.name; }
  public void setName(String name) { this.name = name; }

  public boolean getHasFullAccess() { return this.hasFullAccess; }
  public void setHasFullAccess(boolean hasFullAccess) { this.hasFullAccess = hasFullAccess; }

}

```

That's incorrect cause Worker has common field name. So we would expect a getter and setter for name. 

After the change the code generates

```
@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.EXISTING_PROPERTY, property="@type", visible=true)
@JsonSubTypes({
  @JsonSubTypes.Type(value = Employer.class, name = "Employer"),
  @JsonSubTypes.Type(value = Employee.class, name = "Employee")
})
/**
 * Worker represents a union of types: Employer, Employee
 */
public interface Worker {
  String getAtType();
  String getName();
  void setName(String name);
}


public class Employer implements Worker {
  @NotNull
  @JsonProperty("@type")
  private java.lang.String atType;
  @NotNull
  @JsonProperty("name")
  private java.lang.String name;
  @NotNull
  @JsonProperty("hasFullAccess")
  private boolean hasFullAccess;
  @JsonAnySetter
  @JsonInclude(JsonInclude.Include.NON_NULL)

  @Override
  public java.lang.String getAtType() { return this.atType; }

  @Override
  public java.lang.String getName() { return this.name; }
  @Override
  public void setName(java.lang.String name) { this.name = name; }

  public boolean getHasFullAccess() { return this.hasFullAccess; }
  public void setHasFullAccess(boolean hasFullAccess) { this.hasFullAccess = hasFullAccess; }
  
  }
```

Now the override is correct on the name. If we don't do this and we would not be able to retrieve all names of the workers

```
        WorkersCreatedEvent workersCreatedEvent = new WorkersCreatedEvent();
        Employee employee = new Employee();
        employee.setName("John Doe");
        employee.setRegistrationTimestamp(java.time.Instant.now());
        Employer employer = new Employer();
        employer.setName("Jane Doe");
        employer.setHasFullAccess(true);
        workersCreatedEvent.setWorkers(List.of(employee, employer));
        
        //Retrieve all names
        
        List<String> names = workersCreatedEvent.getWorkers().stream()
                .map(Worker::getName)
                .toList();
```

The above code will never work without our changes. There will be no possibility to retrieve the names in easy way

ThomasVerhoeven1998: Can you explain your idea to implement the inheritance on the object model? We do not really understand how that would work for the provided example. 

And yess we know the meta model is for all languages, but we made sure it was optional to use and non-breaking

jonaslagoni: I am just gonna simplify it here, but might need a slightly different approach, but if ObjectModel had say `parent: ObjectModel` and `child: ObjectModel` then during the [interpretation phase](https://github.com/asyncapi/modelina/blob/master/src/helpers/CommonModelToMetaModel.ts#L209) (where we go from validation rules JSON Schema, to structure) you can pattern check for whether to create a union or Inheritance with ObjectModel.

Then during rendering phase in the generators, you can check whether you have a parent or child and render interface or class accordingly.

Does that make sense?

ThomasVerhoeven1998: Sounds good to me, looking forward to your PR. If we can assist somehow, please let us know. Will cherry pick our other fixes once your changes are available to see if they are still relevant. We want to use Modelina in production but the inheritance is blocking us from starting to use it, do you have an ETA?

jonaslagoni: I can't tell, if anything within a year maybe. Its not a top priority for me at all, but at some point it's certainly getting on the roadmap. 

If you want it within Modelina ASAP, I would suggest to focus on one part at a time... Happy to help ping/pong on a suitable implementation, just cant put in the development time at the moment.

Also the suggested solution is not absolute, however this current approach does not seem long term suitable unfortunate...

ThomasVerhoeven1998: Should we move our discussion to an issue and close this PR? We are also open to have a chat via Zoom or Teams,... to have a discussion about it and see if we can implement it

jonaslagoni: Yes, please start start with the issue (separate one for each feature/change) so everyone can follow along.

Also happy to setup a call if we get stuck or its needed. Feel free to ping me on slack 🙂

ThomasVerhoeven1998: Will create an issue, but before creating one. We have a few additional questions. 

In case we want to poc `parent: ObjectModel` and `child: ObjectModel`, is this an extension on `ObjectModel` cause it is also shared code between all generators. 

Also what will happen when the parent also has a discriminator + properties, will you create a union and the "new" interface, or should we allow the JsonTypeMapping also on the new kind of interface to avoid 2 interfaces (one union and one for inheritance)

ThomasVerhoeven1998: We created a detail issue to follow up further https://github.com/asyncapi/modelina/issues/2255

PR #2248, opened 2025-04-24
fix: only import List if a list is present for Java

Sm0keySa1m0n:
## Description
Currently `java.util.List` is always imported if `collectionType` is set to `List`, even if no array properties exist in the model. This PR adds a check to ensure the model contains an array property, in addition to having `collectionType` set to `List`.


jonaslagoni (changes requested):
Remember to run the linting 🙂

jonaslagoni (approved):

PR #2249, opened 2025-04-24
Remove JavaDependencyManager

Sm0keySa1m0n:
## Description
In Java, import statements are only necessary if the type you are using resides outside of the current package. As there is no way to generate models in different packages in modelina (as far as I'm aware), adding import statements for dependency models is unnecessary which therefore renders `JavaDependencyManager` unnecessary.

This is a breaking change - not sure which branch I should be targeting.


jonaslagoni:
This dependency manager can be used as options and provide a way to easily render dependencies in other files as well. 

Ontop of this going forward it makes sense to me to have an explicit dependency manager for Java 🙂

Sm0keySa1m0n:
That's why I left `AbstractDependencyManager` - it still allows you to render custom dependencies and override it via options.

The functionality provided by `JavaDependencyManager` is redundant though as it adds imports for classes that exist in the same package.

PR #2251, opened 2025-04-29
fix: classes implementing union should not have discriminator setter

ThomasVerhoeven1998:
Supported was added in https://github.com/asyncapi/modelina/releases/tag/v4.3.0 to add a setter for the discriminator on the union interface. But after reviewing this carefully, we should not generate setters at all for the discriminator, not on the class and not the union

## Checklist

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1148,6 +1148,16 @@
         "code",
         "test"
       ]
+    },
+    {
+      "login": "xeho91",
+      "name": "Mateusz Kadlubowski",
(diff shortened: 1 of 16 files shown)


jonaslagoni (changes requested):
Just need to run the linter 👍

jonaslagoni:
Changed to not breaking change as this fixed and unintended behavior and the other version only recently introduced it.

jonaslagoni:
Ahh damn, ohh well 🙆‍♂️

PR #2297, opened 2025-06-18
fix: convert dynamic to correct type before switch check for CSharp

rmaced0:
## Description
When deserializing enums from dynamic, the dynamic value is a JsonElemet and should be a different type (string, int, bool etc).
I saw the following errors when trying to deserialize:
```
 Error Message:
         Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : Cannot convert null to 'com.mycompany.app.json_serializer.HousingType' because it is a non-nullable value type
```

We need to check the type of the dynamic before proceeding to the switch case or else we will always see null.

I added a test case that fails without the new changes and passes with the new changes.

## Checklist

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1146,7 +1146,9 @@
       "profile": "https://github.com/ThomasVerhoeven1998",
       "contributions": [
         "code",
-        "test"
+        "test",
+        "bug",
+        "doc"
       ]
     },
     {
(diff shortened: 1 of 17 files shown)


jonaslagoni:
@rmaced0 can you update your branch with master? I cant seem to do it 🤨

rmaced0:
@jonaslagoni I just merged main and enabled edits by maintainers 👍

PR #2302, opened 2025-07-01
fix: add aliasing back in for Pydantic generator

### LINKED ISSUE #2187: [BUG] Pydantic, removal of alias breaks code
### Describe the bug.

Following up on https://github.com/asyncapi/modelina/issues/2174 because the cause seems to be the same changes - especially https://github.com/asyncapi/modelina/commit/e9e46b38b0b2173a4a98ec184b9149ce3be5c0ae#diff-e4313dba292240670e96b378fd98fa02d9dd6a0e7b96153c03787ee71904339fL52

Comparing the freshly generated models with my existing ones, I see that
```py
queue_name: str = Field(description='''The name of the targeted queue''', alias='''queueName''')
```

becomes
```py
queue_name: str = Field(description='''The name of the targeted queue''')
```

with
```yaml
            queueName:
              type: string
              minLength: 1
              description: "The name of the targeted queue
```

So the alias got removed. Sadly, this ain't no no-op change but a change in behaviour. Many tests fail now, especially when I `model_validate_json(some_json)`.

[This](https://github.com/asyncapi/modelina/pull/2111) seems to be the corresponding PR. Sadly, the [4.0.0 release](https://github.com/asyncapi/modelina/releases/tag/v4.0.0) just says "breaking changes!" but doesn't list them. Overall, this is a bad experience - especially since I don't use modelina directly but asyncAPI cli, which just upgraded to 4.x in a non-major update. So I bump the CLI expecting things to just work but stuff keeps blowing up :/

### Expected behavior

* `alias` should not have been removed (I don't understand the reason in the PR) OR
* 4.0.0 release should list this change (and ideally what do do about it) and then
* * asyncAPI cli should not just minor or patch change customer-facing breaking changes

### Screenshots

-

### How to Reproduce

1. use "@asyncapi/cli": "2.14.1"
2. generate model from a schema with some camelCase property
3. note the alias
4. use "@asyncapi/cli": "2.16.0" or later
5. generate model from a schema with some camelCase property
6. note the alias is missing
7. try to `model_validate_json()` from a string with said camelCase property

### 🥦 Browser

Mozilla Firefox

black-snow:
## Description
#2111 removed aliases, which was a breaking (and perhaps unintended) change that broke the API and thus downstream consumers. This change adds auto aliasing back in.

## Related Issue
fixes #2187
partially reverts PR #2111

## Checklist
- [ ] All tests pass successfully locally.(`npm run test`).

## Additional Notes
tests on main are red atm; tests for the generator are green (existing + new)

### DIFF
--- a/src/generators/python/presets/Pydantic.ts
+++ b/src/generators/python/presets/Pydantic.ts
@@ -63,7 +63,10 @@ const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
     ) {
       decoratorArgs.push('exclude=True');
     }
-    if (property.propertyName !== property.unconstrainedPropertyName && propertyName !== "additional_properties") {
+    if (
+      property.propertyName !== property.unconstrainedPropertyName &&
+      propertyName !== 'additional_properties'
+    ) {
       decoratorArgs.push(`alias='''${property.unconstrainedPropertyName}'''`);
     }
 
--- a/src/generators/python/presets/Pydantic.ts
+++ b/src/generators/python/presets/Pydantic.ts
@@ -64,9 +64,8 @@ const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
       decoratorArgs.push('exclude=True');
     }
     if (
-      property.propertyName !== property.unconstrainedPropertyName &&
-      property.property instanceof ConstrainedDictionaryModel &&
-      property.property.serializationType !== 'unwrap'
+  property.propertyName !== property.unconstrainedPropertyName &&
+  (!(property.property instanceof ConstrainedDictionaryModel) || property.property.serializationType !== 'unwrap')
     ) {
       decoratorArgs.push(`alias='''${property.unconstrainedPropertyName}'''`);
     }
--- a/src/generators/python/presets/Pydantic.ts
+++ b/src/generators/python/presets/Pydantic.ts
@@ -64,8 +64,9 @@ const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
       decoratorArgs.push('exclude=True');
     }
     if (
-  property.propertyName !== property.unconstrainedPropertyName &&
-  (!(property.property instanceof ConstrainedDictionaryModel) || property.property.serializationType !== 'unwrap')
+      property.propertyName !== property.unconstrainedPropertyName &&
+      (!(property.property instanceof ConstrainedDictionaryModel) ||
+        property.property.serializationType !== 'unwrap')
     ) {
       decoratorArgs.push(`alias='''${property.unconstrainedPropertyName}'''`);
     }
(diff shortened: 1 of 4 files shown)


black-snow:
I still have no idea what I'm doing here - I just hope this is the minimal change required. Linting doesn't work (ESLint couldn't find the config "oclif" to extend from) and tests on mainline are red so I might've accidentally broken something else.

jonaslagoni:
@black-snow run `npm run test:examples:update` 🙂

black-snow:
@jonaslagoni thanks :)

> @black-snow run `npm run test:examples:update` 🙂

Done, no changes

/edit: wait a sec, wrong branch
/edit2: still no changes, all green

/edit3: ran `npm run test:library -- -u` and it reverted the changes made to the snapshots 🤔

jonaslagoni:
> /edit: wait a sec, wrong branch /edit2: still no changes, all green
> 
> /edit3: ran `npm run test:library -- -u` and it reverted the changes made to the snapshots 🤔

@black-snow the snapshots definitely need to be updated for the examples: https://github.com/asyncapi/modelina/actions/runs/16024362105/job/45208571506?pr=2302 🙂 

`npm run test:examples:update` should handle that

black-snow:
`npm run test:examples:update` doesn't change anything. But I see that it only runs `npm run test:examples:regular && npm run test:examples:websites` and not the lib.

But more importantly, it seems like the suggested changes don't have the effect we though they'd have (given that they roll back my manual adjustments to the snapshots). Any ideas @jonaslagoni ?

jonaslagoni:
> `npm run test:examples:update` doesn't change anything. But I see that it only runs `npm run test:examples:regular && npm run test:examples:websites` and not the lib.

`npm run test:examples:update` runs https://github.com/asyncapi/modelina/blob/e78295da6eae8eead14b956fa5a77e180d758cd4/package.json#L89 not what you pasted 🙂 

> But more importantly, it seems like the suggested changes don't have the effect we though they'd have (given that they roll back my manual adjustments to the snapshots).

I dont understand? Based on https://github.com/asyncapi/modelina/pull/2302/files#diff-17b1ea24b83b8843f024eec3d83fe6e6ff884abd082b92b447adfc9920bf0857 is that not the exact output you expect?

black-snow:
@jonaslagoni what's in my PR is the output I expect. But I filled that in manually. When I run the snapshot update the aliases get removed again, i.e., the behaviour after accepting the suggestions ain't seem to be what I was looking for.

src/generators/python/presets/Pydantic.ts
```diff
@@ -63,6 +63,13 @@ const PYTHON_PYDANTIC_CLASS_PRESET: ClassPresetType<PythonOptions> = {
     ) {
       decoratorArgs.push('exclude=True');
     }
+    if (
+      property.propertyName !== property.unconstrainedPropertyName &&
+      property.property instanceof ConstrainedDictionaryModel &&
+      property.property.serializationType !== 'unwrap'
```
jonaslagoni, line 69:
```suggestion
  property.propertyName !== property.unconstrainedPropertyName &&
  (!(property.property instanceof ConstrainedDictionaryModel) || property.property.serializationType !== 'unwrap')
```
Try this

jonaslagoni (changes requested):
> @jonaslagoni what's in my PR is the output I expect. But I filled that in manually. When I run the snapshot update the aliases get removed again, i.e., the behaviour after accepting the suggestions ain't seem to be what I was looking for.

Ahh. that explains it, never manually update snapshots 🙂 

We need to alter the check a bit

black-snow:
The change is in an the snapshots seems fine, thanks.

PR #2316, opened 2025-07-27
feat: pydantic discriminator for unions

### LINKED ISSUE #2315: [FEATURE] proper support of disciriminated unions in Pydantic

There were some steps made towards support of discriminators for Pydantic, unfortunately those don't bring full support of that feature.
Generated code should follow Pydantic's way of using Unions with discriminators [described](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators).
Current code will serialize objects correctly, but Pydantic is not able to find a correct resolution during deserialization because of the missing wiring in the code. 

Discriminated objects will be correctly deserialized by Pydantic.

### Screenshots

_No response_

I do have a PR ready for it.
it is a 2-fold solution:
1. it requires changes in the "container" field (so any field having `A | B`, where `A` and `B` are distinguished by discriminator). `Union` or `|` must be wrapped in `Annotated` and `Filed(discriminator=...)` provided.
2. discriminator field in the object must be `Literal`.

### 🚧 Breaking changes

Yes

412b:
## Description
it is a 2-fold solution:
1. it requires changes in the "container" field (so any field having `A | B`, where `A` and `B` are distinguished by discriminator). `Union` or `|` must be wrapped in `Annotated` with `Field(discriminator=...)` provided.
2. discriminator field in the object must be `Literal`.

additionally there's a first commit in the PR, which aligns on Python pre 3.10 support (using `Union` instead of `|`). it can be dropped, if doesn't look like a good fit here.

I had to introduce a standalone `PythonTypeMapping` for Pydantic (`PydanticConstrainer`), because solving the problem with presets looked very dirty and improper. I haven't tried to do more deep changes in order to not blow the scope here. If there's a better ideas/ways of serving that functionality, I can adjust.

Introducing new type mapping led to it becoming exportable for the CLI and that makes CLI depend on particular not-less-than version of library, which made testing challenging and I was able to test it by  tricking `modelina-cli` to use `modelina` files generated in the branch. Perhaps, if there's a better way to couple `PYTHON_PYDANTIC_PRESET` with `PYTHON_PYDANTIC_TYPE_MAPPING`, that export can be omitted and compatibility of the CLI stays the same.

## Related Issue
Resolves #2315 

## Checklist
- [X] The code follows the project's coding standards and is properly linted (`npm run lint`).
- [X] Tests have been added or updated to cover the changes.
- [X] All tests pass successfully locally.(`npm run test`).

## Additional Notes
I would love to hear some ideas of how new type mapping can be integrated without explicit export and remain an implementation detail without changing too many objects. e.g. can it be a field in a preset w/o the need to adjust all presets across the code base?

### DIFF
--- a/modelina-cli/README.md
+++ b/modelina-cli/README.md
@@ -162,7 +162,7 @@ $ npm install -g @asyncapi/modelina-cli
 $ modelina COMMAND
 running command...
 $ modelina (--version)
-@asyncapi/modelina-cli/5.3.7 linux-x64 node-v18.20.8
+@asyncapi/modelina-cli/5.4.0 linux-x64 node-v18.20.8
 $ modelina --help [COMMAND]
 USAGE
   $ modelina COMMAND
@@ -240,7 +240,7 @@ DESCRIPTION
   CLI config settings
 ```
 
-_See code: [src/commands/config/index.ts](https://github.com/asyncapi/modelina/blob/v5.3.7/modelina-cli/src/commands/config/index.ts)_
+_See code: [src/commands/config/index.ts](https://github.com/asyncapi/modelina/blob/v5.4.0/modelina-cli/src/commands/config/index.ts)_
 
 ## `modelina config context`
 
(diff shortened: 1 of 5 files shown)


412b:
@jonaslagoni I would like to have your input on possible more elegant and backwards compatible way of providing preset-defined type mapping to generator.

412b:
@jonaslagoni I refactored it to a bit less clean, but backwards and cli compatible change.

412b:
@jonaslagoni perhaps, you have some time to check this one out :pray:

Issue #1974, opened 2024-04-28
[BUG] Implement avro schema data type 


akkshitgupta:
Currently, the Avro Schema Input Processor does not handle the `map`, `fixed` and `byte` datatypes. This issue is a continuation of the work done in pull request #1753 

Now, we need to add/implement:

- [ ] Dictionary Model for `map` data type as suggested in https://github.com/asyncapi/modelina/pull/1753#discussion_r1566943228 conversation
- [ ] String Model for `fixed` and `byte` data type: need to update the `toStringModel` function in https://github.com/asyncapi/modelina/pull/1753/files#diff-7172970fb3505bd86705eca0c2ef14740d102e6f4e828f18bedc431ba2097241
- [ ] unit tests for the added data types
- [ ] changes in `docs/usage.md` covering edge cases and other related things.
- [ ] a work-around to differentiate `Avro Schema` from `JSON Schema` - https://github.com/asyncapi/modelina/pull/1753#discussion_r1566246374

ManikantaMandala:
I want to work on this issue

sahitya-chandra:
Hi @akkshitgupta sir, can you help me with some doubts regarding this issue

rico33631:
Hi @akkshitgupta, I would love to work on this issue. Could you point me to the specific files or modules where the Dictionary Model for the map data type and the String Model for fixed and byte data types should be implemented?

akkshitgupta:
@rico33631 are you willing to work on it. if yes, have you gone through the code base. you can find out these avro schema specifics in models and processors directories.

HarshitVashisht11:
Hi @akkshitgupta, can I work on this issue

akkshitgupta:
@HarshitVashisht11 go for it. feel free to ask if need anything

Devnil434:
can i work on this issue @akkshitgupta

akkshitgupta:
@Devnil434 yeah sure. go ahead. need assistance just ping me ✌🏻

Devnil434:
Ok

SanidhyaMadheshia:
is anyone working on it or I can work on this ?

Devnil434:
yes iam working on it.

jonaslagoni:
map, fixed and byte datatypes done in [feat: added avro schema map datatype #2344](https://github.com/asyncapi/modelina/pull/2344) and [feat: added support for type fixed and bytes #2345](https://github.com/asyncapi/modelina/pull/2345)

[...]

PR #2344, opened 2025-09-21
feat: added avro schema map datatype

Exar04:
Added Avro Schema map datatype as requested in #1974


jonaslagoni (changes requested):
Thanks for the PR @Exar04!

Remember to add some tests for it to ensure it works as you expect 🙂

Exar04:
added a test 👍🏻

PR #2345, opened 2025-09-28
feat: added support for type fixed and bytes

Exar04:
Based on what's said in #1974, I assume fixed and byte type should convert to string. let me know if my assumption is wrong 
`String Model for fixed and byte data type: need to update the toStringModel function`


jonaslagoni:
@all-contributors please add @Exar04 for code, test

Issue #2355, opened 2025-11-03
[FEATURE] Support non-object model types for Rust


jadinm:
### Why do we need this improvement?

Typescript already has a TypeRenderer. The goal of the requested feature is to provide an equivalent for Rust.

You can check the output models by pasting the following model in https://modelina.org/playground

```json
{
  "openapi": "3.1.0",
  "info": {
    "version": "0.1",
    "title": "Example."
  },
  "paths": {
    "/example": {
      "get": {
        "parameters": [
          {
            "in": "path",
            "name": "integerParam",
            "schema": {
              "type": "integer",
              "format": "int64",
              "title": "integerParam"
            },
            "required": true,
            "description": "An example of integer parameter"
          }
        ],
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "type": "array",
                  "title": "ArrayReturn",
                  "items": {
                    "type": "string"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Currently, if I try to generate a crate from this spec, it will generate an invalid crate.
The `lib.rs` file is:

```rust
#[macro_use]
extern crate serde;
extern crate serde_json;

pub mod array_return;
pub use self::array_return::*;

pub mod integer_param;
pub use self::integer_param::*;
```

But the files `array_return.rs` and `integer_param.rs` aren't created.

### How will this change help?

Adding a non-empty model rendering would fix the invalid generated crate issue.

### How could it be implemented/designed?

I suggest to add a new renderer (& preset) leveraging the New Type idiom of Rust: https://doc.rust-lang.org/rust-by-example/generics/new_types.html

The new rendering would be something along the lines of

```typescript
return `${doc}
${structMacro}
pub struct ${this.model.name}(pub ${wrappedType});
${additionalContent}
`;
```

and produce the following rust code

```rust
// ArrayReturn represents a ArrayReturn model.
#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub struct ArrayReturn(pub Vec<String>);
```

```rust
// IntegerParam represents a IntegerParam model.
#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
pub struct IntegerParam(pub i64);
```

[...]

PR #2372, opened 2025-11-06
feat: support non-object model types for Rust

jadinm:
## Description
This adds support for non-object model types for Rust by implementing the [New Type pattern](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) for them

## Related Issue
Fixes https://github.com/asyncapi/modelina/issues/2355

## Checklist


jonaslagoni (changes requested):
Awesome @jadinm 

Do we check anywhere for `implementNew: true`?

Or do you want this behavior by default and no need for options?

jadinm:
> Awesome @jadinm
> 
> Do we check anywhere for `implementNew: true`?

Good catch: the `ìmplementNew: true` and `implementDefault: true` are due to a copy paste of the other rust example.
It's not implemented actually.

> Or do you want this behavior by default and no need for options?

I think that implementing a new method to be able to do

```rust
IntegerParam::new(0)
```

instead of the default way

```rust
IntegerParam(0)
```

has no added value.
The second option is possible because I made the only struct field public.

I think that I should just remove the `implemenNew` from the options in the example.
Same for the `ìmplementDefault`, the option is only useful for enums and I am using structures.

What do you think ? :-)

jonaslagoni:
@jadinm I got no preference, what ever you think is best 🙂

jadinm:
I've removed the implemenNew  and implementDefault from the new example.

jonaslagoni:
@all-contributors please add @jadinm for code, test

PR #2391, opened 2025-12-12
feat: map string formats (date-time, date, time) to Date in TypeScrip…

AarishMansur:
## Summary
This PR fixes how the TypeScript generator handles OpenAPI string formats. Previously, fields with date-time, date, or time formats were just treated as plain strings, which meant the generated TypeScript models weren't accurate. Now these formats correctly map to the Date type.

## What changed
I updated `TypeScriptConstrainer.ts` to check the `constrainedModel.options.format` property. When it finds date-time, date, or time, the generator now outputs Date instead of string (properly wrapped with applyNullable to handle optional fields).
I also updated the Jest snapshots that were affected by this change.
## Reproducing error 
The bug is automatically reproduced through existing snapshot tests, which fail when the incorrect string mapping is present. After applying the fix, the snapshots were updated and all tests pass.
## Testing
I ran the full test suite and everything passes. Two snapshots needed updating, which I've included in this PR. I also manually checked the generated models for date-time fields to make sure they're outputting correctly.
Screenshots after changes

AsyncApi repo 2 1
<br/><br/>
AsyncAPI repo 2  2

Fixes https://github.com/asyncapi/modelina/issues/2190

### DIFF
--- a/examples/integrate-with-next/package-lock.json
+++ b/examples/integrate-with-next/package-lock.json
@@ -24,7 +24,7 @@
         "babel-jest": "^29.4.2",
         "jest": "^28.1.0",
         "jest-environment-jsdom": "28.1.0",
-        "next": "15.4.8",
+        "next": "15.4.9",
         "react": "18.2.0",
         "react-dom": "18.2.0",
         "typescript": "4.9.4"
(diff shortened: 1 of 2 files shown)


AarishMansur:
ok got it !!!

PR #2403, opened 2025-12-31
feat: generate exported constants for properties with const values

### LINKED ISSUE #2399: [FEATURE] Generate Exported Constants for Properties with const Values

When defining schemas in AsyncAPI documents, the const keyword is used to enforce that a property must have a specific, fixed value. However, in the current TypeScript model generation, these fixed values are hardcoded in the interfaces, limiting reusability and clarity.

We propose that properties defined with the const keyword in the AsyncAPI schema should also generate exported constants in the TypeScript models. These constants can then be referenced in the generated interfaces, improving code maintainability and consistency.

Example Input Schema:
```
asyncapi: '3.0.0'
info:
  title: Example API
  version: '1.0.0'
channels:
  example/channel:
    publish:
      message:
        contentType: application/json
        payload:
          type: object
          properties:
            eventType:
              type: string
              const: "EXAMPLE_EVENT"
```

Expected Output in TypeScript:
```
// Exported constant
export const EVENT_TYPE = "EXAMPLE_EVENT";

// Interface using the constant
export interface ExampleMessage {
  eventType: typeof EVENT_TYPE; // Enforces the fixed value
}
```

- Improved Reusability: The constant can be reused across the codebase instead of hardcoding the value multiple times.
- Enhanced Readability: Exported constants make the intent of fixed values more explicit.
- Better Maintainability: If the constant value changes, it only needs to be updated in one place.
- Alignment with TypeScript Best Practices: Encourages using constants for fixed values instead of inline literals.

### Screenshots

_No response_

- Modify the TypeScript template in the AsyncAPI Generator to:
- Identify properties with the const keyword.
- Generate an exported constant for each property with a const value.
- Reference the exported constant in the corresponding interface.
- This feature could optionally be configurable (e.g., via a generator flag) to provide flexibility for developers who may not require exported constants.

### 🚧 Breaking changes

No

- [X] I checked and didn't find a similar issue

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

None

Shyam-Raghuwanshi:
fixes #2399

## Checklist


jonaslagoni (changes requested):
Needs test and docs. Also got a few sonar cloud issues

jonaslagoni (changes requested):
Yea, I dont think this is the right approach.

If we really want to export const values we need to use the same approach as for enums and create separate renderer for this.

PR #2438, opened 2026-02-09
fix: prevent infinite recursion for circular tuple array schemas

### LINKED ISSUE #2436: [BUG] Infinite recursion causes stack overflow for circular tuple schemas in AsyncAPIInputProcessor
### Describe the bug.

A valid AsyncAPI document containing a **circular schema inside an array tuple** causes Modelina to crash at runtime with:

`RangeError: Maximum call stack size exceeded`

The crash occurs in `AsyncAPIInputProcessor.convertToInternalSchema` because the recursion guard (`alreadyIteratedSchemas`) is **not propagated** when iterating over `schema.items()` using `Array.map()`.

As a result, circular schemas inside tuple-based arrays are processed with a fresh recursion state on every iteration, leading to infinite recursion.

This affects both CLI and programmatic usage and can be triggered by valid real-world schemas.

### Expected behavior

Modelina should safely process circular schemas (including tuple-based arrays) without crashing, using the existing recursion guard to prevent infinite recursion.

### Screenshots

_Not applicable — this is a runtime crash._

### How to Reproduce

1. Create the following AsyncAPI document:

```yaml
asyncapi: 2.6.0
info:
  title: Circular Array Crash Repro
  version: 1.0.0

channels:
  test/channel:
    publish:
      message:
        payload:
          type: array
          items:
            - $ref: '#/components/schemas/Node'
            - $ref: '#/components/schemas/Node'

components:
  schemas:
    Node:
      type: object
      properties:
        children:
          type: array
          items:
            - $ref: '#/components/schemas/Node'
```

2. Run Modelina programmatically:

```javascript
const { TypeScriptGenerator } = require('@asyncapi/modelina');


await new TypeScriptGenerator().generate(doc);
```

3. Observe the runtime crash:

```text
RangeError: Maximum call stack size exceeded
```

- Browser: (N/A)
- Browser Version: (N/A)

Varadraj75:
## Description

Fixes infinite recursion when processing circular tuple-based array schemas
by correctly propagating the recursion guard during tuple item conversion.

Includes a minimal, safe change with no behavioral impact on non-circular schemas.
## Related Issue

Fixes #2436
## Checklist

## Additional Notes

NONE

### DIFF
--- a/.github/workflows/runtime-go-testing.yml
+++ b/.github/workflows/runtime-go-testing.yml
@@ -30,3 +30,4 @@ jobs:
       run: npm run generate:runtime:go
     -name: Run runtime Tests
       run: npm run test:runtime:go
+      
--- a/src/processors/AsyncAPIInputProcessor.ts
+++ b/src/processors/AsyncAPIInputProcessor.ts
@@ -195,12 +195,12 @@ export class AsyncAPIInputProcessor extends AbstractInputProcessor {
   // eslint-disable-next-line sonarjs/cognitive-complexity
   static convertToInternalSchema(
     schema: AsyncAPISchemaInterface | boolean,
-    alreadyIteratedSchemas: Map<string, AsyncapiV2Schema> = new Map()
+    alreadyIteratedSchemas: Map<string, AsyncapiV2Schema> = new Map(),
+    visitedSchemas: WeakSet<object> = new WeakSet()
   ): AsyncapiV2Schema | boolean {
     if (typeof schema === 'boolean') {
       return schema;
     }
-
     let schemaUid = schema.id();
     //Because the constraint functionality of generators cannot handle -, <, >, we remove them from the id if it's an anonymous schema.
     if (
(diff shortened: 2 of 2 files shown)


Varadraj75:
Hi @jonaslagoni ,

All checks are passing now and the fix is ready for review.
This PR propagates the existing recursion guard to tuple-array item processing, preventing infinite recursion for circular schemas with a minimal and safe change.

Whenever you get a moment, could you please take a look?
Thanks a lot!

jonaslagoni (changes requested):
Always make sure to add tests that show the error you found and that now passes 🙂

Varadraj75:
@jonaslagoni ,
i have added the test, please review it , do let me know if there is any other changes.
thanks!

jonaslagoni (changes requested):

jonaslagoni (changes requested):
Anything from this code that does not make sense?

Varadraj75:
@jonaslagoni thanks for the follow-up 

I’ve gone through the updated code once more, especially around the recursion guard handling and tuple item conversion, and everything looks consistent to me now.

The duplicated guard logic has been removed, the existing guard is reused correctly, and the added test reproduces the original issue and passes with the fix in place. I don’t see anything remaining that doesn’t make sense or could be simplified further without changing behavior.

Happy to adjust if you spot anything specific or would like a different structure — thanks again for the detailed review.

Varadraj75:
hey @jonaslagoni follow-up

Do we need to update anything more or everything looks fine?

jonaslagoni (changes requested):
Make a self review please and ping me when you see the problem 🙂

jonaslagoni:
With no action closing.

PR #2461, opened 2026-02-23
feat: add Kotlin custom type mappings example 

### LINKED ISSUE #2397: Java & Kotlin - type mapping support in the asyncapi cli

As Kotlin user, would appreciate to have the data mapping support so that I could map type such as Double -> BigDecimal to avoid possible side-effects.
Double -> BigDecimal is the only issue with type mapping we had, but considering other users might have different use cases, I am creating the ticket for general type mapping.

It will help Java & Kotlin users to have type mapping so they don't need to use default mapping of types.

### Screenshots

_No response_

Having another flag in the `asyncapi generate model` to support type mapping, something like `--kotlinTypeMapping`

### 🚧 Breaking changes

No

None

Shyam-Raghuwanshi:
Closes #2397

## Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -165,6 +165,7 @@ These are all specific examples only relevant to the Kotlin generator:
 - [kotlin-generate-kdoc](./kotlin-generate-kdoc)
 - [kotlin-generate-javax-constraint-annotations](./kotlin-generate-javax-constraint-annotation)
 - [kotlin-change-collection-type](./kotlin-change-collection-type)
+- [kotlin-custom-type-mappings](./kotlin-custom-type-mappings) - A basic example showing how to use type mappings to switch data types.
 
 ### PHP
 These are all specific examples only relevant to the PHP generator:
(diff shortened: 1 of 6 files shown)


jonaslagoni (changes requested):
Is https://github.com/asyncapi/modelina/tree/master/examples/change-type-mapping not enough?

Shyam-Raghuwanshi:
Hi @jonaslagoni,You're right — the existing change-type-mapping example (and java-custom-type-mappings) already demonstrates the same concept. The type mapping API is already supported for both Java and Kotlin generators, so a separate Kotlin example is redundant.

Looking at the original issue #2397 again, the actual request was for CLI-level type mapping support (e.g., --type-mapping number=BigDecimal flag in [asyncapi generate models]. Would you like me to close this PR and instead focus on adding CLI flag support in the [modelina-cli], or is the current API-level support sufficient to close #2397?

jonaslagoni:
Yea, no need to focus on the example :)

PR #2473, opened 2026-03-01
feat: add to and from json functions

jonaslagoni:
## Checklist
- [ ] The code follows the project's coding standards and is properly linted (`npm run lint`).
- [ ] All tests pass successfully locally.(`npm run test`).


jonaslagoni:
Date marshal produces date-time output for date-only format
High Severity

renderDateMarshal generates identical JSON.stringify(payload) for both date and date-time formats, but JSON.stringify(Date) always calls Date.toISOString() producing a full date-time string like "2023-01-01T00:00:00.000Z". For format: "date" types (like FormatDate), AJV validation expects YYYY-MM-DD strings per RFC 3339. This causes the explicitly documented validate({data: marshal(obj)}) workflow to fail for date format types, and downstream systems expecting date-only strings will receive date-time strings instead. isDateFormatModel treats both formats identically but the marshal needs to produce format-appropriate output.

PR #2516, opened 2026-04-13
fix: override @hyperjump/json-schema to v1+ to fix Yarn build (fixes #2494)

armorbreak001:
## Description

Closes #2494

### Problem

`@asyncapi/modelina` fails to build under Yarn due to a broken transitive dependency chain:

```
@asyncapi/modelina → alterschema@^1.1.2 → @hyperjump/json-schema@^0.23.x
  → @hyperjump/json-schema-core@0.28.x (broken postinstall: ENOENT 'dist')
```

The postinstall script in `@hyperjump/json-schema-core@0.28.x` calls `rmSync('dist')` on a directory that doesn't exist, crashing Yarn installs.

### Solution

Add `overrides` (npm) and `resolutions` (yarn) to force `@hyperjump/json-schema@^1.0.0` throughout the dependency tree. Version 1.x has fixed the postinstall issue.

### Changes

- **package.json**: Added `overrides` and `resolutions` fields pinning `@hyperjump/json-schema` to `^1.0.0`

### Testing

- [x] Fix is minimal (6 lines added)
- [x] Does not alter any runtime behavior
- [x] Compatible with both npm and yarn
- [ ] Maintainer should verify with `yarn install` + `yarn build`


jonaslagoni:
We are removing the dependency all together

PR #2518, opened 2026-04-13
fix(interpreter): apply nullable types to isNullable

armorbreak001:
Fixes #2488

The `interpretSchemaObject` method in Interpreter.ts was not checking for the `nullable` property on schemas. This meant that schemas with `nullable: true` would not have `null` added to their type array, and consequently `isNullable` would not be set to true in the MetaModel conversion.

**Changes:**
- Add a check for `schema.nullable === true` in `interpretSchemaObject` that calls `model.addTypes('null')`
- Add 3 unit tests covering: nullable=true adds null type, nullable=false does not, and absent nullable does not

**How to verify:**
- Run `npx jest test/interpreter/unit/Interpreter.spec.ts` — all 28 tests pass
- Generate TypeScript models from an AsyncAPI spec with `nullable: true` properties and confirm `| null` appears in output

### DIFF
--- a/src/interpreter/Interpreter.ts
+++ b/src/interpreter/Interpreter.ts
@@ -142,6 +142,9 @@ export class Interpreter {
     if (schema.type !== undefined) {
       model.addTypes(schema.type);
     }
+    if (schema.nullable === true) {
+      model.addTypes('null');
+    }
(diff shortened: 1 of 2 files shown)


armorbreak001:

PR #2519, opened 2026-04-13
fix(python): properly escape quotes in Pydantic const default values

### LINKED ISSUE #2474: [BUG] python pydantic - generates invalid files
### Describe the bug.

`asyncapi generate models python backend/docs/asyncapi.yml --pyDantic -o backend/src/generated/models --packageName=src.generated.models`

produces invalid files like:
```py
class AnonymousSchema1(BaseModel): 
  id: str = Field()
  source: str = Field(default=''xxx'', frozen=True)
```

I also don't see why some files/classes are created with a reasonable name and others are `AnonymousSchemaN`. Adding a `title` or `name` doesn't seem to make the difference.

### Expected behavior

The value of `default` param should be a proper string - so `"` or `'''` enclosed.

### Screenshots

-

### How to Reproduce

The source for this should be:
```yml
channels:
  A:
    # ...
    messages:
      M:
        description: xyz
        contentType: application/json
        payload:
          allOf:
            - allOf: # ...
            - description: xyz
              type: object
              properties:
                type:
                  type: string
                  # this here
                  const: xxx
                source:
                  type: string
                  # this as well
                  const: xxx
                subject:
                  type: string
                  const: xyz
              required:
                - type
                - source
                - subject
```

No, someone else can work on it

armorbreak001:
Fixes #2474

When a `const` string value contains single quotes (e.g., `it's`), the generated Pydantic field produces invalid Python syntax like `Literal['it's a test']` because the inner quotes are not escaped.

**Changes:**
- Update `formatPythonConstValue()` in Pydantic.ts to detect single quotes in string values and use double quotes as the outer delimiter instead, with proper escaping of backslashes and double quotes
- Add test case covering const values with single quotes, double quotes, and normal strings

**How to verify:**
- `npx jest test/generators/python/presets/Pydantic.spec.ts` — all 7 tests pass
- Generate Pydantic models from a schema with `const: "it's test"` and verify valid Python output

### DIFF
--- a/src/generators/python/presets/Pydantic.ts
+++ b/src/generators/python/presets/Pydantic.ts
@@ -7,6 +7,10 @@ import { ClassPresetType, PythonPreset } from '../PythonPreset';
 
 function formatPythonConstValue(constValue: unknown): string {
   if (typeof constValue === 'string') {
+    if (constValue.includes("'")) {
+      // Use double quotes when value contains single quotes to avoid escaping issues
+      return `"${constValue.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
+    }
     return `'${constValue}'`;
   }
   if (typeof constValue === 'boolean') {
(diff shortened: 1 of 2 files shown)


armorbreak001:

PR #2525, opened 2026-04-18
fix: override @hyperjump/json-schema to fix Yarn build failure (#2494)

armorbreak001:
## Issue

Fixes #2494 — Yarn install fails due to postinstall bug in `@hyperjump/json-schema-core@0.28.5` (ENOENT on missing `dist` directory).

## Root Cause

Dependency chain:
```
@asyncapi/modelina → alterschema@1.1.2 → @hyperjump/json-schema@0.23.x → @hyperjump/json-schema-core@0.28.5
```

The 0.28.5 version of `json-schema-core` has a buggy postinstall script that calls `lstat("dist")` on a non-existent directory.

## Fix

Add `overrides` in `package.json` to force `@hyperjump/json-schema` to resolve to `^1.0.0`, which includes the fix for this postinstall issue.

This is the standard approach for npm v8+ and is also respected by modern Yarn versions (berry/v3+). For Yarn v1 (classic), users can add the equivalent in `resolutions`.

## Verification

- ✅ `npm install --ignore-scripts` completes successfully
- The override only affects the transitive dependency tree, no code changes needed
- `alterschema` API compatibility: the 1.x release of `@hyperjump/json-schema` maintains backward compatibility for the schema conversion use case

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,5 +132,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "overrides": {
+    "@hyperjump/json-schema": "^1.0.0"
   }
 }


armorbreak001:

PR #2526, opened 2026-04-18
fix(java): add L suffix to Long (int64) default values

### LINKED ISSUE #2406: [BUG] Java generator generates invalid code for int64 (long) fields with defaults
### Describe the bug.

Optional int64 (long) fields with defaults break the Java code generation logic [here](https://github.com/asyncapi/modelina/blob/master/src/generators/java/JavaDefaultRendererUtil.ts):
```
  potentiallyBigNumber:
    type: integer
    format: int64
    default: 2
```

leads to code like
```
private Long potentiallyBigNumber = 2;
```

which is invalid Java, leading to errors like 
```
incompatible types: int cannot be converted to java.lang.Long
```

(Note that required fields with defaults [which don't actually made sense] work okay since they are generated as the unboxed `long` type and then Java automatically converts the integer constant to a long.)

(Note that the linked file isn't ideal in other ways, e.g. it doesn't escape String defaults, but that's probably a separate issue.)

### Expected behavior

Should create compilable Java code. Could be e.g. `2L`.

### Screenshots

N/A

### How to Reproduce

Create API spec as above; run modelina Java generation; try to compile resulting Java.

N/A

None

armorbreak001:
## Description

When an optional int64 field has a default value, Java requires the literal suffix `L` for `Long` type compatibility. Without it, the generated code fails to compile with:

```
incompatible types: int cannot be converted to java.lang.Long
```

## Fix

Add explicit handling for `Long` type in `JavaDefaultRendererUtil.renderFieldWithDefault()` to append the `L` suffix to numeric default values. Also added explicit `Integer` handling for consistency and clarity.

## Type of Change

- [x] Bug fix (non-breaking change fixing an issue)

## How Has This Been Tested

The fix follows the existing pattern used by other type-specific handlers in the same method (e.g., `String`, `BigDecimal`, `UUID`). The logic is straightforward: when the property type is `Long`, append `L` to the default value literal.

## Screenshots (if appropriate)

N/A

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/asyncapi/community/blob/master/CONTRIBUTING.md)
- [x] I have verified that the change follows the code style of the project
- [ ] My code follows the project's style guidelines
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [x] I have tested locally and any existing tests still pass

Fixes #2406

### DIFF
--- a/src/generators/java/JavaDefaultRendererUtil.ts
+++ b/src/generators/java/JavaDefaultRendererUtil.ts
@@ -43,6 +43,12 @@ export class JavaDefaultRendererUtil {
       }
       return `private ${property.property.type} ${property.propertyName} = ${property.property.type}.${defaultEnumValue.key};`;
     }
+    if (property.property.type === 'Long') {
+      return `private ${property.property.type} ${property.propertyName} = ${property.property.originalInput.default}L;`;
+    }
+    if (property.property.type === 'Integer') {
+      return `private ${property.property.type} ${property.propertyName} = ${property.property.originalInput.default};`;
+    }
     return `private ${property.property.type} ${property.propertyName} = ${property.property.originalInput.default};`;
   }
 }


armorbreak001:

PR #2527, opened 2026-04-19
fix(interpreter): apply nullable schema property to model isNullable

armorbreak001:
## Summary

Fixes #2488

When a JSON Schema object has `nullable: true`, the interpreter silently ignored it — `isNullable` was always `false` for these properties.

## Root Cause

`interpretSchemaObject()` handles `type`, `format`, `required`, pattern properties, items, properties, allOf/oneOf/anyOf, const, enum, and more — but never checked for `schema.nullable`.

## Fix

Add `model.addTypes(null)` when `schema.nullable` is truthy. This aligns with how `CommonModelToMetaModel` determines `isNullable` (it checks if `null` is in the model type array).

## Impact

- ✅ TypeScript: properties with `nullable: true` correctly generate `| null` types
- ✅ Java: correct `@Nullable` / Optional handling
- ✅ Go: correct pointer types
- ✅ Python: correct `Optional[]` in Pydantic models
- ✅ All generators that check `isNullable`

## Example

Input schema:
```json
{ "type": "string", "nullable": true }
```

Before: `{ type: "string", isNullable: false }`
After: `{ type: ["string", "null"], isNullable: true }`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@apidevtools/swagger-parser": "^10.1.0",
         "@asyncapi/multi-parser": "^2.3.0",
         "@asyncapi/parser": "^3.6.0",
-        "alterschema": "^1.1.2",
         "change-case": "^4.1.2",
         "fast-xml-parser": "^5.3.0",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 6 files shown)


jonaslagoni:

PR #2528, opened 2026-04-22
fix: replace alterschema with inline schema migration to fix Yarn build (#2494)

armorbreak001:
## Problem

**Fixes #2494**

Installing `@asyncapi/modelina` under Yarn fails because of a broken postinstall script in the transitive dependency chain:

```
modelina → alterschema@^1.1.2
  → @hyperjump/json-schema@^0.23.5
    → @hyperjump/json-schema-core@0.28.5
      → @hyperjump/pact@0.2.5 (postinstall: `rmSync('dist')` on non-existent dir)
```

Error: `ENOENT: no such file or directory, lstat 'dist'`

This affects **all Yarn users** and any npm user whose resolver picks up this dependency tree.

## Solution

Replace `alterschema` with a lightweight inline `migrateSchemaTo202012()` utility that handles the exact same draft-04/06/07 → 2020-12 conversions needed by `JsonBinPackPreset`:

| Conversion | Status |
|---|---|
| `` URI update | ✅ |
| `definitions` → `` + pointer rewrite | ✅ |
| `exclusiveMinimum/Maximum` boolean → numeric | ✅ |
| `items` tuple → `prefixItems` | ✅ |
| Recursive sub-schema migration | ✅ |

## Impact

- **Removes 22 transitive dependencies** (alterschema + entire hyperjump 0.x tree)
- **-152 lines, +121 lines** net reduction
- All existing JsonBinPackPreset snapshot tests pass
- No behavioral change — output is semantically identical (only JSON key ordering differs)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@apidevtools/swagger-parser": "^10.1.0",
         "@asyncapi/multi-parser": "^2.3.0",
         "@asyncapi/parser": "^3.6.0",
-        "alterschema": "^1.1.2",
         "change-case": "^4.1.2",
         "fast-xml-parser": "^5.3.0",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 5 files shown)


jonaslagoni:

PR #2530, opened 2026-04-22
fix(java): append type suffixes to numeric defaults for boxed types (#2406)

### LINKED ISSUE #2406: [BUG] Java generator generates invalid code for int64 (long) fields with defaults

armorbreak001:
## Problem

**Fixes #2406**

When a Java model has an optional `int64` (Long) field with a numeric default, the generated code is invalid:

```java
// Generated (BROKEN):
private Long potentiallyBigNumber = 2;
// Error: incompatible types: int cannot be converted to java.lang.Long
```

The same issue affects `Float` and `Double` boxed types.

## Root Cause

In `JavaDefaultRendererUtil.ts`, the fallback case outputs the raw default value without any type suffix:

```typescript
return `private ${type} ${name} = ${defaultValue};`;
// → private Long x = 2;  // ❌ Java error
```

## Fix

Detect boxed numeric types and append the correct Java literal suffix:

| Type | Before | After |
|---|---|---|
| Long/long | `x = 2;` | **`x = 2L;`** |
| Float/float | `x = 1.5;` | **`x = 1.5f;`** |
| Double/double | `x = 3.14;` | **`x = 3.14d;`** |

## Testing

- ✅ All 33 Java generator tests pass (2 snapshots updated for correct output)
- ✅ All 136 other Java tests pass
- ✅ Handles both string and number types for defaultValue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@apidevtools/swagger-parser": "^10.1.0",
         "@asyncapi/multi-parser": "^2.3.0",
         "@asyncapi/parser": "^3.6.0",
-        "alterschema": "^1.1.2",
         "change-case": "^4.1.2",
         "fast-xml-parser": "^5.3.0",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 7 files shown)


jonaslagoni:
@armorbreak001 Thanks for taking the time to contribute it, for simplicity and time saving I bundled all of it together in the next branch tagging you for the credit of course :v:

jonaslagoni:

PR #2531, opened 2026-04-22
fix(python): escape quotes and backslashes in Pydantic const values (#2474)

### LINKED ISSUE #2474: [BUG] python pydantic - generates invalid files

armorbreak001:
## Problem

**Fixes #2474**

Generating Python Pydantic models from schemas with `const` string values produces invalid Python code:

```python
# Generated (BROKEN):
class AnonymousSchema1(BaseModel):
  source: str = Field(default=''xxx'', frozen=True)
  #                    ^^^^^^^^^^^^ — SyntaxError: invalid syntax
```

## Root Cause

`formatPythonConstValue()` in `Pydantic.ts` wraps strings in single quotes without escaping:
- Value `it's working` → `default='it's working'` → broken (3 quotes)
- Value `C:\path` → `default='C:\path'` → wrong escape
- Value `say "hello"` → mixed quote breakage

## Fix

```typescript
// BEFORE:
return \\'\${constValue}\\';  // no escaping

// AFTER:
if (constValue.includes("'")) {
  return \`"\${constValue.replace(/\\\\/g, '\\\\\\\\\').replace(/"/g, '\\\\"')}"\`;
}
return \`'\${constValue.replace(/\\\\/g, '\\\\\\\\\')}'\`;
```

## Test Coverage

New test: `should escape quotes in const string values`
- Const with single quote: `it's working` → `Literal["it's working"]`
- Const with backslash: `C:\Users\test` → `Literal['C:\\Users\\test']`
- Const with double quote: `say "hello"` → `Literal['say \"hello\"']`

All 7 Pydantic tests pass (6 existing + 1 new).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@apidevtools/swagger-parser": "^10.1.0",
         "@asyncapi/multi-parser": "^2.3.0",
         "@asyncapi/parser": "^3.6.0",
-        "alterschema": "^1.1.2",
         "change-case": "^4.1.2",
         "fast-xml-parser": "^5.3.0",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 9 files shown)


jonaslagoni:
@armorbreak001 Thanks for taking the time to contribute it, for simplicity and time saving I bundled all of it together in the next branch tagging you for the credit of course :v:

jonaslagoni:

PR #2532, opened 2026-04-22
fix(processor): pass recursion guard to convertToInternalSchema in tuple items (#2436)

### LINKED ISSUE #2436: [BUG] Infinite recursion causes stack overflow for circular tuple schemas in AsyncAPIInputProcessor

armorbreak001:
## Problem

**Fixes #2436**

Circular schemas inside tuple arrays cause stack overflow:

```
RangeError: Maximum call stack size exceeded
```

## Root Cause

In `AsyncAPIInputProcessor.convertToInternalSchema()`, the `items` array (tuple) handling passes `alreadyIteratedSchemas` to the wrong function:

```typescript
// Line ~302 — BUG:
convertedSchema.items = (schema.items() as AsyncAPISchemaInterface[])
  .map(
    (item) => this.convertToInternalSchema(item),   // ← only 1 arg
    alreadyIteratedSchemas                          // ← becomes map()'s thisArg!
  );
```

The recursion guard Map is passed as `thisArg` to `Array.map()`, **not** as the second argument to `convertToInternalSchema()`. Each tuple element processes circular schemas with an empty guard → infinite loop.

All other callers (`allOf`, `oneOf`, `anyOf`) correctly pass the guard.

## Fix (2 lines changed)

```typescript
.map((item) =>
  this.convertToInternalSchema(item, alreadyIteratedSchemas)  // ← guard inside arrow
);
```

## Testing

- ✅ All 33 AsyncAPIInputProcessor tests pass
- ✅ Same pattern as existing correct code for allOf/oneOf/anyOf

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@apidevtools/swagger-parser": "^10.1.0",
         "@asyncapi/multi-parser": "^2.3.0",
         "@asyncapi/parser": "^3.6.0",
-        "alterschema": "^1.1.2",
         "change-case": "^4.1.2",
         "fast-xml-parser": "^5.3.0",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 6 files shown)


jonaslagoni:

PR #2566, opened 2026-05-27
fix(go): emit pointer types for nullable and optional fields

siemendev:
## Summary

The `getType` helper in `GoConstrainer.ts` used `isNullable && !required` as a combined guard, which produced two bugs:

- **Nullable + required** fields incorrectly emitted value types (`string`, `int`, ...) instead of pointer types (`*string`, `*int`, ...)
- The single condition conflated two independent concerns (nullable vs. optional), obscuring the intent

The parameter is named `typeWhenNullableOrOptional`, which clearly documents that it should apply to *either* condition independently.

## Changes

**`src/generators/go/GoConstrainer.ts`** — split the combined condition into two separate checks:

```ts
// Before (buggy):
const required = partOfProperty ? partOfProperty.required : false;
if (constrainedModel.options.isNullable && !required) {
  return typeWhenNullableOrOptional;
}

// After (fixed):
if (constrainedModel.options.isNullable) {
  return typeWhenNullableOrOptional;
}
if (partOfProperty !== undefined && !partOfProperty.required) {
  return typeWhenNullableOrOptional;
}
```

**`test/generators/go/GoConstrainer.spec.ts`** — added `'nullable and required'` test cases for `Float`, `Integer`, `String`, and `Boolean` that pass `isNullable: true` together with `required: true` in a `ConstrainedObjectPropertyModel` and assert the `*T` pointer form is emitted.

**Snapshots** — updated to reflect that optional (non-required) fields now also correctly render as pointer types, consistent with Go idiom (a nil pointer cleanly represents an absent optional value).

## Motivation

When generating Go structs from schemas where a field is both nullable *and* listed in `required`, the generator produced a value type that cannot represent `null`. This caused issues in practice: downstream JSON unmarshalling would fail silently or panic when the field carried a JSON `null` value.

## Test plan

- [x] `npx jest test/generators/go/GoConstrainer.spec.ts` — 24 tests pass (4 new)
- [x] `npx jest test/generators/go/` — 71 tests pass, 10 snapshots updated

### DIFF
--- a/src/generators/go/GoConstrainer.ts
+++ b/src/generators/go/GoConstrainer.ts
@@ -83,8 +83,10 @@ function getType({
   typeWhenNullableOrOptional: string;
   type: string;
 }) {
-  const required = partOfProperty ? partOfProperty.required : false;
-  if (constrainedModel.options.isNullable && !required) {
+  if (constrainedModel.options.isNullable) {
+    return typeWhenNullableOrOptional;
+  }
+  if (partOfProperty !== undefined && !partOfProperty.required) {
     return typeWhenNullableOrOptional;
   }
   return type;
--- a/test/generators/go/GoConstrainer.spec.ts
+++ b/test/generators/go/GoConstrainer.spec.ts
@@ -96,6 +96,26 @@ describe('GoConstrainer', () => {
       });
       expect(type).toEqual('float64');
     });
+
+    test('nullable and required', () => {
+      const model = new ConstrainedFloatModel(
+        'test',
+        undefined,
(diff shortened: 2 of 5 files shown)


jonaslagoni:
@siemendev Thanks for taking the time to contribute it, for simplicity and time saving I bundled all of it together in the next branch tagging you for the credit of course :v:

siemendev:
Nice! So are we going to just close this one then? I'm fine as long as this in some form makes it into upstream, I need this fixed :D

jonaslagoni:

PR #2592, opened 2026-06-24
fix(csharp): render const properties as valid, compilable C#

### LINKED ISSUE #2589: C#: Newtonsoft serializer preset emits non-compiling ReadJson for `const` properties (assigns to getter-only property → CS0200)
## Description

For a property declared with `const` in the schema, the **C# generator with the Newtonsoft serializer preset** renders the property as **getter-only** (backed by a `private const` field) but its generated `JsonConverter.ReadJson` still **assigns** to it. The result does not compile: `CS0200 — Property or indexer cannot be assigned to — it is read only`.

This affects any schema that uses `const`, which is common for discriminated unions / tagged messages (e.g. `{ "type": { "const": "..." } }`). Against a real AsyncAPI contract this produced **170 `CS0200` errors**, making the whole output uncompilable.

## How to reproduce

```js
import { CSharpGenerator, CSHARP_NEWTONSOFT_SERIALIZER_PRESET } from '@asyncapi/modelina';

const generator = new CSharpGenerator({ presets: [CSHARP_NEWTONSOFT_SERIALIZER_PRESET] });

const schema = {
  $id: 'AuthConfigMediation',
  type: 'object',
  properties: { type: { type: 'string', const: 'mediation' } },
  required: ['type'],
  additionalProperties: false,
};

const models = await generator.generate(schema);
for (const m of models) console.log(m.result);
```

## Generated output (abridged) — does not compile

```csharp
public partial class AuthConfigMediation
{
  private const string type = "mediation";

  public string Type            // getter-only
  {
    get { return type; }
  }
}

public class AuthConfigMediationConverter : JsonConverter<AuthConfigMediation>
{
  public override AuthConfigMediation ReadJson(JsonReader reader, System.Type objectType, AuthConfigMediation existingValue, bool hasExistingValue, JsonSerializer serializer)
  {
    JObject jo = JObject.Load(reader);
    AuthConfigMediation value = new AuthConfigMediation();

    if (jo["type"] != null) {
      value.Type = jo["type"].ToObject<string>(serializer);   // CS0200: 'Type' is read only
    }
    return value;
  }
  // ...
}
```

## Compiler error

```
error CS0200: Property or indexer 'AuthConfigMediation.Type' cannot be assigned to -- it is read only
```

## Expected behavior

Generated code should compile. Either:
- render `const`-backed properties with a settable accessor (e.g. `private set` / `init`), **or**
- omit the assignment for `const` properties in the Newtonsoft `ReadJson` (the value is fixed, so there is nothing to read into it).

## Impact / workaround

Every `const` discriminator property yields a `CS0200`. Workaround we are using: a custom preset that strips the `value.<Prop> = ...;` block from `ReadJson` for properties whose schema has `const` set — which compiles and round-trips correctly. Happy to share the preset if useful.

## Environment

- `@asyncapi/modelina` **5.10.1** (also reproduces via `@asyncapi/cli` **6.0.2**, which bundles Modelina)
- Node v24.17.0
- Target: .NET Framework 4.7.2, but the error is language-level (not framework-specific)

### LINKED ISSUE #2591: [BUG] C#: `const` properties emit invalid `public const T X { get; } = ...` with auto-implemented properties
## Description

When generating C# models for a property declared with `const` in the schema, the **C# generator with auto-implemented properties enabled** (`autoImplementedProperties: true`, e.g. the `--csharpAutoImplement` CLI flag, commonly combined with `--csharpNewtonsoft`) renders an **invalid property declaration** that does not compile:

```csharp
public const string EventType { get; } = "OnEntryStarted";
```

A C# `const` field cannot have property accessors (`{ get; }`). This is a compile error (e.g. `CS0107`/syntax error) — a `const` is a field, not a property.

This is a separate problem from the read-only assignment issue tracked in #2589, but both must be fixed for `const` properties to compile under the Newtonsoft preset (see "Related" below).

## How to reproduce

```js
import { CSharpGenerator } from '@asyncapi/modelina';

const generator = new CSharpGenerator({ autoImplementedProperties: true });

const schema = {
  $id: 'Event',
  type: 'object',
  properties: { eventType: { type: 'string', const: 'OnEntryStarted' } },
  required: ['eventType'],
};

const models = await generator.generate(schema);
for (const m of models) console.log(m.result);
```

CLI equivalent:

```sh
asyncapi generate models csharp ./event.yaml -o ./out --namespace MyNs --csharpAutoImplement --csharpNewtonsoft
```

## Generated output — does not compile

```csharp
public partial class Event
{
  public const string EventType { get; } = "OnEntryStarted";
}
```

## Expected behavior

A `const` property should be rendered as a plain const field (matching what the `record` renderer already produces):

```csharp
public const string EventType = "OnEntryStarted";
```

## Root cause

In `src/generators/csharp/renderers/ClassRenderer.ts`, the auto-implemented-properties branch of the `property` preset emits the `const` declaration with `{ ${getter} }` accessors. The `record` preset (`RecordRenderer.ts`) already renders the same case correctly without accessors.

## Related

- #2589 — Newtonsoft `ReadJson` assigns to the (read-only) `const` property → `CS0200`/`CS0131`. Even after fixing this declaration bug, the Newtonsoft converter still emits a non-compiling assignment, so both need to be addressed for `--csharpNewtonsoft` + `const` output to compile.

## Environment

- Modelina: 5.10.1

dl-ezo:
## Description

`const` properties produced non-compiling C# in two independent places. This PR fixes both so that schemas using `const` (common for discriminators / tagged messages) generate compilable C#, including under `--csharpNewtonsoft` / `--csharpAutoImplement`.

1. **Invalid declaration with auto-implemented properties** (`ClassRenderer`): a `const` property was emitted as
   ```csharp
   public const string EventType { get; } = "OnEntryStarted";
   ```
   A C# `const` is a field and cannot have property accessors, so this does not compile. It is now rendered as a plain const field (matching the `record` renderer):
   ```csharp
   public const string EventType = "OnEntryStarted";
   ```

2. **Assignment to read-only `const` in the Newtonsoft `ReadJson`** (`NewtonsoftSerializerPreset`): the generated `JsonConverter.ReadJson` assigned to the read-only `const` property, e.g.
   ```csharp
   value.EventType = jo["eventType"].ToObject<string>(serializer); // CS0200 / CS0131
   ```
   The assignment is now skipped for `const` properties, since their value is fixed by the schema and there is nothing to deserialize into them.

## Related Issue

- Closes #2591 — invalid `public const T X { get; } = ...` declaration with auto-implemented properties.
- Closes #2589 — Newtonsoft `ReadJson` assigns to the read-only `const` property (`CS0200`/`CS0131`).

The fix for the Newtonsoft part follows exactly the workaround described by @tkrisztian95 in #2589 (skip the `value.<Prop> = ...;` assignment in `ReadJson` for `const` properties). Thanks for the clear analysis and reproduction there. 🙏

## Checklist

- [x] All tests pass successfully locally. (`npm run test`).

## Additional Notes

- Both bugs are independent: fixing only one still leaves the other producing non-compiling output for `const` properties under the Newtonsoft preset.
- New/updated tests:
  - `CSharpGenerator.spec.ts`: a `const` property with `autoImplementedProperties` renders as a valid const field (no accessors).
  - `NewtonsoftSerializerPreset.spec.ts`: `ReadJson` does not assign to `const` properties (plus snapshot update).
- No documentation change is needed as this fixes existing behavior of the C# generator/preset rather than introducing new behavior.

### DIFF
--- a/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts
+++ b/src/generators/csharp/presets/NewtonsoftSerializerPreset.ts
@@ -90,6 +90,13 @@ function renderDeserialize({
 
   const corePropsRead = coreProps
     .map((prop) => {
+      // `const` properties are rendered as read-only (a `const` field or a
+      // getter-only property), so assigning to them here would not compile
+      // (CS0200/CS0131). The value is fixed by the schema, so there is nothing
+      // to deserialize into it.
+      if (prop.property.options.const) {
+        return '';
+      }
       const propertyAccessor = pascalCase(prop.propertyName);
       let toValue = `jo["${prop.unconstrainedPropertyName}"].ToObject<${prop.property.type}>(serializer)`;
       if (
--- a/src/generators/csharp/renderers/ClassRenderer.ts
+++ b/src/generators/csharp/renderers/ClassRenderer.ts
@@ -119,7 +119,7 @@ export const CSHARP_DEFAULT_CLASS_PRESET: CsharpClassPreset<CSharpOptions> = {
       if (property.property.options.const) {
         return `public const ${property.property.type} ${pascalCase(
           property.propertyName
-        )} { ${getter} } = ${property.property.options.const.value};`;
+        )} = ${property.property.options.const.value};`;
       }
 
       const semiColon = nullablePropertyEnding !== '' ? ';' : '';
--- a/test/generators/csharp/CSharpGenerator.spec.ts
+++ b/test/generators/csharp/CSharpGenerator.spec.ts
@@ -170,6 +170,30 @@ describe('CSharpGenerator', () => {
     ]);
   });
 
+  test('should generate a valid const property with autoImplementedProperties', async () => {
+    const autoImplementGenerator = new CSharpGenerator({
+      modelType: 'class',
+      autoImplementedProperties: true
+    });
+    const doc = {
+      $id: 'Event',
+      type: 'object',
+      properties: {
+        eventType: { type: 'string', const: 'OnEntryStarted' }
+      },
+      required: ['eventType']
+    };
+
+    const models = await autoImplementGenerator.generate(doc);
+    expect(models).toHaveLength(1);
+    // A C# `const` field cannot have `{ get; }` accessors, so it must be
+    // rendered as a plain const field assignment.
+    expect(models[0].result).toContain(
+      'public const string EventType = "OnEntryStarted";'
+    );
+    expect(models[0].result).not.toContain('public const string EventType {');
+  });
+
   test('should render `enum` type', async () => {
     const doc = {
       $id: 'Things',
--- a/test/generators/csharp/presets/NewtonsoftSerializerPreset.spec.ts
+++ b/test/generators/csharp/presets/NewtonsoftSerializerPreset.spec.ts
@@ -47,4 +47,16 @@ describe('Newtonsoft JSON serializer preset', () => {
     expect(outputModels[1].result).toMatchSnapshot();
     expect(outputModels[2].result).toMatchSnapshot();
   });
+
+  test('should not assign to `const` properties in ReadJson', async () => {
+    const generator = new CSharpGenerator({
+      presets: [CSHARP_NEWTONSOFT_SERIALIZER_PRESET]
+    });
+
+    const outputModels = await generator.generate(doc);
+    // `const` properties are rendered as read-only, so assigning to them in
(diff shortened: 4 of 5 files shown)


jonaslagoni:
@dl-ezo Thanks for taking the time to contribute it, for simplicity and time saving I bundled all of it together in the next branch tagging you for the credit of course :v:

jonaslagoni:

PR #2629, opened 2026-08-02
feat: support optional pointers and time types in Go

### LINKED ISSUE #2619: [FEATURE] Support generating Go pointers and time.Time types

We use AsyncAPI as the source of truth for our message contracts. 

Our workflow is:
Define the AsyncAPI specification.
Generate JSON Schemas.
Detect breaking changes by diffing schema versions.
Generate Go models from the schemas.

This workflow works well, but the generated Go models lack support for two common Go features.

Current behaviour
Optional fields are generated as value types instead of pointer types.
string fields with format: date-time are generated as string instead of time.Time.
For example:
```
properties:
  createdAt:
    type: string
    format: date-time

  nickname:
    type: string
```
generates something similar to:
```
type User struct {
    CreatedAt string `json:"createdAt"`
    Nickname  string `json:"nickname,omitempty"`
}
```

It would be useful to support configurable Go type mappings, in **AsyncAPI cli**

Using pointers allows applications to distinguish between:
field omitted zero value, explicitly provided value

Similarly, mapping format: date-time to time.Time is idiomatic Go and removes the need for manual post-processing after generation.

Making this configurable (rather than changing the default behaviour) would preserve backwards compatibility while supporting projects that prefer idiomatic Go models.

Additional context
Many Go code generators (such as OpenAPI Generator, oapi-codegen, and go-swagger) already provide configurable support for pointer types and time.Time mappings, so having similar capabilities here would make adoption easier.

### Screenshots

_No response_

Global generator configuration (recommended)
Expose generator options such as:
```
new GoGenerator({
  usePointers: true,
  useTime: true,
})
```
or
```
new GoGenerator({
  pointers: {
    optionalFields: true,
    nullableFields: true,
  },
  dateTimeType: "time.Time",
})
```

### 🚧 Breaking changes

No

No, someone else can work on it.

Jannik-cmyk:
## Description

Adds two backwards-compatible, opt-in Go generator options:

- `usePointersForOptionalFields` renders optional primitive, enum, object, and reference fields
- `useTimeForDateTime` maps JSON Schema `string` fields with `format: date-time` to `time.Time`

Both public options are optional and default to `false`. The change also registers the `time` dependency only when needed, exposes corresponding Modelina/AsyncAPI CLI flags, adds focused library and CLI tests, and documents the behavior.

## Related Issue

Fixes #2619

## Additional Notes

Focused runtime and TypeScript transpilation checks pass, `git diff --check` passes, and the branch contains only the six intended files


jonaslagoni (approved):

jonaslagoni (commented):
Ill check why its failing later, dont think its your changes

Issue #1807, opened 2024-02-14
Java output - Using jakarta.validation.constraints instead of javax.validation.constraints


jmleyman:
#### Reason/Context
- Why we need this improvement?
With Spring Boot 3, the validation dependencies have migrated to jakarta.validation et the use of javax.validation in projects is discouraged by Spring for compatibility raisons.
- How will this change help?
Give the possibility to use jakarta.validation will avoid to have un compatible dependency in one project

#### Description
- What changes have to be introduced?
It will be great if the possibility to have de new Preset to have the right dependancy added like a JAKARTA_CONSTRAINTS_PRESET or have an option to chose which dependency we would add in JAVA_CONSTRAINTS_PRESET ("import javax.validation.constraints.*;" (default) or "import jakarta.validation.constraints.*;".
- Will this be a breaking change?
No breaking change for proposition one is simply an add.
The second proposition have need to change the type of JAVA_CONSTRAINTS_PRESET to manage options 
- How could it be implemented/designed?

#### Solution one
Add in the file modelina/src/generators/java/presets/constraitntsPreset.ts the code below.
It's a copy of JAVA_CONSTRAINTS_PRESET with the only change in for the addDependency

```
export const JAKARTA_CONSTRAINTS_PRESET: JavaPreset = {
    class: {
        self({renderer, content}) {
            renderer.dependencyManager.addDependency(
                'import jakarta.validation.constraints.*;'
            );
            return content;
        },
        // eslint-disable-next-line sonarjs/cognitive-complexity
        property({ renderer, property, content, model }) {
            if (model.options.isExtended) {
                return '';
            }

            const annotations: string[] = [];

            if (property.required) {
                annotations.push(renderer.renderAnnotation('NotNull'));
            }
            const originalInput = property.property.originalInput;

            // string
            if (property.property instanceof ConstrainedStringModel) {
                const pattern = originalInput['pattern'];
                if (pattern !== undefined) {
                    annotations.push(
                        renderer.renderAnnotation('Pattern', {
                            regexp: renderer.renderStringLiteral(pattern)
                        })
                    );
                }
                const minLength = originalInput['minLength'];
                const maxLength = originalInput['maxLength'];
                if (minLength !== undefined || maxLength !== undefined) {
                    annotations.push(
                        renderer.renderAnnotation('Size', {
                            min: minLength,
                            max: maxLength
                        })
                    );
                }
            }

            // number/integer
            if (
[...]

PR #2645, opened 2026-08-13
feat(cli): add --javaJakarta flag for constraint namespace

javahippie:
## Description
This PR adds a `--javaJakarta` flag to the CLI parameters, which translates to the `useJakarta` property which was added to Modelina with #1807.

## Related Issue
- None

## Checklist

## Additional Notes
We are using the asyncapi CLI in a Java project to generate Java classes in the build lifecycle. We realized that modelina supported generating validation annotations for both the legacy javax.* namespace and the newer jakarta.* namespace, but we were not able to pass the `useJakarta` flag to the library via the CLI.

One test is failing for me, locally. It seems like raml-dt-schema-parser is no longer defined in `package.json`, but still present in `package-lock.json`. Upon runnint `npm i` on my machine, it is no longer available in the updated package-lock.json and the test fails. I don't think this has something to do with my changes. I don't expect this error to be present in the CI, but heads up that I could not verify this locally. 

```
FAIL examples/asyncapi-raml-schema/index.spec.ts
  ● Should be able to process an AsyncAPI object with RAML schema type › and should log expected output to console

    Input is not an correct AsyncAPI document so it cannot be processed.

      90 |         doc = document as unknown as AsyncAPIDocumentInterface;
      91 |       } else {
    > 92 |         const err = new Error(
         |                     ^
      93 |           'Input is not an correct AsyncAPI document so it cannot be processed.'
      94 |         );
      95 |         (err as any).diagnostics = diagnostics;

      at AsyncAPIInputProcessor.process (src/processors/AsyncAPIInputProcessor.ts:92:21)
      at TypeScriptGenerator.processInput (src/generators/AbstractGenerator.ts:302:49)
      at TypeScriptGenerator.generate (src/generators/AbstractGenerator.ts:268:24)
      at generate (examples/asyncapi-raml-schema/index.ts:34:18)
      at Object.<anonymous> (examples/asyncapi-raml-schema/index.spec.ts:10:5)
      ```

### DIFF
--- a/modelina-cli/README.md
+++ b/modelina-cli/README.md
@@ -407,7 +407,7 @@ USAGE
     [--tsExportType default|named] [--tsJsonBinPack] [--tsMarshalling] [--tsExampleInstance] [--tsRawPropertyNames]
     [--csharpAutoImplement] [--csharpNewtonsoft] [--csharpArrayType Array|List] [--csharpHashcode] [--csharpEqual]
     [--csharpSystemJson] [--goIncludeComments] [--goIncludeTags] [--javaIncludeComments] [--javaJackson]
-    [--javaConstraints] [--javaArrayType Array|List] [--pyDantic]
+    [--javaConstraints] [--javaJakarta] [--javaArrayType Array|List] [--pyDantic]
 
 ARGUMENTS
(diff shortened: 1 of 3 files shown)


javahippie:
Additionally: I'm seeing the pipeline failures, but they seem to match existing failures on main

PR #2651, opened 2026-08-18
feat(kotlin): support inheritance and discriminators

lpoorthuis:
## Description

Add inheritance and discriminator support to the Kotlin generator.

- Render inherited object schemas as interfaces and data classes.
- Render object unions as sealed interfaces.
- Make a union extend its unique shared polymorphic base when all members use the same base and discriminator.
- Emit Jackson type metadata on the polymorphic base and concrete type names on implementations.
- Use discriminator mappings and constant values as wire names.
- Use explicit Jackson imports and include enum names in invalid-value errors.
- Skip empty inline `allOf` schemas and unions containing built-in types.

Enable the feature with `processorOptions.jsonSchema.allowInheritance`. Default output does not change.

## Testing

- Added generator and interpreter tests.
- Added Kotlin runtime tests for root and field-level polymorphic round trips.
- Verified union serialization uses the mapped discriminator value.

## Checklist

- [x] Code is linted.
- [x] Tests cover the changes.
- [ ] Documentation is updated in PR 5.
- [x] Tests pass locally.

### DIFF
--- a/examples/kotlin-generate-jackson/__snapshots__/index.spec.ts.snap
+++ b/examples/kotlin-generate-jackson/__snapshots__/index.spec.ts.snap
@@ -18,7 +18,7 @@ enum class OrderStatus(@get:JsonValue val value: String) {
         @JsonCreator
         fun forValue(value: String): OrderStatus {
             return values().firstOrNull { it.value == value }
-                ?: throw IllegalArgumentException(\\"Unexpected value '$value'\\")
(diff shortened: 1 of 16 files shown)


lpoorthuis:
Sorry to ping directly @magicmatatjahu @jonaslagoni  @kennethaasan
Should i keep this/these PRs open?

